PR #8, opened 2025-01-04
refactor: generate workflow list, add test coverage

rodrigomf24:
## Why?

A bug was introduced yesterday in which multiple api endpoints for the same workflow caused steps to be duplicated

## What?

- Restore refactored logic to map workflows by step once when generating the fastify workflow endpoints
- Fix to avoid multiple api endpoints for the same workflow duplicating steps
- Add jest 
- Add unit test for the function that generates the workflows list for the rest endpoints

### DIFF
--- a/packages/wistro/dev/__tests__/workflows-endpoint.test.ts
+++ b/packages/wistro/dev/__tests__/workflows-endpoint.test.ts
@@ -1,6 +1,8 @@
 import { z } from 'zod';
 import { Config, WorkflowStep } from '../config.types'
 import { generateWorkflowsList } from '../workflows-endpoint';
+import { mockWorkflowSteps } from '../../test/fixtures/mockedWorkflowSteps';
+import { mockConfigWithoutTriggers, mockValidConfig } from '../../test/fixtures/mockedConfig';
 
 // Mock randomUUID for consistent test results
 jest.mock('crypto', () => ({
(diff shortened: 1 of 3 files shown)


packages/wistro/dev/__tests__/workflows-endpoint.test.ts
```diff
@@ -0,0 +1,232 @@
...
+        port: 3000,
+      },
+      cron: {
+        cron1: { name: 'Cron Job 1', emits: 'event1', workflow: 'workflow1', cron: '* * * * *' },
+      },
+      state: {
+        adapter: 'redis',
+        host: 'localhost',
+        port: 6379,
+      }
```
sergiofilhowz: suggestion: create a fixture for and only add the relevant test data here, because we will change this type quite often and will be annoying to remember to update it here too

packages/wistro/dev/__tests__/workflows-endpoint.test.ts
```diff
@@ -0,0 +1,232 @@
...
+            name: 'Cron Job 1',
+            description: undefined,
+            emits: ['event1'],
+            action: 'cron',
+            cron: '* * * * *',
+          },
+        ],
+      },
+    ]);
+  });
```
sergiofilhowz: suggestion: in this test we want to make sure the steps are here, maybe we can check just for the ids

```suggestion
 expect(result.map(({ id }) => id)).toEqual(['workflow1']);

expect(result.map(({ steps }) => steps.map(step => step.name)).flat()).toEqual([
  'Path 1',
  'Path 2',
  'Step 1', 
  'Step 2',
  'Step 3',
  'Cron Job 1'
]);
```

sergiofilhowz: and add a test to validate the structure as a whole, but in one single item

packages/wistro/dev/__tests__/workflows-endpoint.test.ts
rodrigomf24: I think it is ok to remember to update the fixture, otherwise an unexpected change can make the test to break.

rodrigomf24: Well at least until we have Ci checks in place

PR #14, opened 2025-01-06
feat: merge UI with dev

sergiofilhowz:
we're now supporting only one command line

```
$ cd playground
$ pnpm dev
``` 

You don't need to build anything beforehand, however, changes to wistro core require a restart.
In a follow-up PR I will work on the hot reload so we don't need to care too much about restarting the process when changing something.

Only one address will be created [http://localhost:3000](http://localhost:3000) and you should see the UI

This will enable us to have custom nodes, and has better dev experience

image

### DIFF
--- a/packages/ui/src/hooks/use-list-workflows.tsx
+++ b/packages/ui/src/hooks/use-list-workflows.tsx
@@ -10,7 +10,7 @@ export const useListWorkflows = () => {
   const [workflows, setWorkflows] = useState<Workflow[]>([])
 
   useEffect(() => {
-    fetch('/workflows') // TODO add env
+    fetch('/workflows')
       .then((res) => res.json())
       .then(setWorkflows)
       .finally(() => setIsLoading(false))
--- a/packages/ui/src/routes/workflow/$id.tsx
+++ b/packages/ui/src/routes/workflow/$id.tsx
@@ -4,7 +4,7 @@ import { createFileRoute } from '@tanstack/react-router'
 export const Route = createFileRoute('/workflow/$id')({
   component: Workflow,
   loader: async ({ params }) => {
-    return fetch(`/workflows/${params.id}`) // TODO add env
+    return fetch(`/workflows/${params.id}`)
       .then((res) => res.json())
       .then((workflow) => ({ workflow }))
   },


packages/wistro/dev/node/node-runner.js
```diff
@@ -0,0 +1,101 @@
...
+function parseArgs(arg) {
+  try {
+    return JSON.parse(arg)
+  } catch {
+    return arg
+  }
+}
+
+class StateAdapter {
+  constructor(traceId, stateConfig) {
```
rodrigomf24, line 19: Issue: the state adapter should not be part of this file, there's a separate folder for state adapters

rodrigomf24: Is this a generate file?

packages/wistro/dev/python/python-runner.py
```diff
@@ -23,6 +25,38 @@ async def emit(text: Any):
   # send message
   os.write(NODEIPCFD, bytesMessage)
 
+class StateAdapter:
```
rodrigomf24, line 28: Issue: same here this should be imported

packages/wistro/dev/server.ts
```diff
@@ -35,38 +36,29 @@ export const createServer = (config: Config, workflowSteps: WorkflowStep[], even
...
 
-    fastify.route({
-      method: method.toUpperCase(),
-      url: path,
-      handler: asyncHandler(emits),
-    })
-  }
-
-  fastify.addHook('onRequest', (request, reply, done) => {
-    reply.header('Access-Control-Allow-Origin', '*')
```
rodrigomf24, line 51: Suggestion: we will need to support cors on the express server

sergiofilhowz: we don't need that anymore since it's all on the same server

packages/wistro/dev/node/node-runner.js
sergiofilhowz: it's actually the main file, I forgot to delete the TS file

unfortunately, since it's not built, we had to separate runners from ts runtime, I didn't love this implementation tbh because it's good to reuse some code, but it's not bad to completely separate the runtimes either

packages/wistro/dev/python/python-runner.py
sergiofilhowz: good catch, i removed it on the follow up work I did, but forgot it was an issue on this PR

packages/wistro/dev/server.ts
rodrigomf24: What if I deploy my steps to wistro, then on my website I have a button that performs a fetch to the trigger rest endpoint, the origin will be different.

PR #21, opened 2025-01-10
feat: lock file

rodrigomf24:
## Why?

We want to read from a lock file instead of partsing through folders on dev

## What?

- Scripts to generate the lock file with versions for flows, steps and the whole lock file
- Refactored dev scripts to read from lock file data
- Update mocked values for unit tests

### DIFF
--- a/playground/wistro.lock.json
+++ b/playground/wistro.lock.json
@@ -1,5 +1,5 @@
 {
-  "version": "7645dce3-e85c-4472-a3e7-51ee49c86488:1736474518",
+  "version": "dc318765-7f0d-48fc-a760-1ca0d9da565d:1736559581",
   "port": 3000,
   "state": {
     "adapter": "redis",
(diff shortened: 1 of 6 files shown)


playground/wistro.lock.json
```diff
@@ -0,0 +1,281 @@
...
+        }
+      ],
+      "version": "7645dce3-e85c-4472-a3e7-51ee49c86488:1736474518"
+    },
+    "ecommerce": {
+      "name": "Ecommerce Example",
+      "description": "A flow that demonstrates the use of ecommerce in a flow",
+      "steps": [
+        {
+          "filePath": "/Users/rodrigomorales/Projects/Wistro-Monorepo/playground/steps/ecommerce/cancel-order.step.ts",
```
sergiofilhowz: this shouldn't include the absolute path because each computer will have a different path and it would be agnostic of where it built. it should remove process.cwd()

```suggestion
          "filePath": "./steps/ecommerce/cancel-order.step.ts",
```

PR #37, opened 2025-01-17
add legend

mfpiccolo:


packages/ui/src/views/flow/legend.tsx
```diff
@@ -0,0 +1,63 @@
...
+              </div>
+            </div>
+          </div>
+        ))}
+      </div>
+    </div>
+  )
+}
+
+export default Legend
```
sergiofilhowz: I noticed Vite doesn't handle hot reload really well with default exports, I prefer named exports because it's easier to import and to rename

we could use this instead

```export const Legend ```

PR #39, opened 2025-01-19
fix: loading node module exports

rodrigomf24:
## Why?

image

When using `@motia/snap` as a package the dependencies from the project where motia is being used can't be acessed by `@motia/core` scripts that load the node.js file config and handler. 

Instead of loading the node.js steps through `require` we need to bundle the file and load it to extract its module exports.

## What?

- Additional boilerplate to bundle the node.js file and then extract its module exports


packages/core/src/node/get-config.ts
```diff
@@ -0,0 +1,14 @@
+import { StepConfig } from "../types";
+import { getModuleExport } from "./get-module-export";
```
sergiofilhowz, line 2: we're missing some prettier formatting in a few places

PR #44, opened 2025-01-19
Feat/virtual edges

mfpiccolo:


packages/core/src/flows-endpoint.ts
```diff
@@ -103,3 +59,135 @@ export const flowsEndpoint = (flows: LockedData['flows'], app: Express) => {
...
+  return {
+    ...createBaseStepResponse(step, id),
+    type: 'api',
+    emits: step.config.emits,
+    virtualEmits: step.config.virtualEmits ?? [],
+    subscribes: step.config.virtualSubscribes ?? undefined,
+    action: 'webhook',
+    webhookUrl: `${step.config.method} ${step.config.path}`,
+    bodySchema: step.config.bodySchema ? zodToJsonSchema(step.config.bodySchema) : undefined,
+  } as FlowStepResponse
```
sergiofilhowz: we should avoid casting because it loses the type checks TS provides, what's the error when we remove it?

sergiofilhowz: i checked locally, it's because you're using Partial inside createBaseStepResponse, you should use the following instead

```typescript
Pick<FlowStepResponse, 'name' | 'description' | 'nodeComponentPath' | 'language'| 'id'>
```

packages/core/src/flows-endpoint.ts
```diff
@@ -103,3 +59,135 @@ export const flowsEndpoint = (flows: LockedData['flows'], app: Express) => {
...
+const getNodeComponentPath = (filePath: string): string | undefined => {
+  const filePathWithoutExtension = filePath.replace(/\.[^/.]+$/, '')
+  const tsxPath = filePathWithoutExtension + '.tsx'
+  return fs.existsSync(tsxPath) ? tsxPath : undefined
+}
+
+const createEdge = (
+  sourceId: string,
+  targetId: string,
+  emitType: string,
```
sergiofilhowz, line 78: i noticed we're not using it, do we have plans to use in the future?

PR #57, opened 2025-01-23
Cron

mfpiccolo:
Add cron step

### DIFF
--- a/playground/steps/cronExample/handlePeriodicJob.step.ts
+++ b/playground/steps/cronExample/handlePeriodicJob.step.ts
@@ -10,8 +10,8 @@ export const config = {
   flows: ['cron-example'],
   input: z.object({
     timestamp: z.number(),
-    message: z.string()
-  })
+    message: z.string(),
+  }),
 }
 
 export const handler = async (input: z.infer<typeof config.input>, { logger }: FlowContext) => {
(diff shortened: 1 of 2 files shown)


packages/workbench/src/publicComponents/base-node.tsx
```diff
@@ -12,12 +12,13 @@ const baseBackgroundVariants = cva('absolute -inset-[1px] rounded-md bg-gradient
...
       api: 'from-blue-500/20 to-blue-400/10',
       noop: 'from-white/20 to-white/10',
+      cron: 'from-purple-500/20 to-purple-400/10',
     },
   },
 })
 
 type Props = PropsWithChildren<{
-  variant?: VariantProps<typeof baseBackgroundVariants>['variant']
+  variant?: 'event' | 'api' | 'noop' | 'cron'
```
sergiofilhowz: it's better to keep it because we infer the types from the props we add, otherwise we will need to remember every time we add a new variant

mfpiccolo: Good call. I fixed it I believe.

mfpiccolo: I also added to the legend.

mfpiccolo: @sergiofilhowz when you have time can you run this and verify. It was working for me

PR #64, opened 2025-01-26
chore: adding integration test back to ci

sergiofilhowz:


packages/core/src/python/python-runner.py
```diff
@@ -19,18 +20,18 @@ def parse_args(arg: str) -> Any:
         return arg
 
 class Context:
-    def __init__(self, args: Any, file_name: str):
+    def __init__(self, args: Any, sender: RpcSender, file_name: str):
         self.trace_id = args.traceId
         self.flows = args.flows
         self.file_name = file_name
-        self.sender = RpcSender()
+        self.sender = sender
```
rodrigomf24: nitpick: `sender` doesn't sound very intuitive, I would suggest changing this for state manager or soemthing more intuitive. AFAIK the rpc implementation replaced the http state manager, so maybe renaming it to `state` or `stateManager` might be best

PR #88, opened 2025-02-03
feat: create step cli command

mfpiccolo:
https://www.veed.io/view/3b97896c-52a6-49f3-8152-e9a90f3ea588?panel=share

### DIFF
--- a/packages/snap/src/cli.ts
+++ b/packages/snap/src/cli.ts
@@ -110,17 +110,14 @@ const generate = program.command('generate').description('Generate motia resourc
 generate
   .command('step')
   .description('Create a new step with interactive prompts')
-  .option(
-    '-d, --dir <step file path>',
-    'The path relative to the steps directory, used to create the step file',
-  )
+  .option('-d, --dir <step file path>', 'The path relative to the steps directory, used to create the step file')
   .action(async (arg) => {
-    const { createStep } = require('./create-step')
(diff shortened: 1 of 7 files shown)


packages/snap/src/create-step/templates/override/overrides.tsx.txt
```diff
@@ -0,0 +1,22 @@
...
+          <p className="text-sm text-muted-foreground">{{DESCRIPTION}}</p>
+          {/* Add custom visualization for {{TYPE}} step */}
+          <div className="mt-4">
+            <p className="text-xs text-muted-foreground">Flows: {JSON.stringify({{FLOWS}})}</p>
+            <p className="text-xs text-muted-foreground">Emits: {JSON.stringify({{EMITS}})}</p>
+          </div>
+        </div>
+      </CardContent>
+    </Card>
+  )
```
sergiofilhowz: i don't think this compiles because there's no '@/components/ui/card' in the project

packages/snap/src/create-step/templates/override/overrides.tsx.txt
```diff
@@ -0,0 +1,22 @@
+import React from 'react'
+import { Card, CardHeader, CardTitle, CardContent } from '@/components/ui/card'
+
+export default function {{STEP_NAME}}Override() {
```
sergiofilhowz: let's use this instead

image

PR #152, opened 2025-03-14
feat: implement middleware support for API steps

ytallo:


packages/core/src/middlewares.ts
```diff
@@ -0,0 +1,64 @@
...
+    // Simple example - in real world, you'd validate the token
+    // and check the user's role against the required role
+    if (!authHeader.includes(requiredRole)) {
+      res.status(403).json({ error: `Required role: ${requiredRole}` })
+      return
+    }
+
+    next()
+  }
+}
```
sergiofilhowz: is this a core middleware?

packages/core/src/middlewares.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    if (requests[ipStr].length >= limit) {
+      res.status(429).json({ error: 'Too many requests, please try again later' })
+      return
+    }
+
+    requests[ipStr].push(now)
+    next()
+  }
+}
```
sergiofilhowz: we need to rate limit at infrastructure level

packages/core/src/middlewares.ts
```diff
@@ -0,0 +1,64 @@
...
+
+    if (origin && (allowedOrigins.includes('*') || allowedOrigins.includes(origin as string))) {
+      res.setHeader('Access-Control-Allow-Origin', Array.isArray(origin) ? origin[0] : origin)
+      res.setHeader('Access-Control-Allow-Methods', 'GET,POST,PUT,DELETE,OPTIONS')
+      res.setHeader('Access-Control-Allow-Headers', 'Content-Type,Authorization')
+    }
+
+    next()
+  }
+}
```
sergiofilhowz: this can also be managed at infrastructure level

packages/core/src/middlewares.ts
ytallo: it's for tests

packages/core/src/middleware-composer.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export default (...middlewares: ApiMiddleware[]) => {
+  return async (req: ApiRequest, ctx: FlowContext, handler: () => Promise<ApiResponse>): Promise<ApiResponse> => {
+    const composedHandler = middlewares.reduceRight<() => Promise<ApiResponse>>(
+      (nextHandler, middleware) => () => middleware(req, ctx, nextHandler),
+      handler,
+    )
+
+    return composedHandler()
+  }
```
sergiofilhowz, line 11: question: does it ensure serial execution of middlewares?

ytallo: yeah, it does. Each middleware wraps the next one, so it ensures the serial execution

PR #161, opened 2025-03-16
fix: broken links in home documentation page

vidup:
Hey !
I was reading your docs and noticed two links were invalid in the home page. Here's a little fix :)


mfpiccolo (approved):
Thanks @vidup!

PR #168, opened 2025-03-18
Mermaid3

mfpiccolo:


packages/core/src/mermaid-generator.ts
```diff
@@ -0,0 +1,228 @@
...
+
+// Pure function to get node ID
+const getNodeId = (step: Step): string => {
+  // Create a valid mermaid node ID from the file path
+  return step.filePath.replace(/[^a-zA-Z0-9]/g, '_')
+}
+
+// Pure function to get node label
+const getNodeLabel = (step: Step): string => {
+  // Get display name for node
```
ytallo, line 29: We should get the file path from the application directory to prevent the filename from containing the entire file path, including the user's machine name.

packages/core/src/mermaid-generator.ts
```diff
@@ -0,0 +1,228 @@
...
+        isApiStep(targetStep) &&
+        targetStep.config.virtualSubscribes &&
+        Array.isArray(targetStep.config.virtualSubscribes) &&
+        targetStep.config.virtualSubscribes.includes(topic)
+      ) {
+        const targetId = getNodeId(targetStep)
+        connections += `    ${sourceId} -->|${label}| ${targetId}\n`
+      }
+    })
+  })
```
ytallo: I think we can improve this code a little bit to be better to understand

```
steps.forEach((sourceStep) => {
  const sourceId = getNodeId(sourceStep)
  
  function processEmissions(emissionsArray, stepSource, stepsCollection, sourceIdentifier) {
    if (emissionsArray && Array.isArray(emissionsArray)) {
      return generateConnections(emissionsArray, stepSource, stepsCollection, sourceIdentifier)
    }
    return ""
  }
  
  const supportsEmits = isApiStep(sourceStep) || isEventStep(sourceStep) || isCronStep(sourceStep)
  const supportsVirtualEmits = supportsEmits || isNoopStep(sourceStep)
  
  if (supportsEmits) {
    connectionsStr += processEmissions(sourceStep.config.emits, sourceStep, steps, sourceId)
  }
  
  if (supportsVirtualEmits) {
    connectionsStr += processEmissions(sourceStep.config.virtualEmits, sourceStep, steps, sourceId)
  }
})
```

packages/core/src/mermaid-generator.ts
```diff
@@ -0,0 +1,237 @@
...
+
+// Pure function to generate connections
+const generateConnections = (
+  emits: Array<string | { topic: string; label?: string }>,
+  sourceStep: Step,
+  steps: Step[],
+  sourceId: string,
+  baseDir: string,
+): string => {
+  let connections = ''
```
sergiofilhowz: suggestion: you can make this a string array

and in the end `connections.join('\n')`

each addition, instead of
```
connections += `    ${sourceId} -->|${label}| ${targetId}\n`
```

it could be 
```
connections.push(`    ${sourceId} -->|${label}| ${targetId}`)
```

packages/core/src/mermaid-generator.ts
```diff
@@ -0,0 +1,237 @@
...
+// Pure function to generate flow diagram
+const generateFlowDiagram = (flowName: string, steps: Step[], baseDir: string): string => {
+  // Start mermaid flowchart with top-down direction
+  let diagram = `flowchart TD\n`
+
+  // Add class definitions for styling with explicit text color
+  diagram += `    classDef apiStyle fill:#f96,stroke:#333,stroke-width:2px,color:#fff\n`
+  diagram += `    classDef eventStyle fill:#69f,stroke:#333,stroke-width:2px,color:#fff\n`
+  diagram += `    classDef cronStyle fill:#9c6,stroke:#333,stroke-width:2px,color:#fff\n`
+  diagram += `    classDef noopStyle fill:#3f3a50,stroke:#333,stroke-width:2px,color:#fff\n`
```
sergiofilhowz: same here for string list

ytallosilva (approved):

PR #184, opened 2025-03-25
Fix: Add data prop to BaseNode

mdp:
Double check my work here, but I get an error in the `noop.step.tsx` that the data prop is missing from BaseNode. I think this is the correct type for it.

### DIFF
--- a/packages/workbench/src/publicComponents/base-node.tsx
+++ b/packages/workbench/src/publicComponents/base-node.tsx
@@ -5,6 +5,7 @@ import React, { PropsWithChildren } from 'react'
 import { BaseHandle } from './base-handle'
 import { LanguageIndicator } from '../views/flow/nodes/language-indicator'
 import { colorMap } from './colorMap'
+import { BaseNodeProps } from './node-props'
 
 const baseDot = cva('w-[6px] h-[6px] rounded-full', {
   variants: {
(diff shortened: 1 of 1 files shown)


ytallosilva:
hey @mdp can we see your `noop.step.tsx` code?

mfpiccolo:
On a roll @mdp! Love to see it. We still have some rough edges. Glad you are helping smooth them out

mdp:
> hey @mdp can we see your `noop.step.tsx` code?

It’s the noop.step.tsx from the template, no changes. But let me replicate the type error on another system to make sure it’s not just me.

packages/workbench/src/publicComponents/base-node.tsx
```diff
@@ -21,6 +22,7 @@ type Props = PropsWithChildren<{
   title: string
   variant: VariantProps<typeof baseDot>['variant']
   language?: string
+  data?: BaseNodeProps
```
sergiofilhowz: I don't see where we use this data prop in this file. Can you share the code you're having issues with?

mdp:
Sorry, my PR is a bit confusing. The default motia 'create' project has a noop.step.tsx that has an error showing in my IDE (VSCode) for the 'data' prop being passed to BaseNode.

The 'noop' template is this one, here:
https://github.com/MotiaDev/motia/blob/main/packages/snap/src/create/templates/default/steps/noop.step.tsx.txt

I added the type to BaseNode because I wasn't sure if it 'data' was being intentionally passed to it (for some future reason). But I think the alternative is to just remove 'data' prop entirely from <BaseNode> in `noop.step.tsx` as it's not needed.

I guess what I'm saying is - I think the default 'create' project should be free of type errors/warnings in all steps, and this is the one remaining warning I had (The other two being a couple unused 'deep-equal's).

PR #185, opened 2025-03-25
Feature/deployment

ytallo:


packages/snap/src/builder/build.ts
```diff
@@ -138,13 +156,16 @@ const buildNode = async (step: Step, builder: Builder) => {
...
       sourcemap: true,
       outfile: outputJsFile,
       platform: 'node',
-    })
+    }
+
+    await esbuild.build(userConfig ? { ...defaultConfig, ...userConfig } : defaultConfig)
 
     await new Promise<void>((resolve, reject) => {
       const archive = archiver('zip', { zlib: { level: 9 } })
```
sergiofilhowz, line 171: can we make level 0 to remove compression since we're performing compression at the deployment?

```suggestion
      const archive = archiver('zip', { zlib: { level: 0 } })
```

PR #197, opened 2025-04-02
WIP Chore: First pass at a state queue example

mdp:
This is less of a PR and more of a proposal to get your thoughts on how to do this.

Problem:
- I have an API endpoint that has strict rate limiting, I can only process a few requests at a time
- I have a large number of items that I need to have an agent go through sequentially
- There's no queuing built in Motia currently

Solution:
- Use flow state to hold a queue list for processing
- If the queue is empty, add the item to the "running" state and run it
- If there's already an flow running, add the incoming item to the queue
- When a workflow finishes, emit to dequeue step to clear the running state and pull the next time off for processing.

Caveats:
- There's no real error handling
- Any retries will need to be done in the step itself.

### DIFF
--- a/playground/steps/state-queue-example/README.md
+++ b/playground/steps/state-queue-example/README.md
@@ -1,0 +1,64 @@
+# State-Managed Sequential Queue Example
+
+This example demonstrates how to build a workflow that processes items sequentially (one at a time) using Motia's built-in state management features to handle locking and queuing.
+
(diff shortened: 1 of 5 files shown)


mfpiccolo:
@mdp we are prioritizing deployment strategies at the moment, but we will definitely have queueing strategies on the roadmap once that is locked in.

PR #199, opened 2025-04-03
Fix Python asyncio event loop deprecation warning

ffleandro:
## Summary 

Removed the `asyncio.get_event_loop()` deprecated call in favor of `get_running_loop` and `new_event_loop`

## Description

`get_running_loop()` will raise a `RuntimeError` when:
- There is no event loop running in the current thread
- The code is running outside of an async context

Common scenarios where this happens:
- When the Python script starts (like in our case) - there's no event loop yet
- When running in a synchronous context that needs to initiate async operations
- In background threads that need their own event loop

In our Motia context, when the Python runner starts up, there typically won't be a running event loop yet.
This is why we need the fallback to create a new one.
The pattern we're using is a common one for scripts that need to bootstrap their own async environment.

### DIFF
--- a/packages/core/src/python/python-runner.py
+++ b/packages/core/src/python/python-runner.py
@@ -27,7 +27,11 @@ class Context:
         self.rpc = rpc
         self.state = RpcStateManager(rpc)
         self.logger = Logger(self.trace_id, self.flows, rpc)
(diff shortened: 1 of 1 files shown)


xibaoer:
信已收到！！莫急~~莫急哦  嘿嘿

PR #206, opened 2025-04-06
Adds: motia create -t python

ffleandro:
## Summary
This PR adds a Python motia template.
It also makes sure the correct Python venv is used during Development.

## Description
- Adds `motia create -t python` template
- Added `motia install` command
  - makes sure python `venv` exists and python dependencies are installed
  - Only runs if Python is configured, if not just emits a warning, but Python step files will fail
  _**- ! Needs testing on Windows**_
- Modified the `motia dev` command to make sure python `venv` is used
- Added generation of `requirements.txt` and `.gitgnore` on all motia templates


Nusnus:
Awesome!!
I was just looking for this yesterday 😅
Ended up generating typescript and asking Cursor to follow up the rules to convert to python.

PR #207, opened 2025-04-06
Adds: motia create --cursor

ffleandro:
- Adds `motia create --cursor` to copy .cursor dot-files
- Merged dot-files [repo](https://github.com/MotiaDev/dot-files) with this monorepo


packages/snap/package.json
```diff
@@ -26,7 +26,8 @@
     "motia": "dist/cjs/cli.js"
   },
   "files": [
-    "dist"
+    "dist",
+    "dot-files"
```
ffleandro: @sergiofilhowz and @ytallo wdyt on this approach?
We bundle all dot-files in the snap package, and use it when setting IDE specific flags during `motia create`, for instance, `motia create --cursor|windsurf|cline`

sergiofilhowz: i don't think we need to add it here, this is mostly for js imports to understand the files that can be imported

ytallo:
I think you should transfer the cursor rules to the dist page using a script. The regular build should ignore the files.

see the [script folder](https://github.com/MotiaDev/motia/blob/fcfb90e20fa470d63429968eb9c592a44d10f4bf/packages/snap/scripts)

ffleandro:
> I think you should transfer the cursor rules to the dist page using a script. The regular build should ignore the files.
> 
> see the [script folder](https://github.com/MotiaDev/motia/blob/fcfb90e20fa470d63429968eb9c592a44d10f4bf/packages/snap/scripts)

@ytallo something like [this](https://github.com/MotiaDev/motia/pull/207/commits/b630dc8d2dfee2e59072421f7306322e877c1242)?

ffleandro:
preview package was successfully deployed and tested ✅ 
Thanks @ytallo

PR #211, opened 2025-04-08
base ruby step for extensions and to get out of global namespace

estiens:
Hey all - I only kicked the tires on this a bit so I wouldn't just merge it right in (and was having trouble figuring out where to write the tests for it)

But I think it would be of Great Use to get step files out of the global namespace, and also provide a nice base class that can be extended or monkeypatched to add hooks, decorators, format responses, etc (input validation of data, output validation, clusters of shared functionality, etc) so just wanted to throw up a proof of concept that it's pretty straightforward

Also I think by doing load_file we are reloading every request basically, so if any constants get defined there etc ruby is not gonna like it too much

feel free to grab something out of this if you want it, otherwise let me know what you think or just close it out if busy with other stuff! I can get it cleaned up a bit more if desired but node is not my daily or even yearly driver so someone needs to just clue me in on where tests go and how they run!

### DIFF
--- a/packages/core/src/ruby/base_step.rb
+++ b/packages/core/src/ruby/base_step.rb
@@ -1,0 +1,157 @@
+# frozen_string_literal: true
+
+# require_relative './base_validator' # Not currently used
+# require_relative './step_decorators' # Not currently used
+
+class BaseStep
+  attr_reader :args, :ctx, :trace_id, :flows
+
+  def self.step_config(config = {})
+    all_emit = ['global.step_success', 'global.step_failure', 'global.error']
+    config[:emits] = (config[:emits] || []) + all_emit
+    config
+  end
+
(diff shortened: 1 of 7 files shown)


mfpiccolo:
Thanks @estiens! We will review and get back to you as soon as we can. We are heads down working on deployment at the moment so don't get discouraged if we take a bit of time.

estiens:
oh no worries :D i can always happily fork and play around if i want to add stuff I like, i imagine you're already too far beyond MVP having ruby in there at all :D Best of luck 🚢

mfpiccolo:
@estiens We ended up deprioritizing Ruby for the moment as we are focusing on deployment strategies for TS/JS/Python. We will circle back to Ruby once those are solid.

PR #214, opened 2025-04-12
Feat: Added python step input validation with Pydantic

ffleandro:
## Summary
- Added core `requirements.txt `
- Improved python `venv` management on all CLI commands
- Added Python steps input validation using Python's builtin `jsonschema` package
- Added examples of generating input schema using `Pydantic` and converting to `jsonschema` 
- Refactored `python-runner` to be more maintainabled
- Update `playground` hybrid example to use `Pydantic` input validation on all python steps


packages/snap/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "motia",
   "description": "Code-first framework for intelligent workflows",
-  "version": "0.1.0-beta.27-04062025",
+  "version": "0.1.0-beta.28",
```
sergiofilhowz: we should let CI increment it

ytallo:
There are some tests failing

packages/snap/package.json
ffleandro: Actually we were missing a merge from `main`.

PR #224, opened 2025-04-28
Readme1

mfpiccolo:


santib:
Hey 👋, just curious, is the removal of the mention to Ruby on purpose? Is the plan to drop support or something?

PR #227, opened 2025-05-01
Simple docs improvements- fix broken links

sguequierre:
- Fix 404s from structural change without redirects
- Fix typo in README.md

Found an issue with the links while looking around the docs that was created by moving Quick Start & Core Concepts into a Getting Started folder. I was unsure of how to stage the site locally from [Contribute](https://github.com/MotiaDev/motia/blob/main/packages/docs/content/docs/contribution/how-to-contribute.mdx) so this should definitely be tested, would suggest including that info on Contribute if possible.

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,7 +139,7 @@ npx motia emit --topic test-state --message '{}'
 
 ## Start building your APIs, agents and automations with simple steps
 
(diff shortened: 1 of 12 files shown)


sergiofilhowz (approved):
I checked the doc and it looks great! Thank you, Sierre, for fixing those broken links!

PR #232, opened 2025-05-14
feat: auto types generation

sergiofilhowz:


packages/core/src/types/generate-types.ts
```diff
@@ -0,0 +1,99 @@
...
+        : 'never'
+      const result = step.config.responseBody
+        ? generateTypeFromSchema(step.config.responseBody as never as JsonSchema)
+        : 'never'
+      handlers[step.config.name] = { type: 'ApiRouteHandler', generics: [input, result, emits] }
+    } else if (isCronStep(step)) {
+      const emits = generateEmitData(step.config.emits)
+      handlers[step.config.name] = { type: 'CronHandler', generics: [emits] }
+    }
+  }
```
ytallo: We can place `const emits = generateEmitData(step.config.emits)` outside the if blocks, once we have emitted for all kinds of steps.

sergiofilhowz: it was the Noop step config that doesn't have emit, but I managed to fix it, I thought the same when I was implementing it and I confess I didn't love it either lmao 🤣

sergiofilhowz: https://github.com/MotiaDev/motia/pull/232/commits/d0979ac4752a2c75a52442375fcebd0399ce85dc

packages/docs/content/docs/examples/sentiment-analysis.mdx
```diff
@@ -142,17 +137,15 @@ Sometimes you don't just want a **simple** "prompt => response." Instead, you wa
...
+    export const config: EventConfig = {
       type: 'event',
       name: 'Positive Sentiment Responder',
       subscribes: ['openai.positiveSentiment'],
       emits: [],
-      input: positiveSchema,
+      input: z.object({
+        sentiment: z.string(),
+        analysis: z.string().optional(),
+      }),
```
ytallo, line 148: why not use the new Handler on this doc?

sergiofilhowz: i thought about adding it but we're going to remove most of these examples because they're too complex

PR #233, opened 2025-05-14
Motia landing revamp

designbypebble:
Kindly change the DOMAIN_URL in utils/constants.ts to "https://motia.dev" before deploying to production.


mfpiccolo:
@designbypebble Hey it looks like styles are broken here:

image

PR #235, opened 2025-05-16
update README

sumitsaurabh927:
This PR will update the docs README with:
- Instructions on how to run the local development server
- How to start contributing
- General guidelines, and more

Here are before/after images for reference.

Before:
image
After:
image

### DIFF
--- a/packages/docs/README.md
+++ b/packages/docs/README.md
@@ -1,19 +1,68 @@
-# docs
+<a name="readme-top"></a>
 
-This is a Next.js application generated with
-[Create Fumadocs](https://github.com/fuma-nama/fumadocs).
+<div align="center">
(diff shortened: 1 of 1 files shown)


packages/docs/README.md
```diff
@@ -1,19 +1,69 @@
...
+      cd packages/docs 
+    ```
+
+4. Install all the packages:
+
+    ``` bash
+      pnpm i
+      # or
+      npm i
+    ```
```
sergiofilhowz: it should be only `pnpm install` in the root of the project, which should take care of everything

sumitsaurabh927: Resolved @sergiofilhowz ✅

PR #241, opened 2025-05-20
fix: win templates

ytallo:


packages/snap/src/create/templates/generate.ts
```diff
@@ -7,7 +7,7 @@ export type Generator = (flowDir: string) => Promise<void>
 export const generateTemplateSteps = (templateDir: string): Generator => {
   return async (flowDir: string): Promise<void> => {
     const stepsDir = path.join(__dirname, templateDir, 'steps')
-    const files = globSync(path.join(stepsDir, '*.txt'))
+    const files = globSync(path.join(stepsDir, '*.txt'), { windowsPathsNoEscape: process.platform === 'win32' })
```
sergiofilhowz, line 10: there are other usages of globSync, should we do the same?

PR #244, opened 2025-05-21
feat: streams

sergiofilhowz:
# Initial sockets implementation plus browser client

## Creating a stream
To be able to use Motia Sockets, you need to define a stream

Create a file called `open-ai.stream.ts` under `steps/` folder

```typescript
import { StateStreamConfig } from 'motia'
import { z } from 'zod'

export const config: StateStreamConfig = {
  /**
   * This will be converted in the property on the FlowContext:
   * 
   * context.streams.openai
   */
  name: 'openai',
  /**
   * Schema is important to define the type of the stream, the API
   * generated to interact with this stream will have the structure defined here
   */  
  schema: z.object({ message: z.string() }),

  /**
   * Base config is used to configure the stream
   */
  baseConfig: {
    /**
     * There are two storage types: state and custom
     * State will use the state manager to store the data.
     * 
     * Custom will use a custom storage, you need to implement 
     * the IStateStream interface
     */
    storageType: 'state',

    /**
     * When storageType is state, this will be the property saved 
     * to the state, similarly to using:
     * 
     * context.state.set(traceId, 'message')
     */
    property: 'message',
  },
}
```

Once a stream is created, it should be immediately available in FlowContext (make sure to have motia running on the project)

Then you can simply create records using the streams API in your step

```typescript
import { ApiRouteConfig, Handlers } from 'motia'
import { z } from 'zod'

export const config: ApiRouteConfig = {
  type: 'api',
  name: 'OpenAiApi',
  description: 'Call OpenAI',
  path: '/open-ai',
  method: 'POST',
  emits: ['openai-prompt'],
  flows: ['open-ai'],
  bodySchema: z.object({ message: z.string({ description: 'The message to send to OpenAI' }) }),
  responseSchema: {
    200: z.object({ message: z.string({ description: 'The message from OpenAI' }) }) 
  },
}

export const handler: Handlers['OpenAiApi'] = async (req, { traceId, logger, emit, streams }) => {
  logger.info('[Call OpenAI] Received callOpenAi event', { message: req.body.message })

  /**
   * This creates a record with empty message string to be populated in the next step
   */
  const result = await streams.openai.create(traceId, { message: '' })

  await emit({
    topic: 'openai-prompt',
    data: { message: req.body.message },
  })

  return { status: 200, body: result }
}
```

The previous step just prepares a message to be created by Open AI via OpenAI SDK stream, which will be populated in the next step

```typescript
import { EventConfig, Handlers } from 'motia'
import { OpenAI } from 'openai'
import { z } from 'zod'

export const config: EventConfig = {
  type: 'event',
  name: 'CallOpenAi',
  description: 'Call OpenAI',
  subscribes: ['openai-prompt'],
  emits: [],
  input: z.object({
    message: z.string({ description: 'The message to send to OpenAI' }),
  }),
  flows: ['open-ai'],
}

const openai = new OpenAI({ apiKey: process.env.OPENAI_API_KEY_ })

export const handler: Handlers['CallOpenAi'] = async (input, context) => {
  const { logger, traceId } = context

  logger.info('[Call OpenAI] Received callOpenAi event', input)

  const result = await openai.chat.completions.create({
    messages: [{ role: 'system', content: input.message }],
    model: 'gpt-4o-mini',
    stream: true,
  })

  const messages: string[] = []

  for await (const chunk of result) {
    messages.push(chunk.choices[0].delta.content ?? '')

    /**
     * Now we're populating a previously created message with the streamed data from OpenAI
     */
    await context.streams.openai.update(traceId, { message: messages.join('') })
  }

  logger.info('[Call OpenAI] OpenAI response', result)
}
```

## Consuming stream on the browser

```
npm install @motiadev/stream-client-react
```

Then add the provider to the root of your project

```tsx
<MotiaStreamProvider address="ws://localhost:3000">
  ...
</MotiaStreamProvider>
```

then on your component or hook, just use

```typescript
const id = '' // get the id back from the API call
// data below will be updated whenever it's updated in the server
const { data } = useStreamItem({ streamName: 'openai', id })
```


packages/core/src/streams/logs-stream.ts
```diff
@@ -0,0 +1,33 @@
...
+  get = async () => null
+  update = async () => null
+  delete = async () => null
+
+  async create(_: string, data: Log): Promise<Log> {
+    await this.send({ groupId: 'default' }, { type: 'log', data })
+    return data
+  }
+
+  async getList(): Promise<Log[]> {
```
ytallo: ?

packages/core/src/streams/flows-stream.ts
```diff
@@ -0,0 +1,41 @@
...
+  async create(_: string, data: Flow): Promise<Flow> {
+    return data
+  }
+
+  async getList(): Promise<Flow[]> {
+    return Object.entries(this.lockedData.flows).map(([id, flow]) => ({ id, name: flow.name }))
+  }
+
+  getGroupId(): string {
+    return 'default'
```
ytallo, line 46: why is it static?

packages/core/src/types.ts
```diff
@@ -4,19 +4,48 @@ import { BaseLogger, Logger } from './logger'
...
+// eslint-disable-next-line @typescript-eslint/no-empty-object-type
+export interface FlowContextStateStreams {}
+
+export interface StateStreamConfig {
+  name: string
+  schema: ZodObject<any> // eslint-disable-line @typescript-eslint/no-explicit-any
+  baseConfig:
+    | { storageType: 'state'; property: string }
+    // eslint-disable-next-line @typescript-eslint/no-explicit-any
+    | { storageType: 'custom'; factory: (state: InternalStateManager) => IStateStream<any> }
```
ytallo: is custom a stream storage type?

packages/snap/src/watcher.ts
```diff
@@ -98,18 +115,77 @@ export class Watcher {
...
+    const config = await getStreamConfig(path).catch((err) => console.error(err))
+
+    if (!stream && config) {
+      this.streamCreateHandler?.({ filePath: path, config })
+    } else if (stream && config) {
+      this.streamChangeHandler?.(stream, { filePath: path, config })
+    } else if (stream && !config) {
+      this.streamDeleteHandler?.(stream)
+    }
+  }
```
ytallo, line 139: Does it receive delete and create events?

ytallo:
It would be nice to have it in the creation templates.

packages/core/src/types.ts
sergiofilhowz: good point! I removed it from the custom streams since state stream is just a configuration now, it was a class before but we had to deal with the different languages issues

packages/snap/src/watcher.ts
sergiofilhowz: this is on the file changed, but it can create or delete

* create happens when the file existed but was invalid
* update happens when the file existed and was valid
* delete happens when file existed, was valid and became invalid

packages/core/src/streams/flows-stream.ts
sergiofilhowz: good question, i added a comment to be more explicit on the why

packages/core/src/streams/logs-stream.ts
sergiofilhowz: good question, i added a comment to be more explicit on the why

PR #246, opened 2025-05-21
Fix/windows environment

ytallo:


sergiofilhowz (dismissed):
Looks good! Let's make sure we push some @next tags and thoroughly test before pushing a new version

sergiofilhowz (approved):

PR #249, opened 2025-05-23
Fix/change to rpc

ytallo:


packages/core/src/node/rpc.ts
```diff
@@ -11,7 +11,6 @@ type RpcResponse = {
 export class RpcSender {
   private readonly pendingRequests: Record<
     string,
-    // eslint-disable-next-line @typescript-eslint/no-explicit-any
```
sergiofilhowz: i think pipeline will block this since the next line has `any`

packages/core/src/process-communication/communication-config.ts
```diff
@@ -0,0 +1,25 @@
+import { SpawnOptions } from 'child_process'
+
+export type CommunicationType = 'rpc' | 'ipc'
+
+export interface CommunicationConfig {
+  type: CommunicationType
+  spawnOptions: SpawnOptions
+}
+
+export function decideCommunicationType(command: string): CommunicationType {
```
sergiofilhowz: are we using this anywhere else?

packages/core/src/process-communication/communication-config.ts
```diff
@@ -0,0 +1,25 @@
...
+  // Decision logic: Python + Windows = RPC, everything else = IPC
+  return command === 'python' && process.platform === 'win32' ? 'rpc' : 'ipc'
+}
+
+export function createCommunicationConfig(command: string): CommunicationConfig {
+  const type = decideCommunicationType(command)
```
sergiofilhowz: can't we just change this to?

```typescript
const communicationType = command === 'python' && process.platform === 'win32' ? 'rpc' : 'ipc'
```

packages/core/src/process-communication/communication-config.ts
ytallo: it was used in other places, but now just on this file

PR #260, opened 2025-05-27
Added landing page to docs package

designbypebble:
Kindly run 
pnpm install from the root of the monorepo as usual

then
cd packages 
cd docs
pnpm build

frinally, 
pnpm dev:docs 
from the root of the monorepo

### DIFF
--- a/packages/docs/components/TextAnimations.tsx
+++ b/packages/docs/components/TextAnimations.tsx
@@ -42,7 +42,7 @@ export const TextAppearBlur: React.FC<TextAppearBlurProps> = ({ children, delay 
     <motion.span variants={variants} initial="hidden" whileInView="visible" viewport={{ once: true, amount: 0.5 }}>
       {words?.map((word, index) => (
         <motion.span key={index} variants={wordVariant} className="inline-block">
-          {word.replace(word, `${word} `)}
+          {word}&nbsp;
         </motion.span>
       ))}
     </motion.span>
--- a/packages/docs/components/Title.tsx
+++ b/packages/docs/components/Title.tsx
@@ -49,8 +49,8 @@ const Title: React.FC<TitleProps> = ({ children, className, delay = 0 }) => {
       viewport={{ once: true, amount: 0.5 }}
     >
       {words?.map((word, index) => (
-        <motion.span key={index} variants={wordVariant} className="mr-2 inline-block">
-          {word}
+        <motion.span key={index} variants={wordVariant} className="inline-block">
+          {word}&nbsp;
         </motion.span>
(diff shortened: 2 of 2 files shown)


packages/docs/README.md
```diff
@@ -1,81 +1,37 @@
...
+You can check out [the Next.js GitHub repository](https://github.com/vercel/next.js) - your feedback and contributions are welcome!
+
+## Deploy on Vercel
 
-We'd love to hear your ideas or feedback. Don't forget to star this repository!
+The easiest way to deploy your Next.js app is to use the [Vercel Platform](https://vercel.com/new?utm_medium=default-template&filter=next.js&utm_source=create-next-app&utm_campaign=create-next-app-readme) from the creators of Next.js.
 
-- X/Twitter -> [@motiadev](https://x.com/motiadev)
-- Discord -> [Join the fun!](https://discord.gg/m3XN23Kp)
+Check out our [Next.js deployment documentation](https://nextjs.org/docs/app/building-your-application/deploying) for more details.
```
sergiofilhowz: @sumitsaurabh927 should this be updated?

packages/docs/.source/index.ts
```diff
@@ -0,0 +1,33 @@
+// @ts-nocheck -- skip type checking
```
sergiofilhowz, line 1: do we need this file to be pushed?

packages/docs/.prettierrc
```diff
@@ -0,0 +1,10 @@
+{
```
sergiofilhowz, line 1: we already have a prettierrc file, why do we need a new one here?

designbypebble: Nope. whatever suits you. That's entirely your preference. We had started working with an isolated repo so we have it included.

packages/docs/.source/index.ts
designbypebble: not really. originally within your repo the .source folder is omitted, I added it for ease of access for you

packages/docs/README.md
designbypebble: This has been updated in the latest commit to your original readme.md

packages/docs/.prettierrc
sergiofilhowz: sounds good, can you remove it and make sure the code is formatted to fit our existing prettier rules?

designbypebble: Wouldn't that happen automatically when you work through the code when you need to with the prettier auto save functionality?

sergiofilhowz (approved):

PR #263, opened 2025-05-27
fix: handle flow errors and improve log console behavior

andersonleal:
https://github.com/user-attachments/assets/12cb2275-ed61-4310-8c2d-f897b149a63c

Screenshot 2025-05-27 at 19 07 13

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -90,3 +90,5 @@ agents/**/components/
 **/tsconfig.app.tsbuildinfo
 .motia/
 .mermaid/
+
+playground/motia-workbench.json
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -89,4 +89,6 @@ agents/**/components/
 
 **/tsconfig.app.tsbuildinfo
 .motia/
-.mermaid/
\ No newline at end of file
+.mermaid/
+
+playground/motia-workbench.json
```
sergiofilhowz, line 94: do we want to remove this file then?

PR #266, opened 2025-05-28
feat: add E2E testing workflows for PRs and releases, including cross…

ytallo:
…-platform support and CLI validation tests


.github/workflows/e2e-tests-pr.yml
```diff
@@ -0,0 +1,82 @@
...
+        run: |
+          echo "Test outcome: ${{ steps.run_tests.outcome }}"
+          if [ "${{ steps.run_tests.outcome }}" == "success" ]; then
+            echo "result=success" >> $GITHUB_OUTPUT
+            echo "✅ PR E2E tests passed successfully"
+          else
+            echo "result=failure" >> $GITHUB_OUTPUT
+            echo "❌ PR E2E tests failed"
+            echo "Check the test results and logs for details"
+            exit 1
```
sergiofilhowz, line 71: should we also have `continue-on-error` here?

ytallo: I don't think so; it can make the pipeline be checked as successful, and the `if: always()` below will get the result.

PR #292, opened 2025-05-30
feat: streams deploy

sergiofilhowz:
Adding streams metadata to deploy

image


packages/snap/src/cloud/build/printer.ts
```diff
@@ -1,5 +1,6 @@
 import { Step } from '@motiadev/core'
 import { Printer } from '@motiadev/core/dist/src/printer'
+import { Stream } from '@motiadev/core/dist/src/types-stream'
```
ytallo, line 3: .

sergiofilhowz: since snap is an internal package we can import like this, but i didn't want to export Stream on core because that could be accessible when importing `from 'motia'`

PR #304, opened 2025-06-02
feat: add interactive project creation prompts and refactor CLI options

andersonleal:
https://github.com/user-attachments/assets/52f8da01-d8c2-42cb-bf07-5e7e57a791d0


sergiofilhowz (dismissed):
Looks great, not on the scope of this task but would be great to have better console output when the project is created, we shouldn't print the absolute path of the generated files

rohitg00:
Looks great, I would like to recommend: Reduce prompts - skip unnecessary questions and proceed with creation. Also add a `--yes` or `-y` flag that auto-starts `npm run dev` after type generation.

PR #310, opened 2025-06-04
feat(workbench): add Jest configuration and initial tests for LogField component

andersonleal:
https://linear.app/motia/issue/MOT-334/cannot-convert-undefined-or-null-to-object-when-opening-log
- Introduced Jest configuration in `jest.config.js` for testing setup.
- Added test scripts in `package.json` for running tests.
- Created `setupTests.ts` to include custom Jest matchers from `@testing-library/jest-dom`.
- Implemented comprehensive tests for `LogField` component covering various value types (string, React nodes, objects, etc.) and edge cases.

### DIFF
--- a/packages/workbench/src/components/logs/log-field.test.tsx
+++ b/packages/workbench/src/components/logs/log-field.test.tsx
@@ -1,0 +1,211 @@
+import { render, screen } from '@testing-library/react'
+import { LogField } from './log-field'
+
+jest.mock('../../lib/utils', () => ({
+  cn: jest.fn((...args) => args.filter(Boolean).join(' ')),
+}))
+
+describe('LogField', () => {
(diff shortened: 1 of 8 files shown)


packages/workbench/src/components/logs/log-field.test.tsx
```diff
@@ -0,0 +1,211 @@
+import { render, screen } from '@testing-library/react'
+import { LogField } from './log-field'
+
+jest.mock('../../lib/utils', () => ({
```
sergiofilhowz, line 4: make sure these tests files won't be in the published artifacts

sergiofilhowz (approved):
Nice! It'll be great to have unit-tests on workbench

sergiofilhowz (approved):

PR #346, opened 2025-06-15
feat: preview version of tracing

sergiofilhowz:
Here's a screenshot of the preview version

image

## There are a few things we will want to improve
* We're storing in memory and in file, but this can slow over time, we're going to limit to 50 traces and limit the number of IO made
* The UI is a placeholder, it's usable, but not fancy and UX can be improved, we're going to work with a designer to make sure it's great to use 
* No Correlation is implemented yet


rohitg00:
LGTM!

packages/core/src/observability/tracer.ts
```diff
@@ -0,0 +1,229 @@
+import { LockedData } from '../locked-data'
```
ytallo, line 2: we can split this file, once it's a bit bigger and with different responsabilities.

packages/core/src/observability/tracer.ts
```diff
@@ -0,0 +1,229 @@
...
+
+export class StreamTracer implements Tracer {
+  constructor(
+    private readonly manager: TraceManager,
+    private readonly traceGroup: TraceGroup,
+    private readonly trace: Trace,
+    logger: Logger,
+  ) {
+    logger.addListener((level, msg, args) => {
+      this.addEvent({
```
ytallo: why is it here instead of in call-step-file.ts?

packages/core/src/observability/tracer.ts
```diff
@@ -0,0 +1,229 @@
...
+        timestamp: Date.now(),
+        level,
+        message: msg,
+        metadata: args,
+      })
+    })
+  }
+
+  end(err?: Error | { message: string; code?: string | number }) {
+    this.trace.status = err ? 'failed' : 'completed'
```
ytallo: why is not saving error message anymore?

sergiofilhowz: i forgot to re-add it, will do

packages/core/src/observability/tracer.ts
sergiofilhowz: because that would be an operation specific to the tracer, i want to listen to all log calls, to be able to add to the trace

packages/core/src/observability/tracer.ts
sergiofilhowz: image

packages/core/src/observability/tracer.ts
sergiofilhowz: good suggestion!

PR #381, opened 2025-06-24
feat: add Tabs component and integrate with Container component

andersonleal:
- Introduced a new Tabs component with TabsList, TabsTrigger, and TabsContent.
- Updated Container component to support Tabs integration.
- Added stories for Tabs and Container components in Storybook.
- Updated package.json to include @radix-ui/react-tabs dependency.

# Container
Screenshot 2025-06-24 at 18 55 48

# Container Component with tabs
Screenshot 2025-06-24 at 18 31 14
# Panel Compoennt
Screenshot 2025-06-24 at 18 31 06
# Tabs Component
Screenshot 2025-06-24 at 18 31 01


packages/ui/src/components/ui/tabs.tsx
```diff
@@ -0,0 +1,64 @@
...
+  return (
+    <TabsPrimitive.Content
+      data-slot="tabs-content"
+      className={cn("flex-1 outline-none text-foreground", className)}
+      {...props}
+    />
+  )
+}
+
+export { Tabs, TabsList, TabsTrigger, TabsContent }
```
sergiofilhowz: Can we export the components directly instead of exporting an object?

packages/ui/src/components/ui/tabs.tsx
```diff
@@ -0,0 +1,64 @@
...
+      className={cn(
+        "inline-flex rounded-none bg-transparent p-0",
+        className
+      )}
+      {...props}
+    />
+  )
+}
+
+function TabsTrigger({
```
sergiofilhowz: I think this one is slightly different than the design, the active tab is rounded and touches the bottom in design, in the code it's not touching the bottom

| Design    | Code |
| -------- | ------- |
| image | image |

packages/ui/src/components/ui/container.tsx
```diff
@@ -0,0 +1,47 @@
...
+ContainerHeader.displayName = 'ContainerHeader'
+
+const ContainerContent = React.forwardRef<HTMLDivElement, React.HTMLAttributes<HTMLDivElement>>(
+  ({ className, ...props }, ref) => (
+    <div ref={ref} className={cn('flex-1 overflow-auto p-5', className)} {...props} />
+  ),
+)
+ContainerContent.displayName = 'ContainerContent'
+
+export { Container, ContainerHeader, ContainerContent }
```
sergiofilhowz: can we make it export individually instead of exporting as object?

packages/ui/src/components/ui/tabs.tsx
sergiofilhowz: by looking at these two images, i think the colors are slightly off

PR #462, opened 2025-07-17
feat: add `--host` option to `dev` command

eeve:
#### Summary

This PR adds a new `--host` (`-H`) option to the `dev` command, allowing users to specify the hostname or network interface that the development server should bind to.

#### :sparkles: Changes

* Added `-H, --host [host]` option to the `dev` command.
* Default host is set to `localhost`.

#### :package: Example usage

```bash
motia dev --host 0.0.0.0
# or
motia dev -H 0.0.0.0
```

This will start the dev server on `http://0.0.0.0:3000`, making it accessible over the network.

#### :pencil: Notes

* `-h` was **not** used as the short flag to avoid conflict with the built-in `--help` option.
* Uses `'localhost'` as the default host to preserve existing behavior.

### DIFF
--- a/packages/docs/content/docs/concepts/cli.mdx
+++ b/packages/docs/content/docs/concepts/cli.mdx
@@ -83,6 +83,7 @@ npx motia dev [options]
 
 Options:
 - `-p, --port <port>`: The port to run the server on (default: 3000).
+- `-H, --host [host]`: The host address for the server (default: localhost).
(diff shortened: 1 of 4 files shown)


sergiofilhowz (approved):
Thank you for making this contribution. This is perfectly valid!

Great job on making sure the documentation is updated as well!

sergiofilhowz:
We will merge it shortly and are planning to publish a new version today

PR #467, opened 2025-07-18
feat: api consolidation

sergiofilhowz:
- Added a new router-node.zip and router-python.zip to the generated bundle
- Added an option to use `python_step.py` instead of `python.step.py` to allow devs to import python steps files from other files
- Removed unnecessary Vite warning
- Improved some logs
- Improved archive code to be cleaner

image


packages/core/src/step-handlers.ts
```diff
@@ -32,7 +32,7 @@ export const createStepHandlers = (motia: Motia): MotiaEventManager => {
         handlerName: step.config.name,
         handler: async (event) => {
           const { data, traceId } = event
-          const logger = event.logger.child(step)
+          const logger = event.logger.child({ step: step.config.name })
```
rodrigomf24, line 35: suggestion: it might be good to inject the traceId in the logs too

packages/snap/scripts/build.sh
```diff
@@ -14,8 +14,8 @@ npx tsc --emitDeclarationOnly --declaration --outDir dist/types
 echo "Copying templates..."
 sh scripts/move-templates.sh
 
-echo "Copying Python files..."
-sh scripts/move-python.sh
+echo "Copying builders..."
+sh scripts/move-builders.sh
```
rodrigomf24, line 18: question: what are builders? are these the step builders? it might be good to leave a note here to specify what the `move-builders.sh` file does, the others are self explanatory but builders is a bit too abstract

packages/snap/src/cloud/build/builder.ts
```diff
@@ -79,6 +80,24 @@ export class Builder {
...
   }
 
+  async buildApiSteps(steps: Step<ApiRouteConfig>[]): Promise<void> {
+    const nodeSteps = steps.filter((step) => this.determineStepType(step) === 'node')
+    const pythonSteps = steps.filter((step) => this.determineStepType(step) === 'python')
+    const nodeBuilder = this.builders.get('node')
+    const pythonBuilder = this.builders.get('python')
+
+    if (nodeSteps.length > 0 && nodeBuilder) {
+      this.printer.printApiRouterBuilding('node')
```
rodrigomf24, line 101: question: what is a `printer`? is this the same as a log to the console?

packages/snap/src/cloud/build/builders/archiver.ts
```diff
@@ -0,0 +1,25 @@
+import archiver from 'archiver'
+import fs from 'fs'
+
+export class Archiver {
```
rodrigomf24, line 4: suggestion/nitpick: archiver as the name can be confusing, I would suggest renaming to `FileArchiver`

packages/snap/src/cloud/build/builders/node/router.ts
```diff
@@ -0,0 +1,76 @@
+import type { ApiRequest, ApiRouteConfig, ApiRouteHandler, FlowContext } from '@motiadev/core'
```
rodrigomf24: question/suggestion: is this a template? if so it might be good to state that in the file name or the parent folder name

packages/snap/src/cloud/build/builders/python/router.py
```diff
@@ -0,0 +1,78 @@
...
+            context.logger.error('Internal server error', {'error': str(error)})
+            response.status_code = 500
+            return {'error': 'Internal server error'}
+
+    return route
+
+def setup_router(context) -> FastAPI:
+    app = FastAPI()
+
+    # Generated code should look like this:
```
rodrigomf24: question: is this comment intentional?

packages/snap/src/dev.ts
```diff
@@ -16,6 +16,8 @@ import { activatePythonVenv } from './utils/activate-python-env'
 import { identifyUser } from './utils/analytics'
 import { version } from './version'
 
+process.env.VITE_CJS_IGNORE_WARNING = 'true'
```
rodrigomf24, line 19: question: is it safe to suppress these warnings? did you try renaming  the vite config file to `vite.config.mjs` or `vite.config.mts.`?

packages/snap/src/utils/install-lambda-python-packages.ts
```diff
@@ -1,33 +1,39 @@
...
 
-  // Check if requirements.txt exists
-  const requirementsPath = path.join(baseDir, 'requirements.txt')
-  if (!fs.existsSync(requirementsPath)) {
-    console.warn('❌ requirements.txt not found')
-    return
-  }
+  for (const requirement of requirementsList) {
+    if (!fs.existsSync(requirement)) {
+      if (isVerbose) {
```
rodrigomf24, line 15: question/suggestion: should there be a global setting for this? if it was global you could avoid these checks throughout the code and have the check for verbose logs inside the logger.

rodrigomf24 (commented):
Looks good, I left some comments

packages/snap/src/cloud/build/builder.ts
sergiofilhowz: yea but it logs in a prettier way

packages/snap/src/cloud/build/builders/archiver.ts
sergiofilhowz: it's a wrapper on `archiver` lib

packages/snap/src/cloud/build/builders/node/router.ts
sergiofilhowz: renamed

packages/snap/src/dev.ts
sergiofilhowz: yea, we acknowledge this warning and will fix in a later version, we just don't want devs to see this warning since it's not related to them

packages/snap/src/utils/install-lambda-python-packages.ts
sergiofilhowz: it should, i agree, there are other variables we need to move to global, but they're not part of the scope of this task, i just did a small refactor on this code

packages/snap/src/cloud/build/builders/python/router.py
sergiofilhowz: yea, it's intentional so we can understand how it can ultimately look like

packages/core/src/step-handlers.ts
sergiofilhowz: it's injecting, this is a child to sub calls, but the main instance has it already

packages/snap/scripts/build.sh
sergiofilhowz: it was just python builders, but now we have some ts builders as well

PR #468, opened 2025-07-18
feat: create start command for docker

rodrigomf24:
## Why?

We need to provide a cli command to spin up the dev server boilerplate, but without some features such as the dev watcher, mermaid, and the dev middleware for workbench. In addition we want to be able to conditionally run workbench.

## What?

Create a start command that will spin up a similar boilerplate to the `dev` command but without any file listeners or extra features that are specific to local development.

### DIFF
--- a/packages/snap/src/start.ts
+++ b/packages/snap/src/start.ts
@@ -1,0 +1,96 @@
+// packages/snap/src/dev.ts
+import {
+  createEventManager,
+  createMermaidGenerator,
+  createServer,
+  createStateAdapter,
+  getProjectIdentifier,
+  trackEvent,
+} from '@motiadev/core'
+import path from 'path'
+import { flush } from '@amplitude/analytics-node'
+import { generateLockedData, getStepFiles } from './generate-locked-data'
+import { createDevWatchers } from './dev-watchers'
+import { stateEndpoints } from './dev/state-endpoints'
+import { activatePythonVenv } from './utils/activate-python-env'
+import { identifyUser } from './utils/analytics'
+import { version } from './version'
+
+// eslint-disable-next-line @typescript-eslint/no-require-imports
+require('ts-node').register({
+  transpileOnly: true,
+  compilerOptions: { module: 'commonjs' },
+})
+
+export const start = async (port: number, disableVerbose: boolean): Promise<void> => {
+  const baseDir = process.cwd()
+  const isVerbose = !disableVerbose
+
+  identifyUser()
+
+  const stepFiles = getStepFiles(baseDir)
+  const hasPythonFiles = stepFiles.some((file) => file.endsWith('.py'))
+
+  trackEvent('server_started', {
+    port,
+    verbose_mode: isVerbose,
+    has_python_files: hasPythonFiles,
+    total_step_files: stepFiles.length,
+    project_name: getProjectIdentifier(baseDir),
+  })
+
+  if (hasPythonFiles) {
+    console.log('⚙️ Activating Python environment...')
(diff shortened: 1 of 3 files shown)


packages/snap/src/start.ts
```diff
@@ -0,0 +1,88 @@
...
+    streams: Object.keys(lockedData.getStreams() || {}),
+    runtime_version: version,
+    environment: process.env.NODE_ENV || 'development',
+  })
+
+  if (!process.env.MOTIA_DOCKER_DISABLE_WORKBENCH) {
+    // eslint-disable-next-line @typescript-eslint/no-require-imports
+    const { applyMiddleware } = require('@motiadev/workbench/dist/middleware')
+    await applyMiddleware(motiaServer.app)
+  }
```
sergiofilhowz: let's remove workbench from docker

packages/snap/src/start.ts
```diff
@@ -0,0 +1,88 @@
...
+  const eventManager = createEventManager()
+  const state = createStateAdapter({
+    adapter: 'default',
+    filePath: path.join(baseDir, '.motia'),
+  })
+
+  const config = { isVerbose }
+  const motiaServer = createServer(lockedData, eventManager, state, config)
+
+  stateEndpoints(motiaServer, state)
```
sergiofilhowz: this is specific to workbench, we can remove it as well

packages/snap/src/start.ts
```diff
@@ -0,0 +1,88 @@
...
+    project_name: getProjectIdentifier(baseDir),
+  })
+
+  if (hasPythonFiles) {
+    console.log('⚙️ Activating Python environment...')
+    activatePythonVenv({ baseDir, isVerbose })
+    trackEvent('python_environment_activated')
+  }
+
+  const lockedData = await generateLockedData(baseDir)
```
sergiofilhowz, line 25: side note: this is where we will accelerate the start up when we build first since the build command generates a json file with all we need to create the locked data

packages/snap/src/start.ts
```diff
@@ -0,0 +1,88 @@
...
+})
+
+export const start = async (port: number, disableVerbose: boolean): Promise<void> => {
+  const baseDir = process.cwd()
+  const isVerbose = !disableVerbose
+
+  identifyUser()
+
+  const stepFiles = getStepFiles(baseDir)
+  const hasPythonFiles = stepFiles.some((file) => file.endsWith('.py'))
```
sergiofilhowz, line 18: side note: this is where we will accelerate the start up when we build first since the build command generates a json file with all we need to create the locked data

packages/snap/src/start.ts
```diff
@@ -0,0 +1,88 @@
...
+    trackEvent('python_environment_activated')
+  }
+
+  const lockedData = await generateLockedData(baseDir)
+
+  const eventManager = createEventManager()
+  const state = createStateAdapter({
+    adapter: 'default',
+    filePath: path.join(baseDir, '.motia'),
+  })
```
sergiofilhowz, line 31: side note: this is where we will allow users to create their own adapters

sergiofilhowz (approved):

packages/snap/src/start.ts
rodrigomf24: @sergiofilhowz  the requirement was to make it optional, is there a reason why you don't want to include workbench? or at least make it configurable?

packages/snap/src/start.ts
rodrigomf24: NOTE: we discussed this item over a meet call, this scope of work will be implemented on a separate PR

PR #504, opened 2025-07-23
feat: show the step source code in the step detail sidepanel

andersonleal:
https://github.com/user-attachments/assets/b5ee3adf-cdc5-458f-961a-9a45be8429e9


packages/core/src/helper/flows-helper.ts
```diff
@@ -138,9 +138,13 @@ const createCronStepResponse = (step: Step, id: string): FlowStepResponse => {
   }
 }
 
-const createStepResponse = (step: Step): FlowStepResponse => {
-  const id = randomUUID()
+export const STEP_NAMESPACE = '7f1c3ff2-9b00-4d0a-bdd7-efb8bca49d4f'
```
sergiofilhowz, line 141: what does this mean?

andersonleal: The goal is to keep the same UUID given a string. In our use case, the file path will generate the same UUID instead of a random one for each request.

packages/workbench/src/components/endpoints/json-editor.tsx
```diff
@@ -1,25 +1,35 @@
...
   const monaco = useMonaco()
   const theme = useThemeStore((state) => state.theme)
   const editorTheme = useMemo(() => (theme === 'dark' ? 'vs-dark' : 'light'), [theme])
 
   useEffect(() => {
     if (!monaco) return
 
+    monaco.languages.typescript.javascriptDefaults.setCompilerOptions({ isolatedModules: true })
     monaco.languages.json.jsonDefaults.setDiagnosticsOptions({
-      validate: true,
```
sergiofilhowz, line 22: won't this remove the validation we have in Endpoints?

packages/core/src/helper/flows-helper.ts
andersonleal: https://www.sohamkamani.com/uuid-versions-explained/#v5-non-random-uuids

sergiofilhowz (approved):

PR #508, opened 2025-07-24
feat: welcome to motia

sergiofilhowz:


rodrigomf24 (commented):
I left most of my feedback in comments on the preview deployment. The summary of my feedback goes around:

- Copy changes that I suggested, take them with a grain of salt, I feel in the end we will all have different perspectives on how to explain the same thing
- Some of the graphics lack context, I consider adding notes or more text next to the icons might help
- I would get rid of the top bullet list and simpli dive into each section
- Either get rid of the motia cool graphic, or position it at the top or at the bottom, it is nice but it removes focus from the main text
- I noticed a lot of focus on AI and Queues in some of the text, I made some suggestions to make it abstract of AI.

PR #509, opened 2025-07-25
[MOT-520] feat: docker command to setup a motia project with motia-docker

rodrigomf24:
## What?

Docker command to setup a motia project with motia-docker:

- Create Dockerfile with motia-docker template
- Create a .dockerignore file
- Commands to build and run the project in a Docker container

## Demo

https://www.veed.io/view/6ef044b4-3f6c-461c-a5f8-b11d75b0dec6?panel=share


rohitg00:
Hey, just went through the video. Is it good to have these setup commands while we have Docker Desktop to start containers in one click?

rodrigomf24:
I don't think all folks will use Docker desktop, so might as well give some pre built commands out of the box

rodrigomf24:
> Hey, just went through the video. Is it good to have these setup commands while we have Docker Desktop to start containers in one click?

I don't think all folks will use Docker desktop, so might as well give some pre built commands out of the box

packages/snap/src/docker/utils/print-intro.ts
```diff
@@ -0,0 +1,21 @@
...
+/_/  /_/   \\____/  /_/     /___/   /_/  |_|           
+                                                      
+________ _______ _______________ ____________________ 
+___  __ \\__  __ \\__  ____/___  //_/___  ____/___  __ \\
+__  / / /_  / / /_  /     __  ,<   __  __/   __  /_/ /
+_  /_/ / / /_/ / / /___   _  /| |  _  /___   _  _, _/ 
+/_____/  \\____/  \\____/   /_/ |_|  /_____/   /_/ |_|  
+    ` +
+      '\n\n',
+  )
```
sergiofilhowz, line 20: is this correctly formatted?
i can't really read it here

packages/snap/src/docker/build.ts
```diff
@@ -0,0 +1,18 @@
...
+export const build = async (projectName?: string): Promise<void> => {
+  printMotiaDockerIntro()
+
+  if (!projectName) {
+    const packageJsonPath = path.join(process.cwd(), 'package.json')
+    const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf-8'))
+    projectName = packageJson.name
+  }
+
+  console.log(`Preparing the docker image for project: ${projectName}`)
```
sergiofilhowz, line 15: we should avoid using console.log to print stuff Motia is doing

Please use `CLIOutputManager` to better format messages

sergiofilhowz:
```suggestion
  output.log('preparing', (message) => {
    message
      .tag('progress')
      .append('Preparing the docker image for project ')
      .append(projectName, 'dark')
  })
```

sergiofilhowz: make sure to update it to sucess after it runs

packages/docs/content/docs/concepts/deployment/self-hosted.mdx
```diff
@@ -7,6 +7,35 @@ We provide a docker image that you can use to deploy your Motia project to produ
...
+  </Step>
+  <Step>
+#### You are good to go, your project should be running on localhost under port 3000, for additional options and configuration run
+
+```bash
+npx motia@latest docker run --help
+```
+
+  </Step>
+</Steps>
```
sergiofilhowz, line 37: it would be nice to list the available inputs of each command and add a short description

packages/snap/src/docker/utils/print-intro.ts
rodrigomf24: Yes it is correclty formatted, pull it down and run it you'll see

Screenshot 2025-07-25 at 6 55 51 PM

packages/snap/src/docker/build.ts
rodrigomf24: this looks more complicated, is there a simpler version that comes close to console.log?

rodrigomf24: I'll put a pin on this and proceed with simple console.logs, I want to discuss with you if there's a simpler approach to this.

PR #513, opened 2025-07-26
[MOT-522] chore: move the motia-docker repo into the motia monorepo

rodrigomf24:
## What?

Migrate the motia-docker repo files into a package inside the motia monorepo


rodrigomf24:
@sergiofilhowz  we can tag the existing image and this should be the solution to rename it to just `motia`

`docker tag old_image_name:tag new_image_name:tag`

Once this is merged and we have the first CI deploy, I'll run the tagging manually and that should complete this work.

PR #514, opened 2025-07-26
feat: track use of docker commands

rodrigomf24:
## What?

Track use of docker commands

### DIFF
--- a/packages/snap/src/docker/build.ts
+++ b/packages/snap/src/docker/build.ts
@@ -2,17 +2,27 @@ import * as fs from 'fs'
 import * as path from 'path'
 import { printMotiaDockerIntro } from './utils/print-intro'
 import { buildDockerImage } from './utils/build-docker-image'
+import { identifyUser } from '@/utils/analytics'
+import { getProjectIdentifier, trackEvent } from '@motiadev/core'
 
 export const build = async (projectName?: string): Promise<void> => {
   printMotiaDockerIntro()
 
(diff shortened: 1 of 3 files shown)


packages/snap/src/docker/run.ts
```diff
@@ -3,17 +3,27 @@ import * as path from 'path'
...
 
+  identifyUser()
+
+  let nextProjectName = projectName
   if (!projectName) {
     const packageJsonPath = path.join(process.cwd(), 'package.json')
     const packageJson = JSON.parse(fs.readFileSync(packageJsonPath, 'utf-8'))
-    projectName = packageJson.name
+    nextProjectName = packageJson.name
   }
```
ytallo, line 19: It can be a util function to avoid duplication with the build command

packages/snap/src/docker/build.ts
```diff
@@ -2,17 +2,27 @@ import * as fs from 'fs'
...
+    nextProjectName = packageJson.name
   }
 
-  console.log(`Preparing the docker image for project: ${projectName}`)
+  trackEvent('docker_build_command', {
+    dockerImageName: nextProjectName,
+    project_name: getProjectIdentifier(process.cwd()),
+  })
+
+  console.log(`Preparing the docker image for project: ${nextProjectName}`)
```
sergiofilhowz, line 25: didn't we use CLIOutput class for these logs?

PR #541, opened 2025-08-02
fix: types generation graceful error

sergiofilhowz:
We have an user whos struggling to find out the error in types generation

image

it's an issue in the schema configuration, but they can't find out which step is failing


yashksaini-coder:
Thank you for this new feature update.

PR #565, opened 2025-08-07
fix: missing templates for docker setup command

rodrigomf24:
## Why?

The docker setup command relies on a shard package templates, on build I missed updating the move templates script to include these files

## What?

add symlink in docker command folder to the docker package templates, update the move templates script to follow symlinks and copy the docker templates

### DIFF
--- a/packages/snap/src/create/interactive.ts
+++ b/packages/snap/src/create/interactive.ts
@@ -18,10 +18,7 @@ interface CreateInteractiveArgs {
   skipConfirmation?: boolean
 }
 
-export const createInteractive = async (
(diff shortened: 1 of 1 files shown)


packages/snap/src/docker/templates
```diff
@@ -0,0 +1 @@
+../../../docker/templates
```
ytallo, line 1: is it supposed to be here?

rodrigomf24: yes it is a symlink, since the final output will have the files in the same folder, for local development we just need a symlink

PR #568, opened 2025-08-08
feat: remove express dependency to consolidated APIs

ytallo:
```
var routerPaths = {
  "Parallel Merge": { path: "/api/parallel-merge", method: "POST", handler, config },
  "Get Content": { path: "/content/:contentId", method: "GET", handler: handler2, config: config2 },
  "Content": { path: "/generate-content", method: "POST", handler: handler3, config: config3 }
};
```

### DIFF
--- a/packages/snap/src/cloud/build/builders/node/index.ts
+++ b/packages/snap/src/cloud/build/builders/node/index.ts
@@ -42,7 +42,12 @@ export class NodeBuilder implements StepBuilder {
       )
       .replace(
         '// {{router paths}}',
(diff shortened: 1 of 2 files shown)


sergiofilhowz:
wouldn't it be better if it was inverted?

```typescript
const routerPaths = {
  "POST /api/parallel-merge": { stepName: "Parallel Merge", handler: handler, config },
  "GET /content/:contentId": { stepName: "Get Content", handler: handler2, config: config2 },
  "POST /generate-content": { stepName: "Content", handler: handler3, config: config3 }
};
```

I think it will be easier to find the route

```typescript
routerPaths['event.routeKey]
```

And routeKey would be `${apiMethod} ${apiPath}`

PR #586, opened 2025-08-16
fix: images from website in google and ios

sergiofilhowz:
I believe that the image is not showing in ios because of the size since it's 3.5MB and added some information for google to get the right image and right details

Also fixing favicon
image


rohitg00:
Hey @sergiofilhowz, Thanks for solving this.

Can you also check into the image preview for `favicon.ico`, it has extra white space, `generate-image.js` in `scripts` must be causing this, as we have different sizes of transparent favicon icons in `/public`

image

rohitg00 (approved):

PR #590, opened 2025-08-17
[MOT-596] refactor: remove the requirement of the steps folder in order to genenerate the locked data

rodrigomf24:
## What?

Remove the requirement of the `steps` folder in order to generate the locked data

## Demo

Screenshot 2025-08-17 at 4 21 09 PM


sergiofilhowz:
This is missing the following files

- `packages/snap/src/dev-watchers.ts`
- `packages/snap/src/watcher.ts`

ytallo:
There are other places using  `step` directory hard coded, should we update that as well?
```
packages/snap/src/generate-locked-data.ts:10
packages/snap/src/generate-types.ts:9
packages/snap/src/create/index.ts:231
packages/snap/src/create/setup-tutorial-flow.ts:26
packages/snap/src/create-step/index.ts:14
```

sergiofilhowz:
Thank you @rodrigomf24 for doing this work. Summary here 

- We identifies potential issues with crashes if we get too broad
- We need to let the developers configure how the framework should discover their steps. We're going to add a configuration file for developers

We don't want to introduce this configuration just yet because we're working on facilitating new languages to come in (which will come with this config) and we don't want to introduce something that will likely change later.

Let's close this Pull Request for now, but this issue will be solved soon.

Issue #587, opened 2025-08-17
Motia docs links in llms.txt system prompt return 404


Mouhamed-B:
There is some links in the file starting with `/llm-docs/` which does not exist in the docs, but the rest of the link might be correct:

This link will return 404:   `[Cron Step](/llm-docs/concepts/steps/cron.md): Documentation for Cron Step.`

But not if written like:  `[Cron Step](/docs/concepts/steps/cron): Documentation for Cron Step.` 
Which point to: [https://www.motia.dev/docs/concepts/steps/cron](url)

Mouhamed-B:
The links are generated by `packages/docs/scripts/prepare-docs.js`, I'll open a PR for the fix.

@rohitg00 I was wondering if I should replace `/community' links which return 404 with the discord link or make it a redirection ?

rohitg00:
Yes, makes sense. In the future, I'll replace it with the `/community` page, which includes important links.

PR #591, opened 2025-08-18
fix(docs): correct documentation links in llms.txt system prompt

Mouhamed-B:
Patch for llms.txt generation #587


Mouhamed-B:
@rohitg00

PR #593, opened 2025-08-18
Feat/python api consolidation

ytallo:


packages/snap/src/cloud/build/builders/python/router_template.py
```diff
@@ -1,78 +1,13 @@
...
-            return Response(
-                content = '{"error": "Not found"}',
-                status_code = 404,
-                media_type = 'application/json'
-            )
-        return response
-
-    return app
+router_paths: Dict[str, RouterPath] = {
+    # {{router paths}}
```
sergiofilhowz, line 16: can you add this?

```
# Generated code should look like this:
and the example
```

PR #595, opened 2025-08-18
C# support

AshAnand34:
Added C# support for Motia

### DIFF
--- a/motia.sln
+++ b/motia.sln
@@ -1,0 +1,35 @@
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio Version 17
+VisualStudioVersion = 17.5.2.0
+MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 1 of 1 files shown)


sergiofilhowz (commented):
Thank you, @AshAnand34! 

I love it, the PR looks rock solid. I'm not an expert in C#, but you

- Implemented the same concepts we have in other runtimes
- Implemented unit-test cases for all concepts

These are some of the key points we have to take care of when creating a new runtime.

Although there are a few more:

### External dependency management

How does it work with C#? We created a motia install command to download and install dependencies for all applicable runtimes in the project. Would we need it for C#?

### Documentation on how to use the runtime
- What do we need to install to start using C# with Motia?
- How to manage dependencies?

### Builder

This one would be mostly our fault, in the way we structured the runtimes and build strategies, they should all be in the same package, but they're currently in separate places. Your PR was great to show us we have a technical debt on that.

## Conclusion

I think your PR is great, and we can merge it as is, but it will be quite important to have these other key points answered so we can maintain it confidently

PR #597, opened 2025-08-18
fix: adding streams send function

sergiofilhowz:
Streams send event is not implemented for steps

### DIFF
--- a/packages/core/src/call-step-file.ts
+++ b/packages/core/src/call-step-file.ts
@@ -3,7 +3,7 @@ import { trackEvent } from './analytics/utils'
 import { Motia } from './motia'
 import { ProcessManager } from './process-communication/process-manager'
 import { Event, Step } from './types'
-import { BaseStreamItem, StateStreamEvent } from './types-stream'
+import { BaseStreamItem, StateStreamEvent, StateStreamEventChannel } from './types-stream'
 import { isAllowedToEmit } from './utils'
 import { Logger } from './logger'
 import { Tracer } from './observability'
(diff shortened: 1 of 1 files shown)


packages/core/src/node/node-runner.ts
```diff
@@ -50,6 +50,7 @@ async function runTypescriptModule(filePath: string, event: Record<string, unkno
             sender.send(`streams.${streams.name}.set`, { groupId, id, data }),
           delete: (groupId: string, id: string) => sender.send(`streams.${streams.name}.delete`, { groupId, id }),
           getGroup: (groupId: string) => sender.send(`streams.${streams.name}.getGroup`, { groupId }),
+          send: (channel: string, event: unknown) => sender.send(`streams.${streams.name}.send`, { channel, event }),
```
ytallo: does it match with the stream type?

```
  send<T>(channel: StateStreamEventChannel, event: StateStreamEvent<T>): Promise<void>
```

sergiofilhowz: great call! i just fixed it

packages/core/src/call-step-file.ts
```diff
@@ -15,6 +15,7 @@ type StateDeleteInput = { traceId: string; key: string }
 type StateClearInput = { traceId: string }
 
 type StateStreamGetInput = { groupId: string; id: string }
+type StateStreamSendInput = { channel: StateStreamGetInput; event: StateStreamEvent<unknown> }
```
ytallo: is it not StateStreamEventChannel?

```
type StateStreamSendInput = { channel: StateStreamEventChannel; event: StateStreamEvent<unknown> }
```

ytallo: StateStreamEventChannel has the id as optional, do you think it could be an issue?

sergiofilhowz: you're right, thank you!

PR #559, opened 2025-08-06
[MOT-571] feat: motia basic tutorial

rodrigomf24:
## Why?

We need to provide a basic tutorial for users to learn how to build and test their first flow using Motia's ecosystem (framework + Workbench tools)

## What?

- Basic tutorial and boilerplate to introduce new tutorials down the road

## NOTE: 

This is not ready to merge, I made the PR to create a preview deploy and expose this for others to start reviewing.

PR #602, opened 2025-08-19
Revert "[MOT-571] feat: motia basic tutorial (#559)"

ytallo:
This reverts commit 68e3fcd1


sergiofilhowz:
We're going to push tutorial today so let's not revert it

PR #614, opened 2025-08-21
feat: fixed python builder to include local modules

ffleandro:
## Summary
This PR fixes the Python builder to trace local modules and include them in the bundled zip file.
Also fixes an issue where the requirements.txt wasn't being processed, which made the lambda site packages not being created, hence, the built zip file didn't contain Python dependencies.

For instance:
```
├── services/
│   ├── l__init__.py
│   └── service1.py
│   └── service2.py
├── steps/
```

Then in the step file:
```
from services import service1
from services.service2 import my_service_function

config = {
  ...
}
 
async def handler(input, context):
  ...
```


packages/snap/src/cloud/build/builders/node/index.ts
```diff
@@ -49,12 +49,18 @@ export class NodeBuilder implements StepBuilder {
         '// {{router paths}}',
         steps
           .map(
-            (step, index) =>
-              `'${step.config.method} ${step.config.path}': { stepName: '${step.config.name}', handler: route${index}.handler, config: route${index}.config }`,
+            (step, index) => {
+              this.listener.onBuildStart(step)
```
sergiofilhowz: we probably don't need this

packages/snap/src/cloud/build/builders/python/index.ts
```diff
@@ -24,59 +29,43 @@ export class PythonBuilder implements StepBuilder {
...
+      this.listener.onBuildProgress(step, `Added ${packages.length} packages to archive`)
+    }
 
     return normalizedEntrypointPath
   }
 
   async build(step: Step): Promise<void> {
     const entrypointPath = step.filePath.replace(this.builder.projectDir, '')
     const bundlePath = path.join('python', entrypointPath.replace(/(.*)\.py$/, '$1.zip'))
-    const normalizedEntrypointPath = entrypointPath.replace(/[.]step.py$/, '_step.py')
```
sergiofilhowz, line 44: why did we remove that? there are still projects that could be running with .step.py and they won't work in cloud

packages/core/src/__tests__/get-config.test.ts
```diff
@@ -9,7 +9,7 @@ describe('Get Config', () => {
 
   it('should get the config from a node file', async () => {
     const baseDir = __dirname
-    const mockApiStep = await getStepConfig(path.join(baseDir, 'steps', 'api-step.ts'))
+    const mockApiStep = await getStepConfig(baseDir, path.join(baseDir, 'steps', 'api-step.ts'))
```
ytallo, line 12: [blocking] This test is failing due to the breaking change in `getStepConfig` API. The function now requires two parameters (`baseDir` and `filePath`) but the test is only passing one parameter. This needs to be updated to:

```typescript
const mockApiStep = await getStepConfig(baseDir, path.join(baseDir, 'steps', 'api-step.ts'))
```

The same issue exists in lines 25 and 38 for Python and Ruby tests.

packages/snap/src/cloud/build/builders/python/index.ts
ytallo, line 44: [check] @sergiofilhowz raised a valid concern about removing the `.step.py` to `_step.py` normalization. This change could break existing projects that still use the `.step.py` naming convention. 

Could you explain the reasoning for removing this normalization? If it's intentional, we should ensure backward compatibility or provide a migration path for existing projects.

packages/snap/src/cloud/build/builders/python/python-builder.py
```diff
@@ -1,221 +1,39 @@
...
-            all_packages.update(get_package_dependencies(package_name, processed_packages))
-    
-    # Filter out built-in packages
-    non_builtin_packages = {pkg for pkg in all_packages if not is_builtin_module(pkg)}
-    
-    return sorted(list(non_builtin_packages))
+NODEIPCFD = int(os.environ.get("NODE_CHANNEL_FD", 0))
 
 def main() -> None:
     """Main entry point for the script."""
```
ytallo, line 12: [info] Good improvement! The refactoring separates concerns nicely - `trace_packages.py` handles dependency tracing while `trace_project_files.py` handles local file discovery. This makes the code more modular and testable.

packages/snap/src/cloud/build/builders/python/index.ts
```diff
@@ -24,59 +29,43 @@ export class PythonBuilder implements StepBuilder {
     const sitePackagesDir = `${process.env.PYTHON_SITE_PACKAGES}-lambda`
 
     // Get Python builder response
-    const { packages } = await this.getPythonBuilderData(step)
+    const { packages, files } = await this.getPythonBuilderData(step)
```
ytallo, line 31: [nit] The interface `PythonBuilderData` is well-structured, but consider adding JSDoc comments to document the structure:

```typescript
interface PythonBuilderData {
  /** Array of Python packages with metadata about direct vs indirect imports */
  packages: Array<{ name: string; version: string; is_direct_import: boolean }>
  /** Local project files that need to be included in the bundle */  
  files: string[]
}
```

packages/snap/src/utils/install-lambda-python-packages.ts
```diff
@@ -15,7 +15,7 @@ export const installLambdaPythonPackages = ({ isVerbose = false, requirementsLis
       if (isVerbose) {
         internalLogger.warn(`requirements.txt not found at ${requirement}`)
       }
-      return
+      continue
```
ytallo, line 18: [blocking] This change from `return` to `continue` fixes the requirements.txt processing issue mentioned in the PR description. However, this is a critical bug fix that should be highlighted in the commit message and potentially considered for a separate fix. 

The original `return` would exit the entire function when the first requirements.txt file wasn't found, preventing processing of subsequent files.

packages/snap/src/cloud/build/builders/node/index.ts
```diff
@@ -49,12 +49,18 @@ export class NodeBuilder implements StepBuilder {
         '// {{router paths}}',
         steps
           .map(
-            (step, index) =>
-              `'${step.config.method} ${step.config.path}': { stepName: '${step.config.name}', handler: route${index}.handler, config: route${index}.config }`,
+            (step, index) => {
```
ytallo: [check] @sergiofilhowz mentioned this might not be needed. The build event listeners (`onBuildStart`/`onBuildEnd`) are being called for each step in the API router building. This seems inconsistent with the approach in the Python builder where these events are called once for the entire router build process.

Could you clarify if this granular per-step tracking is intentional for the Node builder?

packages/core/src/__tests__/get-config.test.ts
ffleandro: Can you clarify your comment?
I'm passing 2 params to all `getStepConfig` calls.

packages/snap/src/cloud/build/builders/node/index.ts
ffleandro: What do you recommend then, removing all individual api steps logs from the build logs?

packages/snap/src/cloud/build/builders/python/index.ts
ffleandro: I believe at some point a merge gone wrong added duplicate code.
This code I removed was duplicate and we were processing python packages twice, both in the `build` function as well as in the `buildStep` function.

ffleandro: These AI generated comments are really annoying.
Let's continue this conversation in the original thread.

packages/snap/src/utils/install-lambda-python-packages.ts
ffleandro: So, do you want to break this PR into a separate PR?
Not sure I follow, the goal of this PR was exactly to fix this issue...

PR #617, opened 2025-08-23
feat: tutorial revamp rebased

sergiofilhowz:
New Tutorial code

https://www.loom.com/share/3486b672989f4b0db097a1c3bc2797ff

- Implemented a tutorial engine using React to walk users through (this will soon be a workbench plugin)
- To create new Tutorials, we just need to create a `tutorial.tsx` in root of the project
- Revamped the templates (there's still work we should do here to simplify)
- Added features highlight to steps so we can show important code (this widget will soon be embedded to documentation)


packages/workbench/src/components/flow/node-organizer.tsx
```diff
@@ -1,50 +1,48 @@
...
-    dagreGraph.setNode(node.id, { width: node.measured?.width, height: node.measured?.height })
-  })
+//   nodes.forEach((node) => {
+//     dagreGraph.setNode(node.id, { width: node.measured?.width, height: node.measured?.height })
+//   })
 
-  edges.forEach((edge) => {
-    if (typeof edge.label === 'string') {
-      dagreGraph.setEdge(edge.source, edge.target, {
-        label: edge.label ?? '',
```
sergiofilhowz, line 22: this organizer was constantly reseting the flow state, i disabled it for now and will create a ticket to fix it

packages/workbench/src/components/tutorial/hooks/use-tutorial-engine.ts
```diff
@@ -0,0 +1,206 @@
...
+        }
+      }
+
+      setCurrentStep(stepNumber + 1)
+      setTitle(step.title)
+      setDescription(await step.description())
+      setImage(step.image)
+      setLink(step.link)
+
+      setTimeout(async () => {
```
rodrigomf24, line 74: suggestion: I would avoid using a timeout, I was seeing some weird issues and moved over to a while loop with a timeout variable making it go down to a 0 value.

packages/workbench/src/components/tutorial/hooks/use-tutorial-engine.ts
```diff
@@ -0,0 +1,206 @@
...
+            const spaceLeft = targetRect.left
+
+            // Helper function to adjust horizontal position within viewport bounds
+            const adjustHorizontalPosition = (x: number) => {
+              // this is important to avoid the tutorial from overflowing at far left or right
+              return Math.max(20, Math.min(x, window.innerWidth - width - 20))
+            }
+
+            // Try to position below first
+            if (spaceBelow >= height + 20) {
```
rodrigomf24, line 109: suggestion: this is quite complex, there are tools that already take care of these calculations, i.e.: popover, it might be good to come back to this later and simplify

packages/workbench/src/components/tutorial/hooks/use-tutorial-engine.ts
sergiofilhowz: unfortunately it's important to use setTimeout since we're using React state, we need to wait for the next render

packages/workbench/src/components/tutorial/hooks/use-tutorial-engine.ts
sergiofilhowz: I agree, we can revisit it when we find edge-case issues

PR #635, opened 2025-08-27
Feature/rest deployment endpoint

ytallo:


packages/snap/src/cloud/endpoints.ts
```diff
@@ -0,0 +1,126 @@
...
+      res.json({
+        success: true,
+        message: 'Deployment started',
+        deploymentId: sessionId,
+        streamName: 'deployment-status',
+        groupId: 'deployments',
+        itemId: sessionId,
+      })
+
+      setImmediate(async () => {
```
sergiofilhowz, line 57: what's the reason we need this?

ytallo: It will enforce the process will keep running in background

PR #645, opened 2025-09-01
fix/e2e-endpointpage-monaco-safe-cast

isksamiul:
Replaced direct window usage in page.evaluate with globalThis to avoid TS “cannot find name window”.

Removed @ts-ignore by safely casting within the browser context.

Added optional chaining and method existence checks when accessing monaco editor.

Preserves existing behavior for setting and reading editor content.

Why:

Fixes TypeScript errors in e2e package without suppressing type checks.

Improves safety and readability of the evaluate blocks.

### DIFF
--- a/packages/e2e/src/page-objects/EndpointPage.ts
+++ b/packages/e2e/src/page-objects/EndpointPage.ts
@@ -22,9 +22,11 @@ export class EndpointPage extends MotiaApplicationPage {
   async setValueInBodyEditor(value: string) {
     await this.page.waitForTimeout(2000)
     await expect(this.editor).toBeVisible()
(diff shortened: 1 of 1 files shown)


rohitg00:
Hey @isksamiul, sorry for the late reply. I thought I commented here, but it seems I just added @ytallo as a reviewer.

Please bear with us until Ytallo approves this; we can definitely merge it if it looks good. Thank you for your patience and contributions.

sergiofilhowz:
hi @isksamiul do you mind rebasing your branch with latest from main? I noticed the tests are not passing and want to make sure this is not causing regression to anything

PR #694, opened 2025-09-15
fix: definitive python builder

sergiofilhowz:
- Using python AST to trace all imports
- traverse on all python files
- create separate requirements.txt file for each step and to router (only the ones referenced in code will be there)
- removed support of `.step.py` files since they're not supported by Python imports
- added automated tests for most use-cases

### DIFF
--- a/packages/snap/src/cloud/build/builders/python/index.ts
+++ b/packages/snap/src/cloud/build/builders/python/index.ts
@@ -136,14 +136,17 @@ export class PythonBuilder implements StepBuilder {
 
   private createRouterTemplate(steps: Step<ApiRouteConfig>[]): string {
     const imports = steps
-      .map((step, index) => `from ${this.getModuleName(step)} import handler as route${index}_module`)
+      .map(
+        (step, index) =>
+          `from ${this.getModuleName(step)} import handler as route${index}_handler, config as route${index}_config`,
+      )
       .join('\n')
 
     const routerPaths = steps
       .map((step, index) => {
         const method = step.config.method.toUpperCase()
         const path = step.config.path
-        return `    '${method} ${path}': RouterPath('${step.config.name}', '${step.config.method.toLowerCase()}', route${index}_module.handler, route${index}_module.config)`
+        return `    '${method} ${path}': RouterPath('${step.config.name}', '${step.config.method.toLowerCase()}', route${index}_handler, route${index}_config)`
       })
       .join(',\n')
 
--- a/packages/snap/src/cloud/build/builders/python/router_template.py
+++ b/packages/snap/src/cloud/build/builders/python/router_template.py
@@ -1,6 +1,6 @@
 from typing import Dict, Callable, Any, Literal
+# from steps.api_step import handler as route0_handler, config as route0_config
 # {{imports}}
-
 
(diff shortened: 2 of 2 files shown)


packages/snap/src/generate-locked-data.ts
```diff
@@ -10,19 +10,24 @@ const version = `${randomUUID()}:${Math.floor(Date.now() / 1000)}`
 export const getStepFiles = (projectDir: string): string[] => {
   const stepsDir = path.join(projectDir, 'steps')
   return [
-    ...globSync('**/*.step.{ts,js,py,rb}', { absolute: true, cwd: stepsDir }),
+    ...globSync('**/*.step.{ts,js,rb}', { absolute: true, cwd: stepsDir }),
```
ytallo, line 13: waas it intentionally?

packages/snap/src/generate-locked-data.ts
```diff
@@ -10,19 +10,24 @@ const version = `${randomUUID()}:${Math.floor(Date.now() / 1000)}`
...
     ...globSync('**/*_step.{ts,js,py,rb}', { absolute: true, cwd: stepsDir }),
   ]
 }
 
+export const getStreamFiles = (projectDir: string): string[] => {
+  const stepsDir = path.join(projectDir, 'steps')
+  return [
+    ...globSync('**/*.stream.{ts,js,rb}', { absolute: true, cwd: stepsDir }),
+    ...globSync('**/*_stream.{ts,js,py,rb}', { absolute: true, cwd: stepsDir }),
+  ]
```
ytallo, line 23: there's a difference between the previous one and this one, was it wrong on this place?

packages/snap/src/cloud/build/builders/python/python-data/read-requirements.ts
```diff
@@ -0,0 +1,37 @@
...
+  for (const line of lines) {
+    const trimmedLine = line.trim()
+
+    // Skip empty lines and comments
+    if (!trimmedLine || trimmedLine.startsWith('#')) {
+      continue
+    }
+
+    // Extract package name (everything before version specifiers)
+    const packageMatch = trimmedLine.match(/^([a-zA-Z0-9_-]+)/)
```
ytallo, line 29: what happens if the user put just the dependency name without version?

packages/snap/src/generate-locked-data.ts
sergiofilhowz: just unified it

packages/snap/src/generate-locked-data.ts
sergiofilhowz: i don't want to support `.step.py` because router file wouldn't work

packages/snap/src/cloud/build/builders/python/python-data/read-requirements.ts
sergiofilhowz: it works because this regexp doesn't have $ in the end, which means the pattern won't be to the en do f the line

sergiofilhowz: pushed a test case for it, confirmed it works

PR #720, opened 2025-09-22
fix: python multi module builder

sergiofilhowz:
## The problem

Pymongo has multi module inside, so you can do one `pip install pymongo` and use up to 3 modules that comes with it, which are: `bson`, `pymongo`, and `gridfs`

image

Inside pymongo definition, has all the files developers can import, and it was retrieving only the first one, which is bson

image

## The fix

We're now mapping for each library in requirements.txt, all the dependencies it supports.

```typescript
test('resolves pymongo correctly', () => {
  const depNames = ['pymongo']
  const result = resolveDepNames(depNames, sitePackagesDir)

  expect(result).toEqual({
    pymongo: 'pymongo',
    bson: 'pymongo',
    gridfs: 'pymongo',
  })
})
```

So whenever we import `bson`, `gridfs` or `pymongo`, it will resolve to pymongo in the `requirements.txt`


ytallo:
How is the new resolveDepNames implementation handling multi-module packages?

PR #751, opened 2025-10-01
fix(python-runner): compute full module path so relative imports work

riturajFi:
## Problem
The Python step runner was incorrectly setting `module.__package__` to only the last directory name (e.g. `"endpoints"`).  
As a result, relative imports like:

```python
from ..repositories import db_repo
````

failed with:

```
ImportError: attempted relative import beyond top-level package
```

This prevented Motia from correctly loading step files with nested package structures.

---

## Solution

* Updated the runner logic in `get-config.py` to compute the **full dotted module path** relative to the `steps/` directory.

  * Example:

    * File: `steps/endpoints/python_api_step.py`
    * Old package: `"endpoints"`
    * New package: `"steps.endpoints"`
    * Full module name: `"steps.endpoints.python_api_step"`
* Set `module.__package__` and `sys.modules[...]` to this full dotted name before executing the module.
* This ensures Python has the correct package context, so relative imports always resolve properly.

---

## Benefits

* Relative imports (`..repositories`, `..utils`, etc.) now work across all nested step files.
* No need to rewrite individual step files.
* Future-proof: supports arbitrary nesting depth under the `steps/` directory.
* Compatible with namespace packages (no `__init__.py` required).

---

## Example

**Before (broken):**

```
module.__package__ = "endpoints"
ImportError: attempted relative import beyond top-level package
```

**After (fixed):**

```
module.__package__ = "steps.endpoints"
Relative import '..repositories' resolves correctly
```

---

### DIFF
--- a/packages/core/src/python/get-config.py
+++ b/packages/core/src/python/get-config.py
@@ -3,6 +3,7 @@ import json
 import importlib.util
 import os
 import platform
+from pathlib import Path
 
 def sendMessage(text):
     'sends a Node IPC message to parent proccess'
@@ -35,7 +36,10 @@ async def run_python_module(file_path: str) -> None:
             raise ImportError(f"Could not load module from {file_path}")
             
         module = importlib.util.module_from_spec(spec)
-        module.__package__ = os.path.basename(module_dir)
+        flows_dir = Path(file_path).parents[1]
+        rel_parts = Path(file_path).relative_to(flows_dir).with_suffix("").parts
+        module_name = flows_dir.name + "." + ".".join(rel_parts)
+        module.__package__ = module_name.rsplit(".", 1)[0]
         spec.loader.exec_module(module)
 
         if not hasattr(module, 'config'):

### CODE REVIEW

**Comment by rohitg00** [comment]
Thanks @riturajFi, let me check with the team to review this quickly, thanks!

**Comment by sergiofilhowz** [comment]
I'm trying to test it and doesn't seem to work, am I doing anything wrong?

Captura de Tela 2025-10-13 às 15 03 39

`api2_step.py`

```python
from pydantic import BaseModel
from typing import Optional
from ..python.repo import Repo

class PetRequest(BaseModel):
    name: str
    photo_url: str

class FoodOrder(BaseModel):
    id: str
    quantity: int

class RequestBody(BaseModel):
    pet: PetRequest
    food_order: Optional[FoodOrder] = None

config = {
    "type": "api",
    "name": "PythonApiTrigger2",
    "description": "basic-tutorial api trigger",
    "flows": ["python-tutorial"],
    "method": "POST",
    "path": "/python-basic-tutorial2",
    "bodySchema": RequestBody.model_json_schema(),
    "emits": [],
}

async def handler(req, context):
    body = req.get("body", {})
    context.logger.info("Step 01 – Processing API Step", {"body": body})

    repo = Repo()
    context.logger.info("Repo", {"repo": repo.say_hello()})

    return {"status": 200, "body": {"traceId": context.trace_id}}
```

riturajFi:
Checking this asap

riturajFi:
**Hi @sergiofilhowz,**

Here’s the detailed process to reproduce the bug and verify the fix:

---

### 🧪 Steps to Reproduce (on `main` branch)

1. Build the project.

2. Create the following folder structure:

   ```
   /steps
     ├── __init__.py
     ├── /endpoints
     │    ├── __init__.py
     │    └── python_api_step.py
     └── /repositories
          ├── __init__.py
          └── db_repo.py
   ```

3. Add the following contents:

**`python_api_step.py`**

```python
from ..repositories.db_repo import get_data

config = {
    "type": "api",
    "name": "nestedpython",
    "path": "/api",
    "method": "GET",
    "emits": []
}

async def handler(req, ctx):
    return {
        "status": 200,
        "body": get_data()
    }
```

**`db_repo.py`**

```python
def get_data():
    return 1
```

4. Run the project.
   You’ll see this error:

   ```
   Error running Python module: attempted relative import beyond top-level package
   ```

---

### ✅ Verification (on feature branch)

After switching to this feature branch and repeating the same steps, the project runs successfully:
You can also hit the api and check the successfull implementation of relative imports in python steps - 
```
➜ [INFO] Activating Python environment 
➜ [INFO] Using Python <projectDir>/python_modules/bin/python
➜ [INFO] Site-packages path <projectDir>/python_modules/lib/python3.13/site-packages
➜ [REGISTERED] Step (API) steps/endpoints/python_api_step.py registered
[motia-plugins] ✓ Validated 3 plugin(s) successfully
[motia-plugins] Initialized with 3 plugin(s)
[motia-plugins] Dev server configured, HMR enabled
[motia-plugins] Build started
🚀 Server ready and listening on port 3000
🔗 Open http://localhost:3000 to open workbench 🛠️
```

sergiofilhowz:
Thanks for providing this information, however, the issue happens on my project, we need to ensure it doesn't happen.

image

packages/core/src/python/get-config.py
```diff
@@ -21,21 +22,27 @@ def sendMessage(text):
...
-        if module_dir not in sys.path:
-            sys.path.insert(0, module_dir)
-            
-        flows_dir = os.path.dirname(module_dir)
-        if flows_dir not in sys.path:
-            sys.path.insert(0, flows_dir)
-
-        spec = importlib.util.spec_from_file_location("dynamic_module", file_path)
+        path = Path(file_path).resolve()
+        steps_dir = next((p for p in path.parents if p.name == "steps"), None)
```
ytallo, line 26: We’re implementing support for steps to be located in the src directory. It will be released this week.

packages/snap/motia.config.ts
```diff
@@ -0,0 +1,10 @@
+import { config } from '@motiadev/core'
+
+const statesPlugin = require('@motiadev/plugin-states/plugin')
+const endpointPlugin = require('@motiadev/plugin-endpoint/plugin')
+const logsPlugin = require('@motiadev/plugin-logs/plugin')
+const observabilityPlugin = require('@motiadev/plugin-observability/plugin')
+
+export default config({
```
ytallo, line 8: Is this file necessary?

packages/snap/types.d.ts
```diff
@@ -0,0 +1,17 @@
+/**
```
ytallo, line 1: Is this file necessary?

packages/snap/motia.config.ts
riturajFi: Sorry i placed these by mistake

packages/core/src/python/get-config.py
riturajFi: Hi @ytallo , I think it would be better to implement this after the release then. But is the approach right or does it need some changes?

packages/core/src/python/python-runner.py
```diff
@@ -22,19 +23,27 @@ def parse_args(arg: str) -> Dict:
...
-                sys.path.insert(0, path)
+        path = Path(file_path).resolve()
+        steps_dir = next((p for p in path.parents if p.name == "steps"), None)
+        if steps_dir is None:
+            raise RuntimeError("Could not find 'steps' directory in path")
+
+        project_root = steps_dir.parent
+        project_parent = project_root.parent
+        if str(project_parent) not in sys.path:
+            sys.path.insert(0, str(project_parent))
```
guibeira, line 34: Do we need to include it at position 0? This will require a bit shift of all elements. If it's not necessary, can we use `append`?

PR #752, opened 2025-10-01
feat: allow users to override the python version when installing through motia

rodrigomf24:
## Why?

Currently `motia install` defaults python to 3.13. This can create conflicts for some dependencies that do not support python 3.13 (i.e.: crewai + onnxruntime).

## What?

Since the install script already allows for specifying a python version, I am adding changes to the cli to allow for an argument to inject a custom python version.

In addition to the above, I noticed this [issue](https://github.com/MotiaDev/motia/issues/753), I am proposing a change to allow skipping the platform enforcements in order to unblock docker deployed motia projects

## Note:

This needs to be tested for cloud deployment, I had a chat with @sergiofilhowz  and he called out that the lambdas are using  python3.13, but at least this change fixes the issue for local development and docker deployed projects.

### DIFF
--- a/packages/snap/src/utils/install-lambda-python-packages.ts
+++ b/packages/snap/src/utils/install-lambda-python-packages.ts
@@ -5,9 +5,14 @@ import { internalLogger } from './internal-logger'
 interface VenvConfig {
   requirementsList: string[]
   isVerbose?: boolean
+  skipPythonPlatform?: boolean
 }
 
-export const installLambdaPythonPackages = ({ isVerbose = false, requirementsList }: VenvConfig): void => {
+export const installLambdaPythonPackages = ({
+  isVerbose = false,
+  requirementsList,
+  skipPythonPlatform = false,
+}: VenvConfig): void => {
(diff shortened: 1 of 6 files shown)


packages/snap/src/utils/install-lambda-python-packages.ts
```diff
@@ -20,7 +25,7 @@ export const installLambdaPythonPackages = ({ isVerbose = false, requirementsLis
 
     try {
       // Install packages to lambda site-packages with platform specification
-      const command = `pip install -r "${requirement}" --target "${sitePackagesPath}" --platform manylinux2014_x86_64 --only-binary=:all: --upgrade --upgrade-strategy only-if-needed`
+      const command = `pip install -r "${requirement}" --target "${sitePackagesPath}" ${!skipPythonPlatform ? '--platform manylinux2014_x86_64 --only-binary=:all: ' : ''}--upgrade --upgrade-strategy only-if-needed`
```
sergiofilhowz, line 28: we can't override this, it needs to be linux for lambda

PR #765, opened 2025-10-03
fix: support non-JSON response types in API routes

kebot:
Currently the API step only supports JSON ouput, if you give it a string or Buffer, it will still return a JSON object.

I guess it also somehow related to:
https://github.com/MotiaDev/motia/issues/670

### DIFF
--- a/packages/core/src/server.ts
+++ b/packages/core/src/server.ts
@@ -192,9 +192,7 @@ export const createServer = (
         res.status(result.status)
 
         // Handle different body types
-        if (Buffer.isBuffer(result.body)) {
-          res.send(result.body)
-        } else if (typeof result.body === 'string') {
+        if (Buffer.isBuffer(result.body) || typeof result.body === 'string') {
           res.send(result.body)
         } else {
           res.json(result.body)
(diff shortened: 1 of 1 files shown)


packages/core/src/server.ts
```diff
@@ -190,7 +190,15 @@ export const createServer = (
...
-        res.json(result.body)
+
+        // Handle different body types
+        if (Buffer.isBuffer(result.body)) {
+          res.send(result.body)
+        } else if (typeof result.body === 'string') {
+          res.send(result.body)
+        } else {
+          res.json(result.body)
+        }
```
sergiofilhowz: We can simplify the logic a little bit here
```suggestion
        if (Buffer.isBuffer(result.body) || typeof result.body === 'string') {
          res.send(result.body)
        } else {
          res.json(result.body)
        }
```

rohitg00 (approved):

rohitg00:
This is ready to merge!

sergiofilhowz (approved):

Tanmay-008:
@sergiofilhowz  “I’ve been working on this issue since last week, so should I put it on PR?”

PR #766, opened 2025-10-03
feat: rfc state streams new operations

sergiofilhowz:
## Improvements to State Management

The current way of working with State and Streams is fine, but we need to add some more features to it.

1. Allow developers to increment/decrement values without fetching first.
2. Allow developers to make multiple updates to State/Streams with a single Transaction and be able to rollback the changes if one of the updates fails.

```typescript
type Options = {
  /**
   * The time to live in seconds
   */
  ttl?: number
}

type StateManager = {
  // Existing methods will continue to work as they do now
  get<T>(groupId: string, key: string): Promise<T | null>
  delete<T>(groupId: string, key: string): Promise<T | null>
  getGroup<T>(groupId: string): Promise<T[]>
  clear(groupId: string): Promise<void>

  set<T>(groupId: string, key: string, value: T, options?: Options): Promise<T>

  /**
   * Updates a single item in the state
   *
   * @param groupId - The group id of the state
   * @param key - The key of the item to update
   * @param fields - The fields to update, only the fields that are provided will be updated
   * @returns The item
   */
  updateFields<T>(groupId: string, key: string, fields: Partial<T>): Promise<T>

  // New methods will be added

  /**
   * Increments a single property in the state
   *
   * @param groupId - The group id of the state
   * @param key - The key of the item to increment
   * @param property - The property to increment
   * @param value - The value to increment
   * @returns The item
   */
  incrementNumber<T>(groupId: string, key: string, property: string, value: number): Promise<T>

  /**
   * Decrements a single property in the state
   *
   * @param groupId - The group id of the state
   * @param key - The key of the item to decrement
   * @param property - The property to decrement
   * @param value - The value to decrement
   * @returns The item
   */
  decrementNumber<T>(groupId: string, key: string, property: string, value: number): Promise<T>
}
```

## Increment/Decrement

Usage

```typescript
type StateType = {
  name: string
  count: number
}

const groupId = 'my-group'
const key = 'my-key'

await state.set<StateType>(groupId, key, { name: 'John', count: 1 })

await state.incrementNumber<StateType>(groupId, key, 'count', 2)
await state.decrementNumber<StateType>(groupId, key, 'count', 1)
```

This means that there's no need to fetch the state first, and then update it, we can just update it directly.

## Set TTL to Items

```typescript
type StateManager = {
  /**
   * Sets a single item in the state
   *
   * @param groupId - The group id of the state
   * @param key - The key of the item to set
   * @param value - The value of the item to set
   * @param options - The options for the item
   * @returns The item
   */
  set<T>(groupId: string, key: string, value: T, options?: Options): Promise<T>
}
```

### Usage

```typescript
type StateType = {
  name: string
  count: number
}

const groupId = 'my-group'
const key = 'my-key'
const value = { name: 'John', count: 1 }

// This will set the item with a TTL of 3600 seconds (1 hour)
await state.set<StateType>(groupId, key, value, { ttl: 3600 })
```

## Updating subset of fields

```typescript
type StateType = {
  name: string
  count: number
}

const groupId = 'my-group'
const key = 'my-key'

// This will update only the name field, count will remain the same
await state.updateFields<StateType>(groupId, key, { name: 'New name' })
```

## Streams

Same logic should be applied to Streams: increment/decrement, set TTL to items.

### DIFF
--- a/contributors/rfc/2025-10-03-improvements-to-state.md
+++ b/contributors/rfc/2025-10-03-improvements-to-state.md
@@ -5,9 +5,14 @@ The current way of working with State and Streams is fine, but we need to add so
 1. Allow developers to increment/decrement values without fetching first.
 2. Allow developers to make multiple updates to State/Streams with a single Transaction and be able to rollback the changes if one of the updates fails.
 
-## Increment/Decrement
-
 ```typescript
+type Options = {
+  /**
+   * The time to live in seconds
+   */
+  ttl?: number
+}
+
 type StateManager = {
   // Existing methods will continue to work as they do now
   get<T>(groupId: string, key: string): Promise<T | null>
@@ -15,16 +20,47 @@ type StateManager = {
   getGroup<T>(groupId: string): Promise<T[]>
   clear(groupId: string): Promise<void>
 
-  set<T>(groupId: string, key: string, value: T): Promise<T>
+  set<T>(groupId: string, key: string, value: T, options?: Options): Promise<T>
+
+  /**
+   * Updates a single item in the state
+   *
+   * @param groupId - The group id of the state
+   * @param key - The key of the item to update
+   * @param fields - The fields to update, only the fields that are provided will be updated
+   * @returns The item
+   */
+  updateFields<T>(groupId: string, key: string, fields: Partial<T>): Promise<T>
 
   // New methods will be added
 
+  /**
+   * Increments a single property in the state
+   *
+   * @param groupId - The group id of the state
+   * @param key - The key of the item to increment
+   * @param property - The property to increment
+   * @param value - The value to increment
+   * @returns The item
+   */
   incrementNumber<T>(groupId: string, key: string, property: string, value: number): Promise<T>
+
+  /**
+   * Decrements a single property in the state
+   *
+   * @param groupId - The group id of the state
+   * @param key - The key of the item to decrement
+   * @param property - The property to decrement
+   * @param value - The value to decrement
+   * @returns The item
+   */
   decrementNumber<T>(groupId: string, key: string, property: string, value: number): Promise<T>
 }
 ```
 
-### Usage
+## Increment/Decrement
+
+Usage
 
 ```typescript
 type StateType = {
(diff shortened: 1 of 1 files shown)


mfpiccolo:
I think we need to think about atomic state and state setting/getting more wholistically. Only adding increment/decrement is too narrow of use-case for this.

```
      case 'state.get':
        return this.stateAdapter.get(traceId, key) as T

      case 'state.set':
        return this.stateAdapter.set(traceId, key, value) as T

      case 'state.delete':
        return this.stateAdapter.delete(traceId, key) as T

      case 'state.clear':
        await this.stateAdapter.clear(traceId)
        return undefined as T

      case 'state.getGroup':
        return this.stateAdapter.getGroup(params.groupId) as T

      case 'state.update':
        const updateFn = new Function('current', `return (${params.updateFn})(current)`) as (current: any) => any
        return this.stateAdapter.update(traceId, key, updateFn) as T

      case 'state.increment':
        return this.stateAdapter.increment(traceId, key, delta) as T

      case 'state.decrement':
        return this.stateAdapter.decrement(traceId, key, delta) as T

      case 'state.compareAndSwap':
        return this.stateAdapter.compareAndSwap(traceId, key, expected, newValue) as T

      case 'state.push':
        return this.stateAdapter.push(traceId, key, ...items) as T

      case 'state.pop':
        return this.stateAdapter.pop(traceId, key) as T

      case 'state.shift':
        return this.stateAdapter.shift(traceId, key) as T

      case 'state.unshift':
        return this.stateAdapter.unshift(traceId, key, ...items) as T

      case 'state.setField':
        return this.stateAdapter.setField(traceId, key, field as any, value) as T

      case 'state.deleteField':
        return this.stateAdapter.deleteField(traceId, key, field) as T

      case 'state.transaction':
        return this.stateAdapter.transaction(traceId, operations) as T

      case 'state.batch':
        return this.stateAdapter.batch(traceId, operations) as T

      case 'state.exists':
        return this.stateAdapter.exists(traceId, key) as T
```

mfpiccolo:
Perhaps we can support both.

```
      case 'state.set':
        return this.stateAdapter.set(traceId, key, value) as T
      
      case 'state.atomicSet':
        return this.stateAdapter.atomicSet(traceId, key, value) as T
```

PR #775, opened 2025-10-05
docs: document Zod default values bug in bodySchema

ProdByBuddha:
Document bug where Zod schema defaults defined in bodySchema are not applied to req.body when fields are omitted.

Adds:
- Bug demonstration step showing undefined values
- Fixed step showing JavaScript default parameter workaround
- Comprehensive documentation with test cases

Issue: When using .default() in Zod bodySchema, those defaults don't get applied to req.body. Fields arrive as undefined instead.

Workaround: Use explicit JavaScript default parameters in
destructuring: const { field = 'default' } = req.body

This affects any API step expecting Zod defaults to work.

### DIFF
--- a/playground/steps/bugfix-examples/README.md
+++ b/playground/steps/bugfix-examples/README.md
@@ -1,0 +1,84 @@
+# Bug: Zod Default Values Not Applied to req.body
+
+## Issue Description
+
+When using Zod schemas with `.default()` values in `bodySchema`, those defaults are **not applied** to `req.body` when fields are omitted from the request.
+
+### Expected Behavior
+
+```typescript
+bodySchema: z.object({
+  message: z.string(),
+  priority: z.string().default('normal'),
+  retryCount: z.number().default(3),
+})
+```
+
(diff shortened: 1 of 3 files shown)


ProdByBuddha:
export const handler: Handlers['ZodDefaultsFixed'] = async (req, { logger }) => {
    // WORKAROUND: Apply defaults manually in JavaScript destructuring
    // This works around the bug where Zod bodySchema defaults aren't applied
    const {
      message,
      priority = 'normal',      // JS default parameter
      retryCount = 3,            // JS default parameter  
      enabled = true             // JS default parameter
    } = req.body

sergiofilhowz:
Hi @ProdByBuddha, you're right, we're not using zod for any type of validation or object assignment, we don't want to be an opinionated framework that has all of these logics inside, instead, we use zod only for documentation and types generation.

But it's pretty easy to do it on your apps, by just using `schema.parse(req.body)` this does validation and object assignment. This could also be built on a shared middleware

ProdByBuddha:
Thank you for the clarification, I'll close this out and proceed accordingly.

PR #776, opened 2025-10-06
Add Large Payload Support via Temporary File Transport in call-step-file and node-runner

riturajFi:
Overview

This PR introduces a robust solution to handle large request payloads (JSON or otherwise) without changing existing data types or introducing environment flags.
Previously, Motia serialized the entire request body into a single JSON string argument passed via argv to the runner. This caused OS-level E2BIG errors for large payloads (typically > 2–8MB).

The new design uses temporary files to transfer large data between the main process and language runners (Node, Python, Ruby), while preserving backward compatibility for smaller payloads.

Key Changes
🧠 call-step-file.ts

Added detection for payload size before spawning the runner.

When data exceeds a threshold (e.g., >1MB):

Writes the full event object to a secure temporary JSON file (in os.tmpdir()).

Passes only the path of this file to the runner via argv.

On completion or error, the temp directory is safely deleted to prevent disk bloat.

Maintains existing behavior for smaller requests to minimize overhead.

⚙️ node-runner.ts

Detects whether the received argument is:

A file path → loads JSON from file, executes handler, and deletes the file.

A JSON string → uses legacy flow (unchanged behavior).

Adds cleanup for temporary files even if the process is interrupted.

Maintains identical interface for user handlers — no breaking changes.

Benefits

✅ Removes OS argument size limitations (supports 100MB+ payloads).
✅ Preserves existing APIs, handlers, and request/response structures.
✅ Reduces memory duplication from repeated JSON serialization.
✅ Ensures secure, auto-cleaned temp directories (mode 0700, file mode 0600).
✅ Fully backward compatible and cross-language ready.

Testing Performed

Verified:

Small payloads (<1MB) → use legacy in-memory mode.

Large JSON payloads (5MB–200MB) → run successfully without E2BIG.

Automatic cleanup of temp files after step completion.

Regression-tested standard steps and event flows.

Next Steps

Add optional streaming support for extremely large binary bodies.

Integrate with multipart/form-data in future versions.

🎥 Demo:

Before -

[Screencast from 2025-10-06 10-33-33.webm](https://github.com/user-attachments/assets/10632123-54cf-4f91-8025-d9dd8b5f2781)

After -

[Screencast from 2025-10-06 10-39-10.webm](https://github.com/user-attachments/assets/3c467fa7-8125-4169-9d10-4b3870972041)

### DIFF
--- a/packages/core/src/call-step-file.ts
+++ b/packages/core/src/call-step-file.ts
@@ -8,6 +8,8 @@ import { isAllowedToEmit } from './utils'
 import { Logger } from './logger'
 import { Tracer } from './observability'
 import { TraceError } from './observability/types'
+import os from 'os'
+import fs from 'fs'
 
 type StateGetInput = { traceId: string; key: string }
 type StateSetInput = { traceId: string; key: string; value: unknown }
@@ -57,6 +59,8 @@ type CallStepFileOptions = {
   tracer: Tracer
 }
 
+const THRESHOLD_BYTES = 1 * 1024 * 1024 // 1 MB
+
 export const callStepFile = <TData>(options: CallStepFileOptions, motia: Motia): Promise<TData | undefined> => {
   const { step, traceId, data, tracer, logger, contextInFirstArg = false } = options
 
@@ -66,16 +70,39 @@ export const callStepFile = <TData>(options: CallStepFileOptions, motia: Motia):
     const streamConfig = motia.lockedData.getStreams()
     const streams = Object.keys(streamConfig).map((name) => ({ name }))
     const jsonData = JSON.stringify({ data, flows, traceId, contextInFirstArg, streams })
+    const jsonBytes = Buffer.byteLength(jsonData, 'utf8')
+
+    // Default: keep old behavior (argv carries inline JSON)
+    let argvPayload = jsonData
+    let tempDir: string | undefined
+    let metaPath: string | undefined
+
+    // If payload is large, write it to a temp file and pass the path instead
+    if (jsonBytes >= THRESHOLD_BYTES) {
+      tempDir = fs.mkdtempSync(path.join(os.tmpdir(), 'motia-'))
+      metaPath = path.join(tempDir, 'meta.json')
+      fs.writeFileSync(metaPath, jsonData, { mode: 0o600 })
+      argvPayload = metaPath
+    }
+
     const { runner, command, args } = getLanguageBasedRunner(step.filePath)
     let result: TData | undefined
 
     const processManager = new ProcessManager({
       command,
-      args: [...args, runner, step.filePath, jsonData],
+      args: [...args, runner, step.filePath, argvPayload],
       logger,
       context: 'StepExecution',
       projectRoot: motia.lockedData.baseDir,
     })
+
+    const cleanupTemp = () => {
+      if (!tempDir) return
+      try {
+        fs.rmSync(tempDir, { recursive: true, force: true })
+      } catch {}
+      tempDir = undefined
+    }
 
     trackEvent('step_execution_started', {
       stepName: step.config.name,
@@ -89,6 +116,7 @@ export const callStepFile = <TData>(options: CallStepFileOptions, motia: Motia):
       .then(() => {
         processManager.handler<TraceError | undefined>('close', async (err) => {
           processManager.kill()
+          cleanupTemp()
 
           if (err) {
             trackEvent('step_execution_error', {
@@ -193,6 +221,7 @@ export const callStepFile = <TData>(options: CallStepFileOptions, motia: Motia):
 
         processManager.onProcessClose((code) => {
           processManager.close()
+          cleanupTemp()
 
           if (code !== 0 && code !== null) {
             const error = { message: `Process exited with code ${code}`, code }
@@ -207,6 +236,7 @@ export const callStepFile = <TData>(options: CallStepFileOptions, motia: Motia):
 
         processManager.onProcessError((error) => {
           processManager.close()
+          cleanupTemp()
           tracer.end({
             message: error.message,
             code: error.code,
@@ -227,6 +257,10 @@ export const callStepFile = <TData>(options: CallStepFileOptions, motia: Motia):
         })
       })
       .catch((error) => {
+        try {
+          // spawn failed before handlers attached — still clean up
+          if (tempDir) fs.rmSync(tempDir, { recursive: true, force: true })
+        } catch {}
         tracer.end({
           message: error.message,
           code: error.code,
--- a/packages/core/src/node/node-runner.ts
+++ b/packages/core/src/node/node-runner.ts
@@ -1,4 +1,5 @@
 import path from 'path'
+import fs from 'fs'
 import { StateStreamEvent, StateStreamEventChannel, StreamConfig } from '../types-stream'
 import { Logger } from './logger'
 import { composeMiddleware } from './middleware-compose'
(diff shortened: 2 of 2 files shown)


packages/core/src/call-step-file.ts
```diff
@@ -227,6 +257,10 @@ export const callStepFile = <TData>(options: CallStepFileOptions, motia: Motia):
         })
       })
       .catch((error) => {
+        try {
+          // spawn failed before handlers attached — still clean up
+          if (tempDir) fs.rmSync(tempDir, { recursive: true, force: true })
```
sergiofilhowz: this can lead to concurrency issues, one process deleting the temp dir with files of other processes

packages/core/src/call-step-file.ts
```diff
@@ -66,17 +70,40 @@ export const callStepFile = <TData>(options: CallStepFileOptions, motia: Motia):
...
+      args: [...args, runner, step.filePath, argvPayload],
       logger,
       context: 'StepExecution',
       projectRoot: motia.lockedData.baseDir,
     })
 
+    const cleanupTemp = () => {
+      if (!tempDir) return
+      try {
+        fs.rmSync(tempDir, { recursive: true, force: true })
```
sergiofilhowz: we need to make it async, we're having some issues with synchronous methods already that I need to fix

packages/core/src/node/node-runner.ts
```diff
@@ -15,11 +16,29 @@ require('ts-node').register({
   compilerOptions: { module: 'commonjs' },
 })
 
-function parseArgs(arg: string) {
+function parseArgs(arg?: string) {
```
sergiofilhowz, line 18: this needs to be implemented in python and ruby as well

sergiofilhowz: we need an automated test to validate it in all languages

riturajFi:
Thanks @sergiofilhowz for your comments! I just want to validate whether the approach is correct. I shall implement this for all the runners (+ write an implementation prompt for future runners)
And write automted tests.

riturajFi:
Hi @sergiofilhowz , implemented the changes you asked for - 

1. made the file deletion async
2. made the cleaning step idempotent to prevent race condition of two processes deleting and accessing the same file
3. wrote tests for the feature

sergiofilhowz:
This is good! I'm going to merge soon

sergiofilhowz:
@riturajFi thanks for the PR, I want to merge it, but first we need to fix the pipeline issues

packages/core/src/node/node-runner.ts
riturajFi: This is done

packages/core/src/node/node-runner.ts
```diff
@@ -15,11 +16,22 @@
...
+  // If there's no arg, keep behavior simple
+  if (!arg) return { data: null }
+
   try {
+    // New big-payload path: if arg is a path to a file, read & parse it
+    if (fs.existsSync(arg) && fs.statSync(arg).isFile()) {
+      const text = fs.readFileSync(arg, 'utf8')
```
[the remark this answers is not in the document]
riturajFi: it's alright

riturajFi:
@sergiofilhowz can you please review this? I fixed some minor concerns that were there. Can we merge it now?

packages/core/src/call-step-file.ts
```diff
@@ -214,9 +251,10 @@ export const callStepFile = <TData>(options: CallStepFileOptions, motia: Motia):
 
         processManager.onStderr((data) => logger.error(Buffer.from(data).toString()))
 
-        processManager.onProcessClose((code) => {
+        processManager.onProcessClose(async (code) => {
           if (timeoutId) clearTimeout(timeoutId)
           processManager.close()
+          await cleanupTemp()
```
sergiofilhowz, line 303: this will lead to race conditions:
1. When two requests happen at the same time, one can delete the folder while the other tries to read it

packages/core/src/call-step-file.ts
```diff
@@ -67,18 +71,51 @@ export const callStepFile = <TData>(options: CallStepFileOptions, motia: Motia):
...
+      isCleaning = true
+
+      try {
+        await fs.promises.rm(dir, { recursive: true, force: true })
+      } catch {}
+
+      isCleaning = false
+      isCleaned = true
+      
+    }
```
sergiofilhowz, line 163: it needs to delete only the file that was created for the process, otherwise there will be race conditions

packages/core/src/call-step-file.ts
riturajFi: Hi @sergiofilhowz , each file is uniquely created for a process so there is no scope of process 1 accessing file of process 2.

packages/core/src/call-step-file.ts
```diff
@@ -67,18 +72,55 @@ export const callStepFile = <TData>(options: CallStepFileOptions, motia: Motia):
...
+    let isCleaning: boolean = false
+
+    // If payload is large, write it to a temp file and pass the path instead
+    if (jsonBytes >= THRESHOLD_BYTES) {
+      // Create a unique, process-specific temp directory
+      const uniqueId = `${process.pid}-${crypto.randomBytes(6).toString('hex')}`
+      tempDir = path.join(os.tmpdir(), `motia-${uniqueId}`)
+      fs.mkdirSync(tempDir, { recursive: true, mode: 0o700 })
+      metaPath = path.join(tempDir, 'meta.json')
+      fs.writeFileSync(metaPath, jsonData, { mode: 0o600 })
```
sergiofilhowz: why not create a single file instead?

packages/core/src/call-step-file.ts
```diff
@@ -67,18 +72,55 @@ export const callStepFile = <TData>(options: CallStepFileOptions, motia: Motia):
...
     const jsonData = JSON.stringify({ data, flows, traceId, contextInFirstArg, streams })
+    const jsonBytes = Buffer.byteLength(jsonData, 'utf8')
+
+    // Default: keep old behavior (argv carries inline JSON)
+    let argvPayload = jsonData
+    let tempDir: string | undefined
+    let metaPath: string | undefined
+    // This is for keeping the cleanup idempotent
+    let isCleaned: boolean = false
+    let isCleaning: boolean = false
```
sergiofilhowz: avoid usage of let

packages/core/src/call-step-file.ts
```diff
@@ -252,7 +295,15 @@ export const callStepFile = <TData>(options: CallStepFileOptions, motia: Motia):
         })
       })
       .catch((error) => {
+        // spawn failed before handlers attached — still clean up (async, best-effort)
+        if (tempDir) {
+          const dir = tempDir
+          tempDir = undefined
+          void fs.promises.rm(dir, { recursive: true, force: true })
+          .catch((err) => {logger.debug(`temp cleanup: ${dir}: ${err.message ?? err}`)})
+        }
```
sergiofilhowz, line 346: why not run it at finally?

riturajFi:
Hi @sergiofilhowz , the changes have been implemented according to your suggestion. Can you kindly review this?

rohitg00:
Hey, Does @sergiofilhowz and you djscussed about this issue on discord or are you still waiting for review?

CoderMageFox:
Aany update?

riturajFi:
> Aany update?

waiting till the refactor to rust

anthonyiscoding:
Hey @riturajFi, thank you for this substantial contribution! Adding large payload support via temporary file transport is a really important capability — handling payloads that exceed IPC message size limits is a real-world need that many users will benefit from, and I can see from the discussion that you put a lot of thought and iteration into the approach.

I wanted to give you a heads up that we're currently working on a major architecture change — we're building a new engine that will completely rework how Motia handles inter-process communication between the core and workers. The new engine is built in Rust and introduces a fundamentally new protocol for how data is transported between the engine core and worker processes, including native handling of large payloads with much better performance characteristics.

Because of the scope of these changes, the IPC transport layer (including how call-step-file and node-runner handle data) will be fundamentally different once the new engine is live. I'd love for you to revisit these kinds of contributions at that point — your understanding of file-based transport strategies will be incredibly valuable in stress-testing and improving the new data transport layer.

I really appreciate the time and effort you invested in this PR, and hope you'll stay involved!

To be the first to know when the architecture change is released please add your email to this form: https://forms.gle/24iCHL9yAk1i6LDc6

PR #801, opened 2025-10-10
chore(docs): states doc updated

rohitg00:
I'm unsure about adding the `Storage Options` section; please review and suggest changes.

### DIFF
--- a/packages/workbench/src/components/header/deploy-button.tsx
+++ b/packages/workbench/src/components/header/deploy-button.tsx
@@ -27,7 +27,7 @@ export const DeployButton = () => {
 
   const onSelfHostedClick = () => {
     analytics.track('deploy_button_self_hosted_clicked')
-    window.open('https://www.motia.dev/docs/concepts/deployment/self-hosted', '_blank')
+    window.open('https://www.motia.dev/docs/deployment-guide/self-hosted', '_blank')
   }
 
   return (


packages/docs/content/docs/development-guide/state-management.mdx
```diff
@@ -1,448 +1,418 @@
...
 state:
   adapter: redis
-  host: localhost # Redis server host (e.g., 'localhost' or IP address)
-  port: 6379 # Redis server port (default: 6379)
-  password: optional # Redis password (if required)
-  ttl: 3600 # Optional: State Time-To-Live in seconds (e.g., 3600 seconds = 1 hour)
+  host: localhost
+  port: 6379
+  password: optional
+  ttl: 3600  # Auto-expire after 1 hour (optional)
```
ytallo: we don't have adapters yet, let's remove it

rohitg00: That's what I'm unsure about: should we keep it or not as users can create [them] if they want, right in future?

ytallo: once we don't have support for this configuration, it could confuse users.

rohitg00: sure, removed @ytallo

PR #819, opened 2025-10-13
feat(cli): make create folder optional with interactive prompt

### LINKED ISSUE #817: Improvement: npx motia create with folder as second argument
Right now it's 
```
npx motia create --name folder-name
# or
npx motia create --n folder-name
```

We want to change it to be 

```
npx motia create folder-name
```

The folder name should be optional and prompted when not provided.

## Make sure to
- [ ] Update docs with this new change
- [ ] update `packages/e2e/scripts/pr-global-setup.ts` to not use -n
- [ ] update `packages/e2e/scripts/global-setup.ts` to not use -n

Tanmay-008:
## Summary

Previously, the npx motia create command required the user to pass the folder name via the --name or -n option.
This was not very user-friendly:

This PR makes the folder name optional:

If the folder name is not provided via CLI, the user is prompted to enter a name interactively.
Removes the --name / -n option usage from scripts and updates documentation to reflect the new behavior.

## Related Issues
Closes #817 

## Type of Change
- [ ] Bug fix
- [ .] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist
- [. ] I have read the [CONTRIBUTING.md](https://github.com/MotiaDev/motia/blob/main/CONTRIBUTING.md)
- [. ] My code follows the code style of this project
- [. ] I have added tests where applicable
- [ .] I have tested my changes locally
- [. ] I have linked relevant issues
- [. ] I have added screenshots for UI changes (if applicable)

### DIFF
--- a/packages/snap/src/cli.ts
+++ b/packages/snap/src/cli.ts
@@ -23,13 +23,13 @@ program
   })
 
 program
-  .command('create [folder]')
+  .command('create [project name]')
   .description('Create a new motia project')
   .option('-t, --template <template>', 'The template to use for your project')
   .option('-i, --interactive', 'Use interactive prompts to create project') // it's default
   .option('-c, --confirm', 'Confirm the project creation', false)
-  .action((folder, options) => {
-    const mergedArgs = { ...options, name: folder };
+  .action((projectName, options) => {
+    const mergedArgs = { ...options, name: projectName };
     return handler(async (arg, context) => {
       const { createInteractive } = require('./create/interactive');
       await createInteractive(
--- a/packages/docs/content/docs/development-guide/cli.mdx
+++ b/packages/docs/content/docs/development-guide/cli.mdx
@@ -23,7 +23,7 @@ npx motia@latest create [options]
 
 Options:
 
-- `[folder]` (optional): The name for your project folder. 
+- `[project name]` (optional): The name for your project folder. 
   If not provided, you will be prompted to enter it. 
   Use `.` or `./` to create it in the current directory.
 
--- a/packages/docs/public/llms.txt
+++ b/packages/docs/public/llms.txt
@@ -936,7 +936,7 @@ npx motia@latest create [options]
 
 Options:
 
-- `[folder]` (optional): The name for your project folder. 
+- `[project name]` (optional): The name for your project folder. 
   If not provided, you will be prompted to enter it. 
   Use `.` or `./` to create it in the current directory.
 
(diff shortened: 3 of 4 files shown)


packages/snap/src/cli.ts
```diff
@@ -25,20 +25,29 @@ program
...
+          },
+        ])
+        folder = projectName
+      }
       await createInteractive(
         {
-          name: arg.name,
+          name: folder,
           template: arg.template,
           confirm: !!arg.confirm,
```
dev-shetty, line 39: I believe this wont work, in commander the first parameter will always be the argument. 

In this case if someone runs `pnpm motia create my-project --template base`

Then `arg` will have the value "my-project", hence `arg.confirm` or `arg.template` will result into `undefined`

Reference: https://github.com/tj/commander.js/blob/master/examples/thank.js

Tanmay-008: Thanks for the feedback! The issue is fixed:

packages/snap/src/cli.ts
```diff
@@ -25,22 +25,31 @@ program
...
+        folder = projectName
+      }
       await createInteractive(
         {
-          name: arg.name,
-          template: arg.template,
-          confirm: !!arg.confirm,
+          name: folder,
+          template: (context as any).template,
+          confirm: !!(context as any).confirm,
```
dev-shetty: Context doesn't have it as well, you need to change the `handler` function which is being passed to `.action()`. Refer to my previous comment for how commander handles `arguments` and `options`

Please try running `pnpm motia create my-project --template nodejs`

This shouldn't prompt you for the template again since you have already mentioned it

You can check the Discord in the `#contributors` channel i have been discussing the approach

packages/snap/src/cli.ts
```diff
@@ -25,22 +25,31 @@ program
...
+      if (!folder) {
+        const { projectName } = await inquirer.prompt([
+          {
+            type: 'input',
+            name: 'projectName',
+            message: 'Enter project folder name:',
+            default: 'my-motia-app',
+          },
+        ])
+        folder = projectName
```
dev-shetty: You might not need this package, since `createInteractive` has the check if `name` is not present then prompt the user for name

Check the implmentation here `packages/snap/src/create/interactive.ts` at Line 43

Tanmay-008: @dev-shetty  place check

dev-shetty: LGTM, Checked it locally the command seem to work, awesome work Tanmay!

Nit pick, maybe we can name the argument as "Project Name" instead of "Folder"

@sergiofilhowz can have a better opinion here.

Tanmay-008: Thanks! I’ve updated the argument name to `Project Name` as suggested. 👍

Issue #841, opened 2025-10-17
Add warning to invalid payload


sergiofilhowz:
When event steps receives invalid payload, we need to log a warning showing what was invalid.
Let's make it with the same validation message as a invalid step configuration

<img width="521" height="97" alt="Image" src="https://github.com/user-attachments/assets/d3e37bd5-1555-4a9e-adfa-001b32a6d6e8" />

In the screenshot above, it's logging as `ERROR`, but we should log as `WARN`

sidgaikwad:
can you share the payload and api.step.ts file plz @sergiofilhowz

sergiofilhowz:
this use-case would happen in javascript projects where they're not strong typed. You can create one api step that emits to an event step with invalid payload. A warn should appear in the logs

PR #845, opened 2025-10-19
chore: warnings on invalid emit payloads

sidgaikwad:
… event step

## Summary
| Component | Role | Trigger Time | Purpose |
|------------|------|--------------|----------|
| `createHandler` | Registers step handlers dynamically | Startup | Setup event subscriptions |
| `validateEventData` | Validates event payload shape | Runtime | Catch invalid event data |
| `printEventInputValidationError` | Logs validation issues | Runtime | Developer feedback |

## Related Issues
- Closes: #841 

## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist
- [ ] I have read the [CONTRIBUTING.md](https://github.com/MotiaDev/motia/blob/main/CONTRIBUTING.md)
- [ ] My code follows the code style of this project
- [ ] I have added tests where applicable
- [ ] I have tested my changes locally
- [ ] I have linked relevant issues

## Screenshots (if applicable)
image

## Additional Context
# 🧩 Motia Event Step Runtime Validation

This document explains how the **runtime event validation system** works inside the Motia framework, specifically focusing on:

- The `createHandler` logic (automatic validation + event runtime execution)
- The `printEventInputValidationError` logger printer

---

## ⚙️ 1. `createHandler` — Runtime Event Step Setup

The `createHandler` function dynamically subscribes event steps at runtime.

When each `Step<EventConfig>` is initialized, `createHandler` automatically:
- Subscribes all the configured events (`subscribes` array)
- Injects a runtime validation layer that checks incoming event payloads
- Safely invokes the step handler if validation passes

### 🧠 Flow Overview

1. **Step Registration**
   ```ts
   const { config, filePath } = step
   const { subscribes, name } = config

   subscribes.forEach((subscribe) => {
     motia.eventManager.subscribe({
       filePath,
       event: subscribe,
       handlerName: step.config.name,
       handler: async (event) => { ... }
     })
   })
   ```

2. **When Event is Emitted**
   - The subscribed handler is triggered.
   - Data is validated against the `step.config.input` schema (if available).
   - Validation issues (missing, extra, or mismatched types) are printed via the printer.
   - If valid, the actual step handler (`callStepFile`) executes.

### 🧩 Built-in Validation

The function `validateEventData()` performs lightweight validation **without Zod**.  
It checks:

| Validation | Description |
|-------------|--------------|
| **Missing Fields** | Required fields not found in data |
| **Extra Fields** | Data fields not defined in schema |
| **Type Mismatches** | Mismatch between expected and actual data types |

#### Example Output
```
⚠ Warning: Emit pet.created validation issues:
│ ⚠ Missing fields: name
│ ⚠ Type mismatches: "age": expected number, got string
└─ Payload does not match schema.
```

If all checks pass, execution continues with `callStepFile()`.

---

## 🖨️ 2. `printEventInputValidationError` — Console Reporter

The printer provides a clean, developer-friendly console output when event data fails schema validation.

### Example Method

```ts
printEventInputValidationError(
  emit: { topic: string },
  details: { missingFields?: string[]; extraFields?: string[]; typeMismatches?: string[] }
)
```

### DIFF
--- a/packages/core/src/__tests__/server.test.ts
+++ b/packages/core/src/__tests__/server.test.ts
@@ -120,6 +120,6 @@ describe('Server', () => {
       expect(found.status).toBe(200)
 
       await server.close()
-    })
+    }, 20000)
   })
 })


packages/core/src/step-handlers.ts
```diff
@@ -31,12 +31,72 @@ export const createStepHandlers = (motia: Motia): MotiaEventManager => {
...
+            // 4️⃣ Check for type mismatches
+            for (const [field, def] of Object.entries(schemaProps as Record<string, any>)) {
+              if (field in data) {
+                const expectedType = (def as any)?.type
+                const actualType = Array.isArray((data as any)[field]) ? 'array' : typeof (data as any)[field]
+
+                if (expectedType && expectedType !== actualType) {
+                  typeMismatches.push(`"${field}": expected ${expectedType}, got ${actualType}`)
+                }
+              }
```
sergiofilhowz: is it possible to simplify this validation with json schema validation?

packages/core/src/step-handlers.ts
```diff
@@ -31,12 +31,72 @@ export const createStepHandlers = (motia: Motia): MotiaEventManager => {
...
+            )
+
+            if (missingFields.length || extraFields.length || typeMismatches.length) {
+              motia.printer.printEventInputValidationError(
+                { topic: event.topic },
+                { missingFields, extraFields, typeMismatches },
+              )
+
+              logger.error(`❌ Validation failed for event "${step.config.name}"`)
+              return // stop execution if invalid
```
sergiofilhowz: let's just print a warning, we shouldn't block the data to flow.

warning means: something will not work correctly, please do something before an actual error happens

it will be very helpful in cases of errors like accessing field from undefined and so, developers will quickly realize it was a configuration issue

sergiofilhowz (changes requested):
Thank you for fixing the issue. It was a small change, but I think we can use JSON schema validation and make it even smaller :)

sidgaikwad:
hey @sergiofilhowz, done with both the comments used schema-based approach and also No return, flow continues.

package.json
```diff
@@ -51,5 +51,8 @@
   "volta": {
     "node": "20.11.1"
   },
-  "version": "0.0.7"
+  "version": "0.0.7",
+  "dependencies": {
+    "ajv": "^8.17.1"
+  }
```
sergiofilhowz: this should be in `packages/core/package.json`

sidgaikwad: i thought if we want to use it in any package in future i would be easy to have in the global package json right ??

sidgaikwad: @sergiofilhowz done with the changes

packages/core/src/step-handlers.ts
```diff
@@ -45,23 +53,82 @@ export const createStepHandlers = (motia: Motia, queueManager: QueueManager): Mo
...
     }
 
     const queueConfig = getQueueConfigWithDefaults(config.infrastructure)
-
     const handlers: Array<{ topic: string; handler: (event: Event) => Promise<void> }> = []
 
     subscribes.forEach((subscribe) => {
       const handler = async (event: Event) => {
-        const { data, traceId } = event
+        let { data, traceId } = event
```
sergiofilhowz: why changing from const to let?

packages/core/src/step-handlers.ts
```diff
@@ -45,23 +53,82 @@ export const createStepHandlers = (motia: Motia, queueManager: QueueManager): Mo
...
+                        field === '(root)'
+                          ? err.params.additionalProperty
+                          : `${field}.${err.params.additionalProperty}`,
+                      )
+                    }
+                    break
+                  default:
+                    typeMismatches.push(`Field "${field}": ${err.message}`)
+                }
+              }
```
sergiofilhowz: please move this logic to a separate function, I want to avoid this function to become too large and having too many concerns.

package.json
sergiofilhowz: we have some UI packages that we don't want to have this as dependency

packages/core/src/step-handlers.ts
sidgaikwad: changed @sergiofilhowz

packages/core/src/step-handlers.ts
sidgaikwad: @sergiofilhowz done with this comment also

packages/core/src/validate-event-input.ts
```diff
@@ -0,0 +1,70 @@
+// validateEventInput.ts
```
sergiofilhowz: sorry for being nitpicky here, but the flie name pattern is actually `validate-event-input.ts` can you place change it?

(that's one thing we could also add to the contributions guidelines)

sergiofilhowz (dismissed):
Almost there!

packages/core/src/validate-event-input.ts
sidgaikwad: @sergiofilhowz donzo sir

sidgaikwad:
@sergiofilhowz, changes were made for tests and linting and now the quality check is passing

PR #864, opened 2025-10-27
remove features in manifesto

rohitg00:


rohitg00:
updated website, don't need these changes.

PR #941, opened 2025-11-18
Update OG image with new CDN version

rohitg00:
- Replace og-image-updated-new.jpg with latest version from CDN
- Source: image


sergiofilhowz:
btw this is going to show only when sharing docs page

rohitg00:
Yes! Another one already added in webflow.

PR #1008, opened 2025-12-03
feat: queue concurrency infrastructure

ytallo:
## Type of Change
- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist
- [ ] I have read the [CONTRIBUTING.md](https://github.com/MotiaDev/motia/blob/main/CONTRIBUTING.md)
- [ ] My code follows the code style of this project
- [ ] I have added tests where applicable
- [ ] I have tested my changes locally
- [ ] I have linked relevant issues

### DIFF
--- a/packages/core/src/types.ts
+++ b/packages/core/src/types.ts
@@ -41,12 +41,20 @@ export type HandlerConfig = {
   timeout: number
 }
 
-export type QueueConfig = {
-  type: 'fifo' | 'standard'
-  maxRetries: number
-  visibilityTimeout: number
-  delaySeconds: number
-}
+export type QueueConfig =
+  | {
+      type: 'fifo'
+      maxRetries: number
+      visibilityTimeout: number
+      delaySeconds: number
+    }
+  | {
+      type: 'standard'
+      maxRetries: number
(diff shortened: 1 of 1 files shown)


packages/core/src/types.ts
```diff
@@ -41,12 +41,20 @@ export type HandlerConfig = {
...
+      maxRetries: number
+      visibilityTimeout: number
+      delaySeconds: number
+    }
+  | {
+      type: 'standard'
+      maxRetries: number
+      visibilityTimeout: number
+      delaySeconds: number
+      concurrency?: number
```
anthonyiscoding: Are there other attributes related to concurrency that might be configured in the future?

If so can this either be made into an object:

```
concurrency: {
    maxThreads: 10
}
```

or rename `concurrency`:
```
{
    maxRetries: 10,
    maxThreads: 10
}
```

I don't care about the specific name just want to make sure that it's flexible in the future since `concurrency` is a pretty broad word. For example in the future there might be soft/hard limits, burst, etc. and these all kind of fall under `concurrency`.

PR #1113, opened 2025-12-29
Fix Python IPC UTF-8 framing

suncan2014Ai:
## Problem
Python IPC reader decodes each 4096-byte chunk as UTF-8. When a multi-byte character is split across reads, decode fails and logs errors like:
- ERROR: Reading IPC failed: 'utf-8' codec can't decode ...
- WARNING: Failed to parse JSON: Expecting value ...

## Fix
Buffer raw bytes, split on newline boundaries, then decode each complete line before JSON parsing. This avoids UTF-8 boundary issues while preserving existing framing behavior.

## Testing
Not run (log-only change).

### DIFF
--- a/packages/core/src/python/motia_ipc_communication.py
+++ b/packages/core/src/python/motia_ipc_communication.py
@@ -100,7 +100,7 @@ class IpcCommunication:
     async def _read_ipc(self) -> None:
         """Read messages from IPC file descriptor in background"""
         loop = asyncio.get_event_loop()
-        buffer = ""
+        buffer = b""
         
         while self.executing:
             try:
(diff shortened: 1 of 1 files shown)


anthonyiscoding:
Hey @suncan2014Ai, thanks for digging into this! The UTF-8 framing issue with multi-byte characters splitting across read boundaries is a real problem, and your approach of buffering raw bytes before decoding looks like a good way to handle it.

I wanted to let you know that we're currently working on a major architecture change — we're building a new engine that will completely rework how Motia handles inter-process communication between the core and worker processes. The new engine is built in Rust and introduces a new protocol for how workers (including Python workers) communicate with the engine core, which will bring significant performance improvements and more robust data handling.

Because of the scope of these changes, the IPC layer — including how Python workers read and write messages — will be fundamentally different once the new engine is live. I'd love for you to revisit this kind of contribution at that point, as your understanding of encoding edge cases in the IPC layer will be really valuable.

Thank you for the contribution, and hope you have more in the future!

To be the first to know when the architecture change is released please add your email to this form: https://forms.gle/24iCHL9yAk1i6LDc6

Issue #1112, opened 2025-12-26
Missing leading slash in API path causes malformed URLs on workbench


abhinav-m22:
### Environment

OS: Ubuntu Linux, Node version: v20, Browser: Chrome, Motia Version: latest pull

### Steps to Reproduce

1. Create an API step (e.g., missing-slash.step.ts) with the following configuration:
   ```typescript
   export const config = {
       name: 'MissingSlashStep',
       type: 'api',
       path: 'repro-missing-slash', // Missing leading slash here
       method: 'GET',
       description: 'Endpoint with a path missing the leading slash',
       emits: [],
   };
   ```
2. Start the development server using `pnpm run dev`.
3. Open the Workbench UI in your browser.
4. Navigate to the **Endpoints** tab in the sidebar.
5. Select the **MissingSlashStep** from the list.
6. Observe the **URL Preview** field in the right-hand panel.

### Expected Behavior

The Workbench UI should normalize the path or the system should enforce a leading slash during validation. The expected URL preview should be: `http://localhost:3000/repro-missing-slash`

### Actual Behavior

The URL preview is generated as: `http://localhost:3000repro-missing-slash`

### Relevant Logs/Console Output

```shell
N/A - This is a visual/logic defect in the UI component responsible for URL preview.
```

### Screenshots

<img width="1920" height="618" alt="Image" src="https://github.com/user-attachments/assets/20c1514f-a2ae-4c5e-9438-92bf2ae91255" />

### Motia Version

Latest main branch

### Additional Context

The root cause is located in [plugins/plugin-endpoint/src/components/endpoint-path-preview.tsx], where the full URL is constructed by simple string concatenation:
```tsx
const fullUrl = `${baseUrl}${pathUrl}`
```
Since `pathUrl` is derived directly from the step config without ensuring a leading slash, the result is malformed. 

**Proposed Solution:**
Enforce a leading slash in [packages/core/src/step-validator.ts] using Zod's `.startsWith('/')` for API paths.

abhinav-m22:
I’ve proposed a fix and would like to work on it, incorporating any feedback from the maintainers

rohitg00:
@ytallo What do you think about this?

Thanks @abhinav-m22 for the issue, We'll look into this.

iamadityamaurya:
fix: missing leading slash in API path #1117 

I have fixed it and uploaded a pull request.

ytallo:
[...]

PR #1117, opened 2025-12-30
fix: missing leading slash in API path

addy1947:
## Summary
I have fixed the issue of the missing leading slash in the api path

const fullUrl = `${baseUrl}${pathUrl.startsWith('/') ? pathUrl : '/' + pathUrl}`

I have edited the code such that if the pathUrl has "/", it will no nothing; if not, it will add a "/" before the pathUrl

## Related Issues
#1112 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] Breaking change
- [ ] Refactor

## Checklist
- [x] I have read the [CONTRIBUTING.md](https://github.com/MotiaDev/motia/blob/main/CONTRIBUTING.md)
- [x] My code follows the code style of this project
- [x] I have added tests where applicable
- [x] I have tested my changes locally
- [x] I have linked relevant issues


anthonyiscoding:
Hey @addy1947, thanks for tracking this down and fixing it! The missing leading slash in the API path is exactly the kind of subtle bug that can cause real headaches for developers, so good catch.

We wanted to give you a heads up that we're currently in the middle of a large architecture change and have actually already fixed this at least internally.

I'd love for you to revisit these kinds of contribution once the new architecture is in place — your awareness of edge cases like this will be really helpful in ensuring the new routing layer is solid.

Thank you for the fix and for being part of the community!

To be the first to know when the architecture change is released please add your email to this form: **https://forms.gle/24iCHL9yAk1i6LDc6**

PR #1229, opened 2026-03-04
feat: send pid for the engine

guibeira:
Previous prs:
https://github.com/iii-hq/sdk/pull/40
https://github.com/iii-hq/iii/pull/180

## Summary by CodeRabbit

* **New Features**
  * Workers can now include an optional process ID (PID) when registering, exposing PID in listings for better identification and tracking across SDKs.

* **Logging / Observability**
  * Registration and unregistration now emit runtime logs that include worker ID, IP address, and PID when available.

* **Tests**
  * Added tests validating PID presence, absence, storage, and listing behaviors.


ytallo (changes requested):
The monorepo uv path source is fine for local development, but the published motia package dependency should not be relaxed to an unbounded iii-sdk requirement. Please keep a compatible version floor or range so downstream consumers do not resolve older or future incompatible SDK releases.

PR #1247, opened 2026-03-06
Additional cursor rules

anthonyiscoding:
## Release Notes

* **New Features**
  * Added comprehensive how-to guides for configuration management, state handling, real-time streaming, queues, dead-letter queues, ephemeral workers, and trigger conditions.
  * Introduced interactive configuration reference tool for viewing engine settings.

* **Documentation**
  * Restructured documentation organization for clearer navigation.
  * Updated all code examples across Node.js, Python, and Rust with consistent naming patterns.
  * Expanded Discovery documentation with architecture diagrams and use cases.

* **Improvements**
  * Bridge URL now configurable via environment variables for flexible deployments.

### DIFF
--- a/.cursor/rules/docs.mdc
+++ b/.cursor/rules/docs.mdc
@@ -32,3 +32,18 @@ Before proceeding with any documentation request, take a deep breath and center 
 
 4. **Reference**:
    - Always read `.cursor/skills/doc_workflow.md` for rules regarding Tone, Component Usage (Callouts), and Chunked Execution.
+
+5. **Code**:
+   - Use 2 spaces for indents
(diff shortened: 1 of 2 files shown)


rohitg00:
Can we use these cursor rules for docs outside the monorepo? It's not really needed in monorepo and will make the repo heavy later on. That's why I kept skills separate. Let me know what you think.

You can also have symlink.

anthonyiscoding:
> Can we use these cursor rules for docs outside the monorepo? It's not really needed in monorepo and will make the repo heavy later on. That's why I kept skills separate. Let me know what you think.
> 
> You can also have symlink.

Where should we store them instead? Also some doc-adjacent things will need to be covered too (ex. READMEs, jsdoc) Happy to store them outside as long as they're still automatically discovered by agents

cc @sergiofilhowz

PR #1418, opened 2026-04-07
Preserve queue names during storage recovery

abishekgiri:
## Summary

- preserve queue names when rebuilding persisted file-based queues
- parse stored job keys from the right so queue names containing `:` or `:jobs:` are not truncated
- add a regression test covering a queue name like `billing:jobs:v2`

## Testing

- `cargo test -p iii test_rebuild_from_storage_preserves_queue_names_with_colons -- --nocapture`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of queue names containing special characters (colons) during state recovery from persistent storage. Queue names are now correctly preserved and jobs can be retrieved after system restart.
  
* **Tests**
  * Added test coverage for queue names with special characters in persistence scenarios.


abishekgiri:
Updated the file-based queue recovery path so persisted job keys keep the full queue name, even when the queue name contains  or . This also adds a regression test that rebuilds a delayed job from  after persistence.

abishekgiri:
Updated the file-based queue recovery path so persisted job keys keep the full queue name, even when the queue name contains colons or the literal segment :jobs:. This also adds a regression test that rebuilds a delayed job from billing:jobs:v2 after persistence.

PR #1419, opened 2026-04-07
Preserve stream names in adapter storage keys

abishekgiri:
## Summary

- add shared stream storage key helpers so stream names containing colons round-trip correctly
- keep group ids with colons intact when listing groups or enumerating stream metadata
- update both the built-in KV adapter and the Redis adapter to use the shared key handling
- add regression tests for colon-containing stream names, colon-containing group ids, and names that start with the reserved encoding prefix

## Testing

- `cargo test -p iii stream_storage_key_ -- --nocapture`
- `cargo test -p iii list_all_stream_preserves_stream_names_with_colons -- --nocapture`

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of stream names that include colons so streams and their groups are reliably listed and retrieved.
  * Improved preservation of stream and group identifiers during storage enumeration to prevent misclassification.

* **Tests**
  * Added tests to verify correct handling of stream names and group IDs with special characters.


abishekgiri:
Updated the stream storage key handling shared by the built-in KV adapter and the Redis adapter so stream names with colons are encoded safely, group ids with colons still round-trip, and list-all metadata no longer truncates names. This also adds regression coverage for colon-heavy names and the reserved encoding prefix.

sergiofilhowz (changes requested):
There's a formatting issue that blocked the pipeline, mind reviewing it? should be really simple

```
cargo fmt --all
```

https://github.com/iii-hq/iii/actions/runs/24058380271/job/70316261651?pr=1419

abishekgiri:
> There's a formatting issue that blocked the pipeline, mind reviewing it? should be really simple
> 
> ```
> cargo fmt --all
> ```
> 
> https://github.com/iii-hq/iii/actions/runs/24058380271/job/70316261651?pr=1419

Applied the formatting fix and pushed it to this branch. cargo fmt --all -- --check passes locally now, and the Build Engine check is re-running on the updated commit.

abishekgiri:
Addressed the remaining stream adapter review items in the latest push.

- Added warning logs in both KV and Redis `list_all_stream` paths when a `stream:` key cannot be parsed, so legacy or malformed keys are surfaced instead of being silently skipped.
- Added a Redis-specific regression test for colon-containing stream names to cover the SCAN-based code path. It is marked ignored because it requires a running Redis instance.

Validated with:
- `cargo fmt --all -- --check`
- `cargo test -p iii stream_storage_key_ -- --nocapture`
- `cargo test -p iii list_all_stream_preserves_stream_names_with_colons -- --nocapture`

anthonyiscoding:
@sergiofilhowz Is this one good to review?

PR #1420, opened 2026-04-07
Preserve RabbitMQ topic names in queue stats

abishekgiri:
## Summary

- preserve the full topic name when RabbitMQ subscription keys are parsed for topic stats and topic listings
- switch the parsing logic to split from the right so topics containing colons are counted under the correct name
- add a regression test for a topic like `orders:v1`

## Testing

- `cargo test -p iii subscription_topic_preserves_topics_with_colons -- --nocapture`

## Summary by CodeRabbit

* **Refactor**
  * Improved internal topic handling logic in message queue subscription processing for greater accuracy when extracting topic information.

* **Tests**
  * Added unit tests to verify topic extraction functionality handles complex topic names correctly.


abishekgiri:
Updated RabbitMQ topic reporting so subscription keys are parsed from the right instead of the left. That keeps topic stats and list-topics output correct when the topic name itself contains colons, and it adds a regression test for that shape.

anthonyiscoding:
Thanks @abishekgiri The right-to-left split approach makes sense for topic names with colons. One question though if subscriber IDs ever contain colons will this splitting always work correctly? (ex. `orders:v1:fn::validate` --> `orders:v1:fn:`)

@sergiofilhowz Do you have thoughts here? I'm not very familiar with RabbitMQ and our colons are more of a suggested design choice for paths than an enforced standard.

PR #1446, opened 2026-04-10
Keep regular topics separate from function queues in queue console

abishekgiri:
## Summary
- keep regular queue topics and configured function queues separate when they share the same visible name
- normalize namespaced function queue rows so the queue console displays them as function_queue entries
- add regression coverage for same-name regular topics and normalized function queue output

## Root Cause
- console_list_topics merged configured function queues into any existing topic whose visible name matched the queue name
- that allowed a regular topic like default to be treated as the function queue row, which overwrote its subscriber count and hid the separate function queue entry
- rows reported by the builtin adapter as __fn_queue::name were also normalized to name without updating broker_type

## Impact
- queue console listings now preserve both rows when a regular topic and a function queue share the same name
- normalized function queue rows are labeled consistently as function_queue

## Testing
- cargo fmt --all
- cargo test -p iii console_list_topics_ -- --nocapture

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed topic merging logic to correctly distinguish between regular and function-queue topics that share the same name, ensuring both remain visible with their appropriate broker type and subscriber count settings.


abishekgiri:
Updated the queue console topic merge logic so regular topics and configured function queues stay as separate rows even when they share the same visible name.

This also normalizes namespaced function queue rows to broker_type function_queue so the console output stays consistent.

Validated with:
- cargo fmt --all
- cargo test -p iii console_list_topics_ -- --nocapture

PR #1463, opened 2026-04-11
FIX: Stream Auth through protocol bug fixed

Atlas1225:
Per RFC 6455, if a WebSocket client sends a Sec-WebSocket-Protocol header during the handshake, the server must include a matching Sec-WebSocket-Protocol in its response — or the browser will immediately close the connection with code 1006.

The iii-stream worker was not echoing back the requested subprotocol, which made it impossible for browser clients to use the Sec-WebSocket-Protocol header as an authentication transport. The connection was silently dropped on every attempt.

FIX:
Based on your documentation the motia client in front should pass the token in websocket with this format
HEADER: Sec-WebSocket-Protocol: Authorization, \<jwt>.

The server must echo: HEADER Sec-WebSocket-Protocol: Authorization. Without that, the ws is terminated automatically.

The server extract the token from the header value in the authenticateStream function for verification.

- [x] I am licensing the entirety of this PR under Apache 2 and have all necessary rights to the code I am contributing.

## Summary by CodeRabbit

* **Bug Fixes**
  * WebSocket handshake now correctly detects and honors the Authorization subprotocol when clients request it, improving protocol negotiation during upgrades.

* **Tests**
  * Added tests validating WebSocket upgrade behavior and whether the protocol response header is included or omitted across various client header scenarios.

### DIFF
--- a/engine/src/workers/stream/stream.rs
+++ b/engine/src/workers/stream/stream.rs
@@ -74,6 +74,18 @@ async fn ws_handler(
 ) -> impl IntoResponse {
     let module = module.clone();
 
+    let has_authorization_protocol = headers
+        .get("sec-websocket-protocol")
+        .and_then(|v| v.to_str().ok())
+        .map(|s| s.split(',').any(|p| p.trim() == "Authorization"))
+        .unwrap_or(false);
+
+    let ws = if has_authorization_protocol {
+        ws.protocols(["Authorization"])
+    } else {
+        ws
+    };
+
     if let Some(auth_function) = module.auth_function.clone() {
         let engine = module.engine.clone();
         let input = StreamAuthInput {
--- a/engine/src/workers/stream/stream.rs
+++ b/engine/src/workers/stream/stream.rs
@@ -75,10 +75,11 @@ async fn ws_handler(
     let module = module.clone();
 
     let has_authorization_protocol = headers
-        .get("sec-websocket-protocol")
-        .and_then(|v| v.to_str().ok())
-        .map(|s| s.split(',').any(|p| p.trim() == "Authorization"))
-        .unwrap_or(false);
+        .get_all("sec-websocket-protocol")
+        .iter()
+        .filter_map(|v| v.to_str().ok())
+        .flat_map(|s| s.split(','))
+        .any(|p| p.trim() == "Authorization");
 
     let ws = if has_authorization_protocol {
         ws.protocols(["Authorization"])
--- a/engine/src/workers/stream/stream.rs
+++ b/engine/src/workers/stream/stream.rs
@@ -2006,4 +2006,135 @@ mod tests {
         assert!(message.contains(&format!("127.0.0.1:{port}")));
         assert!(message.contains("already in use"));
     }
+
+    // ── Sec-WebSocket-Protocol echo tests ───────────────────────────────────
+
+    fn build_ws_test_router() -> axum::Router {
+        use axum::extract::connect_info::MockConnectInfo;
+        crate::workers::observability::metrics::ensure_default_meter();
+        let engine = Arc::new(Engine::new());
+        let adapter: Arc<dyn StreamAdapter> = Arc::new(
+            crate::workers::stream::adapters::kv_store::BuiltinKvStoreAdapter::new(None),
+        );
+        let config = StreamModuleConfig {
+            port: 0,
+            host: "127.0.0.1".to_string(),
+            auth_function: None,
+            adapter: Some(crate::workers::traits::AdapterEntry {
+                name: "kv".to_string(),
+                config: None,
+            }),
+        };
+        let worker = Arc::new(StreamWorker::build(engine.clone(), config, adapter.clone()));
+        let mgr = Arc::new(StreamSocketManager::new(
+            engine,
+            adapter,
+            worker.clone(),
+            None,
+            worker.triggers.clone(),
+        ));
+        Router::new()
+            .route("/", get(ws_handler))
+            .with_state(mgr)
+            .layer(MockConnectInfo(std::net::SocketAddr::from(([127, 0, 0, 1], 0))))
+    }
+
+    fn ws_upgrade_request(protocol_headers: &[&str]) -> axum::http::Request<axum::body::Body> {
+        let mut builder = axum::http::Request::builder()
+            .uri("/")
+            .header("host", "localhost")
+            .header("connection", "Upgrade")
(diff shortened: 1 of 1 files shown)


Atlas1225:
@ytallo So I implemented in my frontend app a Stream authentication with oauth following the documentation: [https://www.motia.dev/docs/development-guide/react-stream-client#motiastreamprovider](https://www.motia.dev/docs/development-guide/react-stream-client#motiastreamprovider) but it seems that it is not working properly. The problem is that motia is not answering with the same header so the websocket is terminated automatically on browser side. So my fix is about this issue.

Atlas1225:
@ytallo Also I want to ask if I use previous versions (before motia 1.0.0) it should be more stable. I am doing some projects that needs to pass from SBX to production in 3 months aprox and I want to know if I can rely on this new framework.

Atlas1225:
@rohitg00 Hi, maybe you can help me clear my doubts as well :D

mfpiccolo:
Thanks for submitting! We have been heads down but @ytallo will take a look soon. 

Please reply with:
 
 Thank you!

ytallo:
Hey @Atlas1225 add at least one integration test?

Atlas1225:
I agree that my contributions in this PR to the engine code are licensed under Apache 2.0.

> Thanks for submitting! We have been heads down but @ytallo will take a look soon.
> 
> This PR touches engine code. For us to accept it, we need your explicit confirmation in this thread that you license your changes to the engine portion of the codebase under Apache 2.0.
> 
> Please reply with:  “I agree that my contributions in this PR to the engine code are licensed under Apache 2.0.”    Thank you!

Atlas1225:
> Hey @Atlas1225 add at least one integration test?

Okay, Sorry for that

anthonyiscoding:
@Atlas1225 Please edit your PR message (the one you wrote to open this PR) to add this string at the bottom:

`- [x] I am licensing the entirety of this PR under Apache 2 and have all necessary rights to the code I am contributing.`

deep-name:
nit on `engine/src/workers/stream/stream.rs` and the corresponding test:

the multi-header detection at the top is correct — `headers.get_all("sec-websocket-protocol")` walks every instance and finds `Authorization` regardless of which header it lives in. but `ws.protocols(["Authorization"])` then selects from the offer list axum parses, and per the comment on `test_ws_authorization_protocol_in_second_header_connection_succeeds`, axum only reads the first `Sec-WebSocket-Protocol` header for that selection. so when `Authorization` is in the second header (clients that emit headers separately rather than comma-joined — some HTTP/2 stacks, certain proxies), the upgrade succeeds with status 101 but with no `Sec-WebSocket-Protocol` in the response.

per [RFC 6455 §4.2.2](https://datatracker.ietf.org/doc/html/rfc6455#section-4.2.2) and the [HTML living spec](https://html.spec.whatwg.org/multipage/web-sockets.html#feedback-from-the-protocol) WebSocket-handshake step 4, if the client requested a subprotocol and the server's response doesn't echo one of the requested values, the browser MUST `Fail the WebSocket Connection`. so the 101 status the test asserts isn't actually the contract that prevents browser termination — the echoed header is.

the current test bakes this gap in (assert-only-status; no protocol assertion), so split-header browser clients would still hit the original "ws terminated automatically" symptom this PR set out to fix.

two ways to close it:

1. **fold the multi-header check into the response directly** — once `has_authorization_protocol` is true, add the response header explicitly (via `response.headers_mut().insert("sec-websocket-protocol", HeaderValue::from_static("Authorization"))` after `ws.on_upgrade(...)`) instead of routing through `ws.protocols([...])`. this bypasses axum's first-header-only selector for this specific subprotocol.

2. **rewrite the request headers before axum sees them** — fold all comma-separated values into a single `sec-websocket-protocol` header so axum's selector sees the flat list. less invasive but feels slightly more like working around axum than expressing intent.

happy to send a follow-up PR with (1) and a strengthened 5th test (`assert_eq!(proto.as_deref(), Some("Authorization"))`) if maintainers want, with attribution to @Atlas1225 — figured i'd flag here first since this is your call. either way, the conformance fix in this PR is real and worth merging; the split-header case is a small follow-on.

Atlas1225:
> nit on `engine/src/workers/stream/stream.rs` and the corresponding test:
> 
> the multi-header detection at the top is correct — `headers.get_all("sec-websocket-protocol")` walks every instance and finds `Authorization` regardless of which header it lives in. but `ws.protocols(["Authorization"])` then selects from the offer list axum parses, and per the comment on `test_ws_authorization_protocol_in_second_header_connection_succeeds`, axum only reads the first `Sec-WebSocket-Protocol` header for that selection. so when `Authorization` is in the second header (clients that emit headers separately rather than comma-joined — some HTTP/2 stacks, certain proxies), the upgrade succeeds with status 101 but with no `Sec-WebSocket-Protocol` in the response.
> 
> per [RFC 6455 §4.2.2](https://datatracker.ietf.org/doc/html/rfc6455#section-4.2.2) and the [HTML living spec](https://html.spec.whatwg.org/multipage/web-sockets.html#feedback-from-the-protocol) WebSocket-handshake step 4, if the client requested a subprotocol and the server's response doesn't echo one of the requested values, the browser MUST `Fail the WebSocket Connection`. so the 101 status the test asserts isn't actually the contract that prevents browser termination — the echoed header is.
> 
> the current test bakes this gap in (assert-only-status; no protocol assertion), so split-header browser clients would still hit the original "ws terminated automatically" symptom this PR set out to fix.
> 
> two ways to close it:
> 
> 1. **fold the multi-header check into the response directly** — once `has_authorization_protocol` is true, add the response header explicitly (via `response.headers_mut().insert("sec-websocket-protocol", HeaderValue::from_static("Authorization"))` after `ws.on_upgrade(...)`) instead of routing through `ws.protocols([...])`. this bypasses axum's first-header-only selector for this specific subprotocol.
> 2. **rewrite the request headers before axum sees them** — fold all comma-separated values into a single `sec-websocket-protocol` header so axum's selector sees the flat list. less invasive but feels slightly more like working around axum than expressing intent.
> 
> happy to send a follow-up PR with (1) and a strengthened 5th test (`assert_eq!(proto.as_deref(), Some("Authorization"))`) if maintainers want, with attribution to @Atlas1225 — figured i'd flag here first since this is your call. either way, the conformance fix in this PR is real and worth merging; the split-header case is a small follow-on.

Yes, please make a follow-on. I am not be able to continue at this moment the PR due to personal issues. If you can complete it, it would be very helpful.

PR #1466, opened 2026-04-13
fix(iii-worker): resolve builtin workers before registry on start

andersonleal:
## Summary

`iii worker add iii-http` (and `iii worker start iii-http`) was failing with:

```
✓ Worker iii-http added to config.yaml
  Worker 'iii-http' not found locally, checking registry...
  WARN iii_worker::cli::managed: Failed to fetch worker info: Worker 'iii-http' not found
error: Worker 'iii-http' not found locally or in registry. Run \`iii worker add iii-http\`.
  ⚠ Could not auto-start worker. Run \`iii worker start iii-http\` manually.
```

### Root cause

`handle_managed_start` in `crates/iii-worker/src/cli/managed.rs` only matched `OCI` / `Local` / `Binary` resolved types. Builtin workers (`iii-http`, `iii-stream`, `iii-state`, `iii-queue`, `iii-pubsub`, `iii-cron`, `iii-observability`) have no `image:`, `worker_path:`, or `~/.iii/workers/{name}` binary — they're served in-process by the iii engine itself via the factory registry in `engine/src/workers/config.rs`. So `resolve_worker_type` returned `Config`, the code fell through to `fetch_worker_info`, and the remote registry has no entry for builtins.

The sibling `handle_managed_add` already guarded on `get_builtin_default` before the same fallback (managed.rs:247). `handle_managed_start` just never got the matching check.

### Fix

- `handle_managed_start`: check `get_builtin_default` before `resolve_worker_type`. For builtins, short-circuit with an info message — no registry call, no spawn attempt.
- `handle_managed_add` (builtin auto-start branch): replace the misleading "Worker auto-started" message with "restart the engine to activate it", since builtins are engine-hosted and config hot-reload isn't wired up.

## Test plan

- [x] `cargo build -p iii-worker` — clean
- [x] `cargo test -p iii-worker --lib` — 221 passed
- [x] `cargo test -p iii-worker --test config_managed_integration` — 14 passed (includes new regression test `handle_managed_start_builtin_short_circuits`)
- [ ] Manual: `iii worker add iii-http` against a running engine no longer emits the registry error.

## Follow-up (not in this PR)

The engine doesn't hot-reload `config.yaml`. A builtin added to a running engine still needs a manual engine restart to take effect. The new message tells the user so, but a real reload path would be cleaner.

## Summary by CodeRabbit

* **Bug Fixes**
  * Builtin workers now report they are served in-process, short-circuiting external install/start paths; they require the engine to be running (users are prompted to start the engine when it isn’t) and will error if the builtin is not configured.
* **Tests**
  * Added integration tests verifying successful start for configured builtin workers and failure when a builtin is unconfigured.

### DIFF
--- a/crates/iii-worker/src/cli/managed.rs
+++ b/crates/iii-worker/src/cli/managed.rs
@@ -853,6 +853,20 @@ pub async fn handle_managed_start(worker_name: &str, _address: &str, port: u16) 
         eprintln!("{} {}", "error:".red(), e);
         return 1;
     }
+    // Builtin workers are served in-process by the iii engine (see
+    // engine/src/workers/config.rs factory registry). They have no external
+    // process to spawn and must not be resolved via the remote registry.
+    if get_builtin_default(worker_name).is_some() {
+        eprintln!(
+            "  {} '{}' is a builtin worker — served by the iii engine process.",
+            "info:".cyan(),
+            worker_name,
+        );
+        if !is_engine_running() {
+            eprintln!("  Start the engine to run it.");
+        }
+        return 0;
+    }
     match super::config_file::resolve_worker_type(worker_name) {
         ResolvedWorkerType::Oci { image, env } => {
             let worker_def = WorkerDef::Managed {
--- a/crates/iii-worker/src/cli/managed.rs
+++ b/crates/iii-worker/src/cli/managed.rs
@@ -255,8 +255,12 @@ pub async fn handle_managed_add(
                 );
             }
 
-            // The engine's file watcher will detect the config change and
-            // reload automatically — no need to start the worker here.
+            // If the engine is already running, its file watcher will detect
+            // the config change and reload automatically. If it isn't, nudge
+            // the user to start it (or customize first).
+            if !is_engine_running() {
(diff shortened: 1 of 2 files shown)


crates/iii-worker/src/cli/managed.rs
```diff
@@ -273,23 +273,15 @@ pub async fn handle_managed_add(
...
-                            "⚠".yellow(),
-                            name
-                        );
-                    }
-                }
+                eprintln!(
+                    "  {} '{}' is a builtin — restart the engine to activate it.",
+                    "info:".cyan(),
+                    name,
+                );
```
sergiofilhowz: Guilherme is working on auto restart at the moment, let's just add
```suggestion
                eprintln!("{} worker added", name.bold().cyan());
```

sergiofilhowz: Image

would  be nice to be in this format

sergiofilhowz: let's delete this line

sergiofilhowz (approved):

PR #1521, opened 2026-04-21
fix(rbac): auto-allow SDK-infrastructure engine::* functions

ytallo:
## Summary

- **Engine:** Expand the always-allowed set in `is_function_allowed` from a single constant to an `INFRASTRUCTURE_FUNCTIONS` slice covering the `engine::*` IDs the SDK invokes transparently (channels, `workers::register`, `log::*`, `baggage::*`). Slice is documented as part of iii's public contract with an additive-only stability policy.
- **Engine:** FORBIDDEN errors now name the offending `function_id` in the message, turning a blank rejection into a copy-into-`expose_functions` diagnostic. `tracing::warn!` fires when `forbidden_functions` denies an infrastructure ID so silent worker breakage becomes debuggable.
- **Node SDK:** New `IIIInvocationError extends Error` with `code`, `function_id`, `stacktrace`. Wraps wire `ErrorBody` in `onInvocationResult` and unifies the invocation timeout path under the same type. Exported from `iii-sdk`. This fixes the `[object Object]` half of the bug report.
- **Docs:** `docs/how-to/worker-rbac.mdx` gains a 2-minute quick-start snippet at the top, an explicit "Infrastructure functions" subsection with the ID table + additive-only policy, and discovery-gating guidance for callers who want `iii.listFunctions()` to work.

## Motivation

A user reported `FORBIDDEN: "function not allowed"` when invoking a handler via HTTP with RBAC enabled — the handler indirectly tripped an SDK-transparent `engine::*` call their `expose_functions` didn't cover. The rejection printed as `[object Object]` unless wrapped in `try/catch`. Every comparable system (Supabase RLS, Firebase Rules, Temporal, NATS) exempts its own plumbing from user-authored policies; iii was the outlier.

## Out of scope (follow-up)

Python and Rust SDK typed-error parity (mirror `IIIInvocationError`) — the engine-side fix is the single source of truth for the carve-out, so every SDK already benefits without a protocol change. Typed-error parity on those SDKs is a DX polish that can ship independently.

## Test plan

- [x] `cargo test -p iii --lib rbac_config` — 24 passed (new: `infrastructure_functions_always_allowed`, `..._respect_forbidden_list`, `..._respect_allowed_list`, `discovery_functions_still_gated`, `discovery_functions_allowed_via_expose`, `no_rbac_config_still_allows_everything`).
- [x] `cargo test -p iii --test rbac_infrastructure_e2e` — 4 passed (infrastructure allowed under restricted expose, discovery denied, forbidden-list wins, middleware bypass preserved).
- [x] `cargo test -p iii --lib` — 1466 passed, 1 ignored (no regressions).
- [x] `bunx vitest run tests/errors.test.ts` — 6 passed (`IIIInvocationError` shape, `String(err) !== '[object Object]'`, real Error subclass).
- [ ] Manual end-to-end reproduction of the reporter's config: run engine with `expose_functions: [match("api::*")]`, call `logger.info` from a handler, confirm 200 + log entry, then call `iii.listFunctions()` and confirm the error prints as `IIIInvocationError: FORBIDDEN: function 'engine::functions::list' not allowed`.
- [ ] Reach out to the reporter with a build from this PR before closing the original issue.

## Summary by CodeRabbit

* **New Features**
  * Infrastructure engine functions are now automatically allowed in RBAC, even with restrictive exposure policies.
  * Added typed error class in SDK for invocation failures, including function ID and error code information.

* **Documentation**
  * Added RBAC quickstart guide with end-to-end examples and configuration samples.
  * Clarified infrastructure vs. discovery function behavior in RBAC policies.

* **Bug Fixes**
  * Forbidden function errors now include the specific function ID in the error message for better debugging.

### DIFF
--- a/docs/how-to/worker-rbac.mdx
+++ b/docs/how-to/worker-rbac.mdx
@@ -5,6 +5,47 @@ description: 'Configure the iii-worker-manager to enable role-based access contr
 
 This guide walks you through enabling RBAC on your iii instance using the **iii-worker-manager**. Workers connect via WebSocket to a dedicated RBAC port, separate from the internal engine bridge.
 
+## Quick start: RBAC in 2 minutes
+
+Drop these two files side-by-side, start the engine with the config, run the worker, and curl the endpoint. If you see `hello from the handler` in the worker's terminal, RBAC is working.
+
+```yaml title="iii-config.yaml"
+workers:
+  - name: iii-http
+    config: { port: 3111, host: 127.0.0.1 }
+  - name: iii-worker-manager
+    config:
+      port: 49135
+      rbac:
+        expose_functions:
+          - match("api::*")
+```
+
+```typescript title="worker.ts"
+import { Logger, registerWorker } from 'iii-sdk'
+
+const iii = registerWorker('ws://localhost:49135')
+const logger = new Logger()
+
+iii.registerFunction('api::hello', async () => {
+  logger.info('hello from the handler')
+  return { ok: true }
+})
+```
+
+```bash title="run it"
+# In one terminal: start the engine with the config above
+iii-worker start --config iii-config.yaml
+
+# In another: run the worker
+bun worker.ts
+
+# In a third: hit the function over HTTP
+curl http://127.0.0.1:3111/api/hello
+```
+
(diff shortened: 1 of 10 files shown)


ytallo:
Superseded by #1525, which consolidates the full RBAC UX fix (engine carve-out + Node SDK + Python SDK + docs + remediation phrase) into a single PR. The two commits from this PR are already present on #1525's branch; no work is lost.

PR #1530, opened 2026-04-22
test(rbac): isolate carve-out handler tests from shared registrations

ytallo:
## Summary

The "real usage" carve-out tests added in #1525 registered a handler with the same function_id (`test::ew::valid-token-echo`) that the shared test harness already registers in `ensure_functions_registered()`.

A worker-client registration supersedes the shared one in the engine's routing table, and when the worker shuts down at end-of-test the function unregisters entirely. Every subsequent serial test that expected `test::ew::valid-token-echo` to exist then fails with `function_not_found: Function test::ew::valid-token-echo not found`.

Observed on PR #1529's SDK Rust Tests job (run 24785381936):
- `should_return_auth_result_for_valid_token` FAILED
- `should_only_list_allowed_functions_for_valid_token` FAILED

Fix applied to all three SDKs: register under a test-unique function_id `test::ew::carveout-logger-handler` instead. Doesn't touch shared fixtures, drops cleanly at end of test.

## Test plan

- [x] `cargo check -p iii-sdk --tests --test rbac_workers` — compiles clean
- [x] Node SDK builds clean
- [x] Python syntax check passes
- [ ] Full Rust RBAC test suite passes end-to-end (requires engine harness, will run in CI)

## Summary by CodeRabbit

* **Tests**
  * Improved test isolation across Node.js, Python, and Rust SDK implementations to prevent interference between test cases during execution.

### DIFF
--- a/sdk/packages/node/iii/tests/rbac-workers.test.ts
+++ b/sdk/packages/node/iii/tests/rbac-workers.test.ts
@@ -406,8 +406,12 @@ describe('RBAC Workers', () => {
     })
 
     try {
+      // Use a test-unique function_id so this test doesn't clobber shared
+      // registrations — a worker-client registration supersedes the shared
+      // one, and the implicit unregister when the worker shuts down would
+      // break every subsequent test that expects the shared function to exist.
       iiiClient.registerFunction(
-        'test::ew::valid-token-echo',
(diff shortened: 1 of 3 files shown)


ytallo:
Rolling the RBAC test isolation fix into #1529 instead.

PR #1547, opened 2026-04-25
fix(state,stream): support nested shallow-merge paths in atomic update ops

### LINKED ISSUE #1546: BUG: Nested object merge not supported in atomic `state::update` (and potentially `stream::update`) ops
Trying to merge nested objects using atomic `merge` operations results in a warning:

```
[WARN] iii::builtins::kv Only root-level merge is supported
```

This prevents building structured/nested state using atomic ops.

---

### Expected Behavior

`merge` operations should support nested paths (or at least allow merging objects at arbitrary depths), enabling use cases like:

* Incrementally building nested structures without overwriting siblings
* Avoiding full object rewrites when only a deeply nested field changes

---

### CurrentBehavior

* `merge` only works at the root level.
* Providing a `path` (e.g., `path: id`) triggers a warning and does not behave as expected.
* Nested merges are effectively unsupported, forcing workarounds like:

  * Flattening keys
  * Rewriting entire objects
  * Encoding nesting manually into keys

---

#### Case 1: Attempting nested merge with dynamic key

```ts
await iii.trigger({
  function_id: "state::update",
  payload: {
    scope: "audio::transcripts",
    key: "session-some-key",
    ops: [
      {
        type: "merge",
        path: id,
        value: {
          // intended nested structure: session -> id -> timestamp
          [data.Timestamp]: json.transcript,
        },
      },
    ],
  },
});
```

#### Case 2: Simple nested field update

```ts
await iii.trigger({
  function_id: "state::update",
  payload: {
    scope: "audio::transcripts",
    key: `session-some-key`,
    ops: [
      {
        type: "merge",
        path: id,
        value: {
          author: authorId,
        },
      },
    ],
  },
});
```

---

### Observed Output

Both cases emit:

```
[WARN] iii::builtins::kv Only root-level merge is supported
```

And do not correctly apply the nested merge.

---

### Impact

* Makes it difficult to model hierarchical state using atomic operations.
* Forces inefficient or brittle workarounds (full rewrites, key flattening).
* Limits usefulness of `merge` in real-world streaming or incremental update scenarios (e.g., transcripts, logs, message trees).

---

### Suggested Improvements

**Option 1: Support nested merge paths**

* Allow `path` to specify nested locations (e.g., `"foo.bar.baz"` or array paths).
* Perform deep merge at the specified path.

**Option 2: Introduce a ****`deepMerge`**** op**

* Explicitly differentiate between shallow root merge and recursive merge.

**Option 3: Add ****`set`**** with path support**

* Allow setting values at arbitrary nested paths without requiring full object reconstruction.

---

### Additional Context

Current limitation forces patterns like:

```ts
// flattening keys instead of nesting
{
  [`${id}:${timestamp}`]: transcript
}
```

or:

```ts
// read-modify-write full object (non-atomic)
```

Both approaches reduce clarity and increase the risk of race conditions.

ytallo:
Closes #1546.

## Summary

Atomic `merge` ops in `state::update` / `stream::update` previously dropped every non-empty `path` with a silent `[WARN] Only root-level merge is supported`. SDK types accepted any string while the engine threw it away — a trust violation that forced flat-key workarounds and full read-modify-write rewrites.

This PR generalizes `UpdateOp::Merge.path` to accept either a single string (legacy / first-level field, backward-compatible) **or** an array of literal segments for nested merge. The engine walks the segments, replacing or auto-creating non-object intermediates RFC 7396-style, then shallow-merges `value` at the target. Both the in-memory builtin and the Redis Lua `JSON_UPDATE_SCRIPT` are updated in lockstep so adapters stay consistent.

## What changed

**Engine (Rust + Lua)**
- `engine/src/update_ops.rs` — nested merge implementation, validation bounds (path depth ≤ 32, segment ≤ 256 B, value depth ≤ 16, ≤ 1024 top-level keys), prototype-pollution rejection (`__proto__` / `constructor` / `prototype`), structured error returns. 15 new unit tests.
- `engine/src/workers/redis.rs` — Lua `JSON_UPDATE_SCRIPT` rewritten with mirrored semantics. The previous `get_path` silently truncated array paths to `path[1]`; now it walks every segment with the identical bounds and proto rejection. Errors return as a 4th element of the response tuple.
- Both Redis adapter parsers updated to handle 3- or 4-element responses (backward-compatible).
- `engine/src/builtins/kv.rs` — plumbs errors through `UpdateResult`.
- `engine/src/workers/state/structs.rs` — JsonSchema regression test catches silent schemars drift.
- `engine/tests/state_stream_update_e2e.rs` (**new file**) — 6 integration scenarios.

**SDKs**
- **Rust** (`sdk/packages/rust/iii/src/types.rs`): new `MergePath` untagged enum (`Single` before `Segments` — variant order is load-bearing for serde), `UpdateOpError`, `UpdateResult.errors` field (omitted from JSON when empty), `merge_at` accepts `impl Into<MergePath>`, new `merge_at_path` helper. 5 new round-trip tests.
- **Node** (`sdk/packages/node/iii/src/stream.ts` + `state.ts`): `MergePath = string | string[]`, `UpdateMerge.path?: MergePath`, `UpdateOpError`, `errors?: UpdateOpError[]` on `StreamUpdateResult` / `StateUpdateResult`. JSON round-trip tests.
- **Browser** (`sdk/packages/node/iii-browser/src/stream.ts` + `state.ts`): same shape; new export-level type tests in `tests/exports.test.ts`.
- **Python** (`sdk/packages/python/iii/src/iii/stream.py`): `path: str | list[str] | None`, `UpdateOpError` model, `errors` field on `StreamUpdateResult`. 5 new pydantic round-trip tests.
- **Console frontend** (`console/packages/console-frontend/src/api/types/shared.ts`): `merge.path?: string | string[]` and `StreamUpdateOpError`.

**Docs**
- `docs/workers/iii-state.mdx` — merge row rewritten with both forms, validation contract, `errors` field documented.
- `docs/workers/iii-stream.mdx` — same prose update + cross-link.
- `docs/how-to/manage-state.mdx` — new "Build per-session structured state with nested merge" recipe in Node / Python / Rust tabs, walking issue #1546's exact use case end to end.

### Use case 1 — issue #1546 case 1 (first-level field)
```ts
await iii.trigger({
  function_id: 'state::update',
  payload: {
    scope: 'audio::transcripts',
    key: 'session-abc',
    ops: [
      { type: 'merge', path: sessionId, value: { [Date.now()]: chunk } },
    ],
  },
})
```
Now writes `{ "<sessionId>": { "<timestamp>": chunk } }` and accumulates timestamps across calls. Previously: silent no-op.

### Use case 2 — nested merge (closes the bug class)
```ts
{
  type: 'merge',
  path: ['sessions', 'abc', 'metadata'],
  value: { author: 'alice' },
}
```
Walks `sessions → abc → metadata`, auto-creating each intermediate object. Sibling keys at every level are preserved.

### Use case 3 — root merge (backward-compatible)
```ts
{ type: 'merge', value: { status: 'active' } }     // path omitted
{ type: 'merge', path: '', value: { status: 'active' } }  // existing senders unchanged
```

### Use case 4 — replace-on-non-object (defends against poisoned state)
```ts
// Existing state: { sessions: null }
{ type: 'merge', path: ['sessions', 'abc'], value: { author: 'alice' } }
// Result: { sessions: { abc: { author: 'alice' } } }
// (intermediate `null` is replaced, not skipped — no more "stray null
//  blocks future merges forever")
```

### Use case 5 — structured error on invalid input
```jsonc
// Request:  merge with path ["__proto__", "polluted"]
// Response:
{
  "old_value": { ... },
  "new_value": { ... },          // unchanged; the merge did NOT apply
  "errors": [{
    "op_index": 0,
    "code": "merge.path.proto_polluted",
    "message": "Path segment \"__proto__\" is a prototype-pollution sink",
    "doc_url": "https://docs.iii.dev/workers/iii-state#merge-bounds"
  }]
}
```

Path segments are **literal**: `["a.b"]` writes one key named `"a.b"`, not `a → b`. This matches iii's existing `FieldPath` literal-key contract on the other ops and sidesteps Firestore-style dotted-string ambiguity.

## Backward compatibility

- Existing `path: ""` and `path: <string>` callers keep working unchanged.
- Existing `path: <non-empty-string>` payloads that previously emitted the warning **start working** as first-level merge (the bug fix).
- The new `errors` field is omitted from the JSON response when empty, so old clients that don't read it see the same `{old_value, new_value}` shape.
- Redis adapter response went from 3 elements to "3 or 4"; old engine ↔ old adapter still works, new engine ↔ old adapter sees only 3 elements when no errors occur.

## Test plan

- [ ] `cargo test -p iii-sdk --lib` → 59 passing (5 new round-trip tests)
- [ ] `cargo test -p iii update_ops:: --lib` → 26 passing (15 new merge / validation tests)
- [ ] `cargo test -p iii --test state_stream_update_e2e` → 6 passing (new file)
- [ ] `cargo test -p iii --lib` → 1501 passing, 0 failing
- [ ] `cd sdk/packages/node/iii && pnpm tsc --noEmit -p tsconfig.json` → clean
- [ ] `cd sdk/packages/node/iii-browser && pnpm vitest run tests/exports.test.ts` → 9 passing (4 new merge tests)
- [ ] `cd console/packages/console-frontend && pnpm tsc --noEmit -p tsconfig.json` → clean
- [ ] `cd sdk/packages/python/iii && uv run pytest tests/test_stream_models.py` → 5 new tests for `UpdateMerge` / `UpdateOpError` / `StreamUpdateResult`
- [ ] Manual repro of issue #1546 against a local engine — both case 1 (timestamps accumulating under a session id) and case 2 (`{author}` merging without nuking siblings) now work as the user described.
- [ ] `[WARN] Only root-level merge is supported` no longer fires anywhere (`rg "Only root-level merge"` returns zero matches in `engine/`, `sdk/`, `docs/`).

## Summary by CodeRabbit

* **New Features**
  * Merge updates can target nested locations via an optional path (string or array of literal segments).
  * Update results may include per-operation structured errors for invalid merge ops.

* **Documentation**
  * Expanded guides clarifying path semantics and nested merge behavior, constraints, and error handling for state and stream updates.

* **Tests**
  * New and updated integration/unit tests validating nested merge semantics and per-op error reporting.

### DIFF
--- a/sdk/packages/node/iii-browser/tests/exports.test.ts
+++ b/sdk/packages/node/iii-browser/tests/exports.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'vitest'
 import { ChannelReader, ChannelWriter, registerWorker, TriggerAction } from '../src/index'
-import type { UpdateAppend, UpdateMerge, UpdateOp, UpdateOpError } from '../src/stream'
+import type { UpdateAppend, UpdateMerge, UpdateOp } from '../src/stream'
 
 describe('Package Exports', () => {
   it('should export main SDK symbols', () => {
@@ -56,21 +56,6 @@ describe('Package Exports', () => {
     expect(parsed).toEqual(op)
   })
 
-  it('should type merge with no path (root merge)', () => {
-    const op: UpdateMerge = { type: 'merge', value: { x: 1 } }
-    expect(op.path).toBeUndefined()
-  })
-
-  it('should type UpdateOpError with required fields', () => {
-    const err: UpdateOpError = {
-      op_index: 0,
-      code: 'merge.path.too_deep',
-      message: 'Path depth 33 exceeds maximum of 32',
-      doc_url: 'https://docs.iii.dev/workers/iii-state#merge-bounds',
-    }
-
-    expect(err.code).toBe('merge.path.too_deep')
-  })
 
   it('should import state module', async () => {
     const stateModule = await import('../src/state')
--- a/sdk/packages/node/iii-browser/tests/exports.test.ts
+++ b/sdk/packages/node/iii-browser/tests/exports.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'vitest'
 import { ChannelReader, ChannelWriter, registerWorker, TriggerAction } from '../src/index'
-import type { UpdateAppend, UpdateOp } from '../src/stream'
+import type { UpdateAppend, UpdateMerge, UpdateOp, UpdateOpError } from '../src/stream'
 
 describe('Package Exports', () => {
   it('should export main SDK symbols', () => {
@@ -30,6 +30,46 @@ describe('Package Exports', () => {
     const ops: UpdateOp[] = [op]
 
     expect(ops[0]).toEqual({ type: 'append', path: 'chunks', value: 'hello' })
+  })
+
+  it('should type merge with a string path (legacy/first-level form)', () => {
+    const op = {
+      type: 'merge',
+      path: 'session-abc',
+      value: { author: 'alice' },
+    } satisfies UpdateMerge
+
+    expect(op).toEqual({ type: 'merge', path: 'session-abc', value: { author: 'alice' } })
+  })
+
+  it('should type merge with an array path (nested form)', () => {
+    const op = {
+      type: 'merge',
+      path: ['sessions', 'abc'],
+      value: { ts: 'chunk' },
+    } satisfies UpdateMerge
+
+    expect(op).toEqual({ type: 'merge', path: ['sessions', 'abc'], value: { ts: 'chunk' } })
+
+    // Round-trips through JSON unchanged.
+    const parsed: UpdateMerge = JSON.parse(JSON.stringify(op))
+    expect(parsed).toEqual(op)
+  })
+
+  it('should type merge with no path (root merge)', () => {
+    const op: UpdateMerge = { type: 'merge', value: { x: 1 } }
+    expect(op.path).toBeUndefined()
+  })
+
+  it('should type UpdateOpError with required fields', () => {
+    const err: UpdateOpError = {
+      op_index: 0,
+      code: 'merge.path.too_deep',
+      message: 'Path depth 33 exceeds maximum of 32',
+      doc_url: 'https://docs.iii.dev/workers/iii-state#merge-bounds',
+    }
+
+    expect(err.code).toBe('merge.path.too_deep')
   })
 
   it('should import state module', async () => {
--- a/sdk/packages/node/iii-browser/tests/exports.test.ts
+++ b/sdk/packages/node/iii-browser/tests/exports.test.ts
@@ -1,6 +1,6 @@
 import { describe, expect, it } from 'vitest'
 import { ChannelReader, ChannelWriter, registerWorker, TriggerAction } from '../src/index'
-import type { UpdateAppend, UpdateMerge, UpdateOp } from '../src/stream'
+import type { UpdateAppend, UpdateOp } from '../src/stream'
 
 describe('Package Exports', () => {
   it('should export main SDK symbols', () => {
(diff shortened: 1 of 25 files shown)


sdk/packages/node/iii-browser/tests/exports.test.ts
```diff
@@ -32,6 +32,46 @@ describe('Package Exports', () => {
...
+  it('should type UpdateOpError with required fields', () => {
+    const err: UpdateOpError = {
+      op_index: 0,
+      code: 'merge.path.too_deep',
+      message: 'Path depth 33 exceeds maximum of 32',
+      doc_url: 'https://docs.iii.dev/workers/iii-state#merge-bounds',
+    }
+
+    expect(err.code).toBe('merge.path.too_deep')
+  })
```
sergiofilhowz: this isn't testing anything actually

sdk/packages/node/iii/tests/stream.test.ts
```diff
@@ -248,6 +256,45 @@ describe('Stream Operations', () => {
...
+      }
+      const result: StreamUpdateResult<{ x: number }> = {
+        old_value: undefined,
+        new_value: { x: 1 },
+        errors: [err],
+      }
+
+      expect(result.errors?.[0].code).toBe('merge.path.proto_polluted')
+    })
+
```
sergiofilhowz: this isn't testing anything actually, we need to run these update operations and verify if they updated the record correctly

sergiofilhowz (approved):

PR #1607, opened 2026-05-06
feat(cli): iii project (init / generate-docker) + retire iii create

guibeira:
## Summary

Adds a built-in `iii project` subcommand and retires the standalone `iii create` flow.

Per Anthony's review feedback, **all template content lives in the canonical templates repo** ([iii-hq/templates#2](https://github.com/iii-hq/templates/pull/2)) — the engine never embeds template files via `include_str!`. Both the bare scaffold (`config.yaml` / `.gitignore` / `data/`) and the Docker assets (`Dockerfile` / `docker-compose.yml`) are fetched at runtime through `scaffolder-core`. Template fixes ship via PRs to the templates repo, independent of engine releases.

- **`iii project init [<name>]`** — fetches the `bare` template via `scaffolder-core`. Persists `.iii/project.ini` (with `device_id`, `project_id`, `project_name`, `source`) using the canonical `scaffolder_core::telemetry::write_project_ini`. `--docker` layers the `docker` template on top.
- **`iii project init --template <name>`** — drives the interactive scaffolder TUI (was `iii create`). `--template-dir`, `--languages`, `--skip-iii`, `--yes` pass through.
- **`iii project generate-docker [--directory <path>]`** — fetches `Dockerfile` + `docker-compose.yml` from the `docker` template directly via `TemplateFetcher::fetch_file_bytes` (skipping the `shared_files` merge so we don't clobber the user's existing `config.yaml`). Engine writes the device_id-aware `.env` afterwards.
- **Removed** `iii create` top-level command, the `iii-tools` managed binary, and its release pipeline.

> [!IMPORTANT]
> **Breaking:** `iii create` is gone. Migrate to `iii project init --template <name>`.

> [!CAUTION]
> **Templates repo dependency:** end-user `iii project init myapp` (no `--template-dir`) will fail until [iii-hq/templates#2](https://github.com/iii-hq/templates/pull/2) lands and adds the `bare`/`docker` templates to the canonical repo. Engine CI passes via local fixtures (`engine/tests/fixtures/templates/`), so this PR is mergeable independently.

### Bare flow — `iii project init` + `generate-docker`

iii project init bare demo

### Template flow — `iii project init --template quickstart`

iii project init template demo

## Architectural changes (Anthony's review)

| Concern | Resolution |
|---|---|
| Templates hardcoded via `include_str!` | Moved to [iii-hq/templates#2](https://github.com/iii-hq/templates/pull/2). Engine fetches at runtime via `scaffolder-core`. |
| `IiiConfig` (impl `ProductConfig`) duplicated | Moved into `scaffolder-core` as `pub struct IiiConfig`. Engine imports `scaffolder_core::IiiConfig`. |
| `write_project_ini` re-implemented in engine | Engine deletes its `ProjectIni` writer; uses `scaffolder_core::telemetry::write_project_ini` (extended in this PR with optional `device_id`). |
| Amplitude transport / `API_KEY` duplicated | `cli/telemetry.rs` no longer maintains its own; uses `AmplitudeClient` from `workers/telemetry/amplitude.rs`. `API_KEY` exists in one place. |
| `sanitize_error` only sanitized CLI events | Moved into `amplitude.rs` and applied at the **send layer** (`AmplitudeClient::send_event`). Every Amplitude event from the engine binary now gets PII redaction. Includes Windows path support. |

## Usage examples

```bash
# Bare scaffold in a new directory (positional name)
iii project init myapp
cd myapp
iii                                    # start the engine

# Bare scaffold + Docker assets in one step
iii project init myapp --docker
cd myapp
docker compose up

# Add Docker assets to an existing project later
cd myapp
iii project generate-docker

# Scaffold from a template (interactive)
iii project init --template quickstart

# Non-interactive template (CI / automation)
iii project init --template quickstart --languages ts,py --yes --directory myapp

# Use a local templates directory (development / tests)
iii project init --template-dir ~/work/iii-templates --directory myapp
```

## What's in the box

| Area | Files |
|---|---|
| New subcommand (in-engine dispatch only) | `engine/src/cli/project/mod.rs` |
| Telemetry helpers | `engine/src/cli/telemetry.rs` (`send_project_init_*`) |
| Telemetry transport (canonical) | `engine/src/workers/telemetry/amplitude.rs` (`API_KEY`, `sanitize_error`, `AmplitudeClient`) |
| Template content | [iii-hq/templates#2](https://github.com/iii-hq/templates/pull/2) (`bare` + `docker` templates) |
| scaffolder-core changes | `crates/scaffolder-core/src/iii.rs` (`IiiConfig`); `telemetry.rs` (`write_project_ini` extended for `device_id`) |
| New deps on engine | `scaffolder-core`, `console`, `ctrlc` |
| Removed | `crates/iii-tools/`, `engine/src/cli/project/{scaffold.rs, docker.rs, project_ini.rs, templates/*}`, registry entry, dispatch arm, CI release job |
| Tests | `engine/tests/project_init_e2e.rs` (11 e2e), `engine/tests/fixtures/templates/` (mirror of iii-hq/templates#2), `crates/scaffolder-core` (40 lib tests) |

## Test plan

- [x] `cargo test -p iii --bin iii` — 101 passing
- [x] `cargo test -p iii --test project_init_e2e` — 11 passing (all use `--template-dir` against fixtures)
- [x] `cargo test -p iii --test cli_integration` — 23 passing
- [x] `cargo test -p scaffolder-core --lib` — 40 passing
- [x] `cargo build -p iii --release` clean
- [x] CodeRabbit findings (12 across three rounds): all addressed
- [x] CI green on Linux runner (most recent: ✓ License Header / ✓ CI / ✓ License Agreement)
- [ ] **Land iii-hq/templates#2 first**, then run `iii project init myapp` (no `--template-dir`) against the live remote to confirm end-to-end
- [ ] Manual smoke: `docker compose up` reaches engine on `:3111` (deferred — assets are now template-sourced rather than `include_str!`)

## Notes for reviewers

- **Templates-repo prerequisite:** the engine PR is mergeable today thanks to local fixtures. End-user impact is gated on [iii-hq/templates#2](https://github.com/iii-hq/templates/pull/2) — review and land that first if you want `iii project init myapp` to work post-merge.
- The `iiidev/iii:latest` Docker base is distroless with binary at `/app/iii`; the generated Dockerfile uses `ENTRYPOINT ["/app/iii"]`.
- Distroless concern: `iii project init --template <name>` runs an interactive TUI; inside Docker it'll fail without a TTY. The bare flow is TTY-free and works in containers.
- **Open behavioral question (raised by Anthony, see thread):** `iii project init` with no args uses the current directory and silently overwrites pre-existing `config.yaml`/`.gitignore` (because `scaffolder_core::copy_template` uses `fs::write`, not `write_if_absent`). No `--allow-non-empty` override exists. Worth deciding whether to add a non-empty guard + `--force` flag in this PR or a follow-up.
- CodeRabbit review fixes spread across multiple commits: `1e7972f0`, `12fbda18`, `5924a545`, `b6dd57d9`, `58b5002b`, `17b855e8`. All actionable findings (12 total across three rounds) addressed.


.github/workflows/create-tag.yml
```diff
@@ -77,7 +77,6 @@ jobs:
                      sdk/packages/python/iii/pyproject.toml \
                      console/packages/console-rust/Cargo.toml \
                      crates/scaffolder-core/Cargo.toml \
-                     crates/iii-tools/Cargo.toml \
```
anthonyiscoding, line 80: Do these changes impact or break the existing templates? Does `iii project init --template quickstart` work now the same as it did when it was `iii create --template quickstart`?

engine/src/cli/project/templates/gitignore.template
```diff
@@ -0,0 +1,31 @@
...
+Cargo.lock.tmp
+
+# Editors
+.vscode/
+.idea/
+*.swp
+.DS_Store
+
+# Env
+.env
```
anthonyiscoding:
```suggestion
.env
.env.*
!.env.example
```

engine/src/cli/project/templates/iii.lock.template
```diff
@@ -0,0 +1,2 @@
+version: 1
```
anthonyiscoding: This filenaming seems to differ from the original scaffolder that didn't require `.template` postfixes which made it easier to modify files and templates.

For example any template that has code files or even json or yaml is a pain to edit when it has .template at the end as you lose both syntax highlighting and type completion.

You can see the code for this functionality here: https://github.com/iii-hq/iii/tree/main/crates/scaffolder-core/src

And examples of the templates here:

root template (applies to all iii templates): https://github.com/iii-hq/cli-tooling/blob/main/templates/iii/template.yaml
iii quickstart template: https://github.com/iii-hq/cli-tooling/tree/main/templates/iii/quickstart

engine/src/cli/project/docker.rs
```diff
@@ -0,0 +1,128 @@
+// Copyright Motia LLC and/or licensed to Motia LLC under one or more
+// contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+// This software is patent protected. We welcome discussions - reach out at team@iii.dev
+// See LICENSE and PATENTS files for details.
+
+use std::path::Path;
+
+const DOCKERFILE_TEMPLATE: &str = include_str!("templates/Dockerfile.template");
```
anthonyiscoding: This too, there were existing template.yaml descriptor files that made it trivial to add/remove files in a template, and rename them, all without hard coding things in Rust.

The hardcoding also breaks the release pipeline for templates being independent of the binaries which previously allowed older versions of iii to receive the newest templates. There's also a facility for mandating minimum iii versions in the templates.

The approach in this PR means that if there is a fix or improvement to a template it requires a user to upgrade iii in every case before they can use the correct template.

engine/src/cli/project/docker.rs
```diff
@@ -0,0 +1,128 @@
...
+const COMPOSE_TEMPLATE: &str = include_str!("templates/docker-compose.yml.template");
+
+pub fn write_docker_assets(project_root: &Path, device_id: &str) -> std::io::Result<()> {
+    write_if_absent(&project_root.join("Dockerfile"), DOCKERFILE_TEMPLATE)?;
+    write_if_absent(&project_root.join("docker-compose.yml"), COMPOSE_TEMPLATE)?;
+    write_env_file(project_root, device_id)?;
+    Ok(())
+}
+
+fn write_env_file(project_root: &Path, device_id: &str) -> std::io::Result<()> {
```
anthonyiscoding: These kinds of functions could be made generic and then used to extend what can be described in `template.yaml`s

engine/src/cli/project/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+        "https://github.com/iii-hq/templates.git"
+    }
+    fn template_url_env(&self) -> &'static str {
+        "III_TEMPLATE_URL"
+    }
+    fn requires_iii(&self) -> bool {
+        true
+    }
+    fn docs_url(&self) -> &'static str {
+        "https://iii.dev/docs"
```
anthonyiscoding: This is also already defined in and should reference there so it doesn't fall out of date if docs ever moves. https://github.com/iii-hq/iii/blob/dba627d4fc95971b46aef73b5217876d72be1384/crates/scaffolder-core/src/runtime/tool.rs#L198

Sidenote the scaffolder actually needs the install links updated, it's currently wrong, can make it part of the revisions for this PR: https://github.com/iii-hq/iii/blob/dba627d4fc95971b46aef73b5217876d72be1384/crates/scaffolder-core/src/runtime/tool.rs#L197 and here where I also don't follow my very own DRY advice and proceed to hardcode the links a second time lol: https://github.com/iii-hq/iii/blob/dba627d4fc95971b46aef73b5217876d72be1384/crates/scaffolder-core/src/templates/version.rs#L89

engine/src/cli/project/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+        "iii update"
+    }
+}
+
+fn template_flow_requested(args: &InitArgs) -> bool {
+    // Template flow is triggered by template-specific flags only.
+    // --yes and --skip-iii are pass-through args that modify scaffolder
+    // behavior but do not by themselves indicate intent to use templates.
+    args.template.is_some() || args.template_dir.is_some() || args.languages.is_some()
+}
```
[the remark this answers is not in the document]
anthonyiscoding: This should also be addressed.

engine/src/cli/project/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+        device_id: Some(device_id.clone()),
+    };
+    if let Err(e) = ini.write(&root) {
+        crate::cli::telemetry::send_project_init_failed("write_project_ini", &e.to_string());
+        return print_err(
+            "could not write .iii/project.ini",
+            &e.to_string(),
+            "check that the target directory is writable",
+        );
+    }
```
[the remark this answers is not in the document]
anthonyiscoding: This should also be addressed.

engine/src/cli/project/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+            }
+        },
+        Err(_) => {
+            warn_missing_project_ini(root, "no .iii/project.ini found");
+            iii::workers::telemetry::environment::get_or_create_device_id()
+        }
+    }
+}
+
+fn warn_missing_project_ini(root: &std::path::Path, problem: &str) {
```
anthonyiscoding: I think this is unnecessary and potentially confusing to users as project.ini device_ids don't service any purpose for them

engine/src/cli/project/mod.rs
```diff
@@ -0,0 +1,387 @@
...
+        "fix:".dimmed()
+    );
+}
+
+fn resolve_root(dir: Option<&str>) -> Result<std::path::PathBuf, String> {
+    match dir {
+        Some(d) => Ok(std::path::PathBuf::from(d)),
+        None => std::env::current_dir().map_err(|e| format!("cannot read cwd: {}", e)),
+    }
+}
```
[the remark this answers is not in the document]
anthonyiscoding: This should also be addressed.

engine/src/cli/project/project_ini.rs
```diff
@@ -0,0 +1,137 @@
...
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct ProjectIni {
+    pub project_id: Option<String>,
+    pub project_name: Option<String>,
+    pub source: Option<String>,
+    pub device_id: Option<String>,
+}
+
+impl ProjectIni {
+    pub fn write(&self, project_root: &Path) -> std::io::Result<()> {
```
anthonyiscoding: There's already code that writes out project.ini and that should be used instead of this. How we define project_names and potentially other values has changed in the past and may change in the future so the same project.ini code should be used in all cases. It's here: https://github.com/iii-hq/iii/blob/dba627d4fc95971b46aef73b5217876d72be1384/crates/scaffolder-core/src/telemetry.rs#L263

If the intent is to improve the code then that is fine but please do it in telemetry.rs and related existing code.

engine/src/cli/project/scaffold.rs
```diff
@@ -0,0 +1,62 @@
+// Copyright Motia LLC and/or licensed to Motia LLC under one or more
+// contributor license agreements. Licensed under the Elastic License 2.0;
+// you may not use this file except in compliance with the Elastic License 2.0.
+// This software is patent protected. We welcome discussions - reach out at team@iii.dev
+// See LICENSE and PATENTS files for details.
+
+use std::path::Path;
+
+const CONFIG_TEMPLATE: &str = include_str!("templates/config.yaml.template");
```
anthonyiscoding: Same comments as earlier, just noting that this is another place where the hardcoding needs to be replaced with proper templating.

engine/src/cli/registry.rs
```diff
@@ -191,12 +174,11 @@ mod tests {
...
     #[test]
-    fn test_resolve_create() {
-        let (spec, sub) = resolve_command("create").unwrap();
-        assert_eq!(spec.name, "iii-tools");
-        assert_eq!(spec.repo, "iii-hq/iii");
-        assert_eq!(spec.tag_prefix, Some("iii"));
-        assert_eq!(sub, Some("create"));
+    fn test_create_no_longer_resolves() {
+        // `create` was a managed-binary dispatch to iii-tools; the command
+        // has moved to `iii project init --template`, and iii-tools is gone
```
anthonyiscoding, line 179: The command has moved, essentially been renamed and had a light functionality change but why would it be removed entirely? Is it now incorporated directly into the main iii binary?

engine/src/cli/telemetry.rs
```diff
@@ -64,6 +64,43 @@ fn build_user_properties(install_method_override: Option<&str>) -> serde_json::V
...
+            buf.clear();
+            // skip the username segment up to next '/' or whitespace
+            while let Some(&peek) = chars.peek() {
+                if peek == '/' || peek.is_whitespace() {
+                    break;
+                }
+                chars.next();
+            }
+            out.push_str("<redacted>");
+        }
```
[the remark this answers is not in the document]
anthonyiscoding: Did we not do this yet? If so then this is a great addition and whether or not other code was already handling this it should definitely be handled where we send data to Amplitude so that all occurrences of this can be sanitized both now and in the future (not just the CLI):

https://github.com/iii-hq/iii/blob/dba627d4fc95971b46aef73b5217876d72be1384/engine/src/workers/telemetry/amplitude.rs#L1

engine/src/cli/telemetry.rs
anthonyiscoding: Looks like we're duplicating this in multiple places, both of these should just live in: https://github.com/iii-hq/iii/blob/dba627d4fc95971b46aef73b5217876d72be1384/engine/src/workers/telemetry/amplitude.rs#L9

But it looks like we duplicate our API_KEY and endpoint here and in scaffolder-core. Can include this as part of the changes and may be necessary based on my next comment.

engine/src/cli/telemetry.rs
```diff
@@ -155,7 +192,34 @@ pub fn send_cli_update_failed(target_binary: &str, from_version: &str, error: &s
...
+pub fn send_project_init_succeeded(with_docker: bool, project_id: &str) {
+    if let Some(event) = build_event(
+        "project_init_succeeded",
+        serde_json::json!({
+            "with_docker": with_docker,
+            "project_id": project_id,
+        }),
+        None,
+    ) {
+        send_fire_and_forget(event);
```
anthonyiscoding, line 150: We already have amplitude code to send events and I believe this fire and forget style too, either way any updates to how it's sent and the actual sending logic should only be here: https://github.com/iii-hq/iii/blob/dba627d4fc95971b46aef73b5217876d72be1384/engine/src/workers/telemetry/amplitude.rs#L63

This one I'm a little less sure on how Rust works here but it looks like we have multiple functions for sending data to amplitude. This and any other duplicate code should be removed.

engine/tests/project_init_e2e.rs
```diff
@@ -0,0 +1,231 @@
...
+//! Exercises the real binary so subcommand routing and filesystem state are both verified.
+
+use std::process::Command;
+use tempfile::tempdir;
+
+fn iii_bin() -> Command {
+    Command::new(env!("CARGO_BIN_EXE_iii"))
+}
+
+#[test]
```
anthonyiscoding, line 25: Love to see all the tests!

As part of migrating to the template.yaml structure some of these tests can also be made generic and used to test template file creation in general.

README.md
```diff
@@ -52,7 +52,14 @@ that is both effortlessly composable, and completely observable.
 
 ## Quick Start
 
-Get started with iii by following the [Quickstart guide](https://iii.dev/docs/quickstart).
+```bash
+iii project init myapp    # scaffold a project
```
anthonyiscoding, line 56: Nice catch!

Question does `iii project init` create in current directory? Does it warn if the current directory is non-empty? And if it does warn does it have something like a `--allow-non-empty` override?

anthonyiscoding (changes requested):
Great first start! I've left comments for review and changes and CodeRabbit also had a few that I agree with and added my thoughts where necessary.

.github/workflows/create-tag.yml
guibeira: It's not; III-create is a wrapper around the scaffolder-core, no issues from templates

engine/src/cli/project/templates/iii.lock.template
guibeira: The template name helps to understand that this lock does not belong to the templates folder. Also, since it's a lock file, I don't know if no highlighting is a problem.

README.md
guibeira: > Question does iii project init create in current directory?  

Yes, when no --directory and no positional name are given.
>Does it warn if the current directory is non-empty?

No. There's no non-empty check anywhere in run_init or the scaffolder dispatch. It will silently scaffold over what's there.

engine/src/cli/telemetry.rs
guibeira: Checked the `send_event` method it's running at the same thread. This way we completely forget about this run, and unblock the main thread.

.github/workflows/create-tag.yml
guibeira: Nop, they works as same as before, the internal are the same, just the command changed

engine/src/cli/project/mod.rs
```diff
@@ -0,0 +1,531 @@
...
+                .file_name()
+                .and_then(|n| n.to_str())
+                .unwrap_or("iii-project")
+                .to_string();
+            let template_label = args.template.as_deref().unwrap_or("init-template");
+            let id = persist_project_ini(root, &project_name, template_label, &device_id)
+                .await
+                .unwrap_or_default();
+
+            if args.docker {
```
anthonyiscoding, line 258: This special casing is okay for now but make a ticket to make this generic and representable in template.yaml as part of future work, can stick it in backlog

engine/src/cli/project/mod.rs
```diff
@@ -0,0 +1,531 @@
+// Copyright Motia LLC and/or licensed to Motia LLC under one or more
```
anthonyiscoding, line 1: There's some code in here that might duplicate already existing functionality or use functionality that could be more simply modified rather than creating a new wrapper around it.

I've not looked at this exhaustively but here's Claude's analysis on possible overlap between this code and existing code:

The proposed run_init_with_template is essentially a copy of crates/iii-tools/src/main.rs::main — same IiiConfig, same CreateArgs shape, same scaffolder_core::run(...) call, same console::Term::stderr().show_cursor() panic/Ctrl+C cursor restoration. The proposed InitArgs → CreateArgs mapping duplicates impl From<CliCreateArgs> for CreateArgs.

2. IiiConfig impl — duplicated. The proposed file imports scaffolder_core::IiiConfig, but the canonical IiiConfig (with name="iii", default_template_url, etc.) is defined in crates/iii-tools/src/main.rs:14-48. If a scaffolder_core::IiiConfig already exists, the iii-tools one is the duplicate; if not, the proposed import will not resolve. Worth confirming scaffolder_core actually exports IiiConfig.

3. persist_project_ini / project.ini handling — the underlying writer already exists at crates/scaffolder-core/src/telemetry.rs::write_project_ini. The proposed wrapper (read existing id → fall back to Uuid::new_v4() → call write_project_ini) is a thin new wrapper, but the read-existing-id pattern is also present in resolve_device_id_for_docker in the same file (self-duplication within the proposal).

4. device_id resolution — iii::workers::telemetry::environment::get_or_create_device_id() already exists (engine/src/workers/telemetry/environment.rs). Reused, not duplicated. ✓

5. III_HOST_USER_ID / RABBITMQ_PASS env conventions — already established in engine/src/workers/telemetry/environment.rs:198, engine/src/workers/telemetry/mod.rs:525, crates/scaffolder-core/src/telemetry.rs:122, and engine/docker-compose.yml:44. The .env writer is consistent with these. ✓

engine/src/cli/telemetry.rs
anthonyiscoding: Great, thanks!

engine/src/cli/telemetry.rs
```diff
@@ -4,37 +4,17 @@
 // This software is patent protected. We welcome discussions - reach out at team@iii.dev
```
anthonyiscoding, line 4: Excellent work here!

engine/tests/fixtures/templates/bare/template.yaml
```diff
@@ -0,0 +1,23 @@
+name: Bare
```
anthonyiscoding, line 1: These are very likely to go out of sync with the main templates repository which makes the tests report successes when there could be failures. Need to move the template tests to be in and run from the templates repository.

If that's not possible due to other dependencies then let's chat on what to do here.

README.md
anthonyiscoding: Okay let's have it warn (really an error) and require `--allow-non-empty` to override.

anthonyiscoding (changes requested):

engine/tests/fixtures/templates/bare/template.yaml
guibeira: It's fine, since there is more to cover the test, ensure the content of the template will be there, instead of the content of the template

engine/tests/fixtures/templates/docker/Dockerfile
```diff
@@ -8,7 +8,11 @@ ENV III_HOST_USER_ID=${III_HOST_USER_ID}
 
 WORKDIR /app
 COPY config.yaml /app/config.yaml
-COPY iii.lock /app/iii.lock
+# iii.lock is engine-managed runtime state (worker registration); the engine
```
anthonyiscoding, line 11: These comments don't need to be here as they describe to the user something that doesn't exist.

anthonyiscoding (changes requested):

engine/src/cli/project/mod.rs
guibeira: Walking through the five sub-points against the current state of the branch:

**1. `run_init_with_template` is a copy of `iii-tools/src/main.rs::main`** — `crates/iii-tools/` is deleted in `c68e782c`. There's no surviving copy; the current `run_init_with_template` is the canonical place this pattern lives. The cursor-restore + ctrlc handler are necessary because scaffolder-core's TUI uses cliclack and we don't want to leave the cursor hidden on panic / Ctrl+C.

**2. `IiiConfig` duplicated** — moved into `scaffolder-core/src/iii.rs` and exposed via `scaffolder_core::IiiConfig` in `b6dd57d9`. The engine imports it; the previous in-tree copy in `iii-tools` was removed alongside the crate.

**3. `persist_project_ini` / project.ini parsing self-duplication (`read_existing_project_id` + `resolve_device_id_for_docker`)** — addressed in `8d628a3d`: extracted `read_project_ini_field(root, key)`. Both call sites now go through one parser. The `persist_project_ini` wrapper itself is intentionally thin — its only job is the merge semantic (read existing id → fall back to fresh UUID → call `scaffolder_core::telemetry::write_project_ini`). The actual writer remains scaffolder-core's.

**4 & 5. `device_id` reuse and env conventions** — agreed, already correct. No change.

Nothing remaining to do on this thread.

anthonyiscoding (approved):

PR #1612, opened 2026-05-07
fix(state,stream): support nested append paths in atomic update ops

### LINKED ISSUE #1552: BUG: `append` operation does not support nested paths
The newly introduced `append` operation does not support nested paths. Any attempt to target a nested structure either:

* Treats the path as a literal string key, or
* Fails validation, or
* Silently does nothing

This makes it impossible to append to arrays within nested objects using atomic operations.

### Expected Behavior

`append` should support nested paths so values can be appended to arrays located deeper in the object hierarchy.

For example, given a structure like:

```ts
{
  entityId: {
    buffer: []
  }
}
```

We should be able to append to `buffer` via a nested path without rewriting the entire object.

---

#### Case 1: Dot-separated path treated as literal key

```ts
await iii.trigger({
  function_id: "state::update",
  payload: {
    scope: "example::scope",
    key: "example-key",
    ops: [
      {
        type: "append",
        path: `${entityId}.buffer`,
        value: "some value",
      },
    ],
  },
});
```

**Result:**

* Creates or appends to a top-level key literally named `"entityId.buffer"`
* Does not traverse into a nested object

---

#### Case 2: Appending object at parent path

```ts
await iii.trigger({
  function_id: "state::update",
  payload: {
    scope: "example::scope",
    key: "example-key",
    ops: [
      {
        type: "append",
        path: entityId,
        value: {
          buffer: "some value",
        },
      },
    ],
  },
});
```

**Result:**

* No observable effect (append does not apply as expected)

---

#### Case 3: Array path syntax (introduced in #1547)

```ts
await iii.trigger({
  function_id: "state::update",
  payload: {
    scope: "example::scope",
    key: "example-key",
    ops: [
      {
        type: "append",
        path: [entityId, "buffer"],
        value: "some value",
      },
    ],
  },
});
```

**Result:**

* Fails with validation error: `expected string, received sequence`

deep-name:
Closes #1552.

## Summary

- `state::update`'s `append` op rejects nested paths at the type-deserialization boundary. RifkiSalim's three repros in #1552 currently behave like this against `origin/main` (`1cdc9d4a`): dotted strings key on `"entityId.buffer"` literally; parent-path appends with object values return `append.type_mismatch` (this transition shipped in #1555, see below); array-form paths fail upstream of the apply loop with `expected string, received sequence`.
- Root cause is a type asymmetry: `UpdateOp::Append.path` is `FieldPath` (single literal-string segment) at `engine/src/update_ops.rs:495` and `sdk/packages/rust/iii/src/types.rs:119`, while `UpdateOp::Merge.path` is `Option<MergePath>` (`Single | Segments`) since #1547. PR #1547 generalized only `Merge`. This PR closes the gap with the same shape.
- Switches `UpdateOp::Append.path` to `Option<MergePath>` and routes the apply loop through `validate_op_path("append", …)` so bounds + `__proto__` / `constructor` / `prototype` rejection match the merge surface. The Lua mirror in `engine/src/workers/redis.rs` is updated lockstep so the in-memory and Redis-backed paths produce identical `(old_value, new_value, errors)` tuples for the same input.

A regression test (`merge_path_single_variant_deserializes_string_first` at `sdk/packages/rust/iii/src/types.rs`) couples `MergePath`'s untagged-enum variant order to a JSON-string deserialization assertion — alphabetizing or auto-formatter reordering the variants fails the build before it can break wire compat. A second test (`append_segments_missing_leaf_string_value_becomes_array`) locks the FR-11 nested-vs-single-path divergence so the issue #1552 fix can't drift back to the legacy single-path string-concat tier.

**Note on `append.type_mismatch` for nested paths:** the structured `append.type_mismatch` error for object/scalar leaves shipped in #1555 ("Surface state update failures consistently", `@ytallo`, Apr 28) for the single-string-path case. The nested-path form added here returns the same error code with the same shape, so consumers parsing `errors[]` need no new branches. Pre-PR, an array-form path failed earlier than the apply loop — at JSON deserialization — so it never reached the type-matrix at all. Callers using `path: ""` or `path: "field"` against array, string, null, or missing-field leaves are unaffected.

The Rust SDK signature change (`UpdateOp::Append.path: FieldPath → Option<MergePath>`) is a compile-time break for downstream callers constructing the variant literally. Same posture #1547 took on `Merge`. I asked about this on #1552 ([comment](https://github.com/iii-hq/iii/issues/1552#issuecomment-4393196506)) and shipped the change with `impl From<FieldPath> for MergePath` + `impl From<&str>` shims so call sites using `.into()` keep compiling — only literal struct construction breaks. PR is **draft until `@ytallo` confirms** the public-API posture; no 24h fallback.

## What changed

**Engine** (`engine/src/`):
- `update_ops.rs` — `Append` match arm rewritten to a 4-step validation order (validate_op_path → root-is-object check → walk_or_create on parent for nested paths → leaf-type matrix). New `path_label_segments` helper formats nested paths in bracket notation `[a, b]` so error messages keep the literal-segment contract visible. 9 new unit tests cover walk-and-push, intermediate auto-creation, missing-leaf string→array (the FR-11 nested-vs-single divergence), dotted-string literal-key behavior, object-leaf type-mismatch, non-object-root rejection, `__proto__` / `constructor` / `prototype` at root / intermediate / leaf positions, and multi-op partial-failure preserving prior successes.
- `workers/redis.rs` — Lua append branch in `JSON_UPDATE_SCRIPT` consumes `merge_path_segments(op.path)` (already handled string|array|nil for merge) and routes through the same four cases. New `path_label_segments` Lua helper paralleling the Rust side keeps error messages byte-identical across adapters.

**Tests** (`engine/tests/state_stream_update_e2e.rs`):
- 7 new e2e cases through `BuiltinKvStore`: 3 RifkiSalim repros verbatim (case 1 dotted-string literal, case 2 nested-path type_mismatch, case 3 array-path happy path) + intermediate auto-creation + proto-pollution validate-before-mutate audit + multi-op partial-failure with `op_index` preservation + a constructor-level round-trip via `MergePath::Segments`.

**SDKs** (`sdk/packages/`):
- `rust/iii/src/types.rs` — new `Option<MergePath>` field shape; `From<&str>` and `From<FieldPath>` for `MergePath`; new `append`, `append_root`, `append_at_path` constructors mirroring the merge surface. `VARIANT-ORDER-LOAD-BEARING` comment above `MergePath`. 8 new round-trip + serde-default + variant-order tests.
- `node/iii/src/stream.ts` and `node/iii-browser/src/stream.ts` — `UpdateAppend.path?: MergePath` (reusing the existing `MergePath = string | string[]` alias from #1547). Browser SDK's `tests/exports.test.ts` gets 2 TS-level smoke tests covering the array-form path and omitted-path.
- `python/iii/src/iii/stream.py` — `UpdateAppend.path: str | list[str] | None = None` via pydantic smart-union. 5 new round-trip tests covering array path, omitted path, explicit `None`, empty string, and dotted-string literal.
- `console/packages/console-frontend/src/api/types/shared.ts` — adds the `append` variant to the `StreamUpdateOp` union (it was previously omitted) with the same `path?: string | string[]` shape as merge.

**Docs** (`docs/workers/iii-state.mdx`):
- Operations table's append example switched to the nested-path form. Merge + append now share a paragraph + side-by-side code examples covering root / first-level / nested paths. New explicit note on the FR-11 nested-vs-single divergence for missing leaves. Note on `append.type_mismatch` clarifying it's the same shape as #1555's surface, just reachable from nested paths now. Two new error-code rows: `append.path.too_deep` and `append.path.empty_segment` mirror the merge surface.

## Test plan

- [x] `cargo test -p iii update_ops:: --lib` — 54 passed (9 new for nested append + leaf matrix + multi-op + root-not-object + proto-pollution at three positions).
- [x] `cargo test -p iii --test state_stream_update_e2e` — 20 passed (7 new: 3 RifkiSalim repros + 4 nested-append behaviors via the KV-store path).
- [x] `cargo test -p iii-sdk --lib` — 71 passed (8 new: 5 round-trip + 1 variant-order regression + 1 serde-defaults + 1 `From<FieldPath>` compat shim).
- [x] `cargo clippy -p iii-sdk --all-targets --all-features -- -D warnings` — clean (the active CI gate).
- [x] `cargo fmt --all --check` — clean.
- [x] `pnpm --filter iii-sdk exec tsc --noEmit` — clean.
- [x] `pnpm --filter iii-browser-sdk exec vitest run tests/exports.test.ts` — 7 passed (2 new for nested + root append).
- [x] `cd sdk/packages/python/iii && uv run pytest tests/test_stream_models.py` — 11 passed (5 new round-trip).
- [x] `cd sdk/packages/python/iii && uv run mypy src` — clean.
- [x] `cd sdk/packages/python/iii && uv run ruff check src/iii/stream.py tests/test_stream_models.py` — clean (existing baseline ruff issues in the wider tree are pre-existing on `main` and unrelated to this PR).
- [ ] **Manual cross-impl parity (maintainer-side):** the Rust ↔ Lua parity invariants are documented in `engine/src/workers/redis.rs` and `engine/src/update_ops.rs` validation-order comment blocks. The repro is a 3-op fixture batch covering happy nested append + `__proto__` rejection + Case-2 type-mismatch, run through both `BuiltinKvStore` and the Redis adapter. The harness here didn't ship Redis test infra so I'm flagging this for your local run rather than leaving an `[x]` I can't honestly stand behind.

## Notes

- One concern per PR per `CONTRIBUTING.md`. `UpdateOp::Set.path`, `Remove.path`, `Increment.path`, `Decrement.path` all share the same `FieldPath` limitation but are out of scope here. Happy to take that as a follow-up if the shape lands.
- `append_at_path` is a Rust-SDK-only ergonomic helper for now (mirrors `merge_at_path` from #1547). If you'd prefer cross-SDK helper symmetry I can add Node / Python equivalents in this PR or a follow-up.
- Rebased mental model after seeing #1555 in the post-#1547 history (caught in the issue thread): #1555 already populated structured errors for append's previously-silent-no-op cases; this PR is strictly the type-shape extension to enable nested paths. No conflicts in the diff.

---

- [x] I am licensing the entirety of this PR under Apache 2 and have all necessary rights to the code I am contributing.

Ridden with [Loa](https://github.com/0xHoneyJar/loa)

## Summary by CodeRabbit

* **New Features**
  * Append now supports root, first-level, or nested-array paths; missing nested leaves are auto-created as arrays and legacy single-segment behavior is preserved. Root appends omit the path field on the wire.

* **Documentation**
  * Expanded docs and SDK docstrings to clarify path shapes, nested-append semantics, and new append error codes (append.type_mismatch, append.target_not_object, path.*).

* **Tests**
  * Added unit and end-to-end tests for nested append, dotted-string literal keys, proto-pollution rejection, partial-failure semantics, and SDK serialization/typing.

### DIFF
--- a/console/packages/console-frontend/src/api/types/shared.ts
+++ b/console/packages/console-frontend/src/api/types/shared.ts
@@ -12,6 +12,11 @@ export type StreamUpdateOp =
   // array of literal segments (nested path). Path may be omitted to
   // target the root value.
   | { type: 'merge'; path?: string | string[]; value: unknown }
+  // Append (issue #1552) mirrors merge's path shape — single string
+  // for first-level keys, array for nested paths, omitted for root.
+  // Per-op errors include `append.path.*` (validation), `append.target_not_object`,
+  // and `append.type_mismatch` (case-2: existing object/scalar at the leaf).
+  | { type: 'append'; path?: string | string[]; value: unknown }
 
 export type StreamUpdateOpError = {
   op_index: number
@@ -23,8 +28,10 @@ export type StreamUpdateOpError = {
 export type StreamUpdateResult = {
   old_value?: unknown
   new_value: unknown
-  // Per-op errors. Currently emitted only by `merge` for validation
-  // rejections (depth/size/proto-pollution). Field omitted when empty.
+  // Per-op errors. Emitted by `merge` and `append` for validation
+  // rejections (depth/size/proto-pollution) and by `append` for the
+  // case-2 `append.type_mismatch` and `append.target_not_object`
+  // surfaces. Field omitted when empty.
   errors?: StreamUpdateOpError[]
 }
 
--- a/docs/workers/iii-state.mdx
+++ b/docs/workers/iii-state.mdx
@@ -208,7 +208,7 @@ name: bridge
 
         **Append at a nested missing leaf is always an array.** When `append` walks to a missing leaf at the end of an array-form path, it creates `[value]` regardless of the value's type — including string values, which would be kept as a string under the legacy single-string path's string-concat tier. This is the core fix for [issue #1552](https://github.com/iii-hq/iii/issues/1552).
 
-        **Behavior change worth noting (issue #1552 case 2):** `append` at a path whose existing leaf is a JSON object or scalar (number/boolean) — i.e. a value that can't be appended to — now returns a structured `append.type_mismatch` error in the response `errors` array. Previously this would silently no-op when reached via the single-string path. Existing callers using `path: ""` or `path: "field"` against array, string, null, or missing-field leaves are unaffected.
+        **Note on `append.type_mismatch` for nested paths:** the structured `append.type_mismatch` error for object/scalar leaves shipped in #1555 for the single-string-path case. The nested-path form added here returns the same error code with the same shape, so consumers parsing `errors[]` need no new branches. Callers using `path: ""` or `path: "field"` against array, string, null, or missing-field leaves are unaffected.
 
         Validation: invalid update inputs are rejected with a structured error in the response's `errors` array. Reasons include path depth > 32 segments, segment > 256 bytes, value depth > 16, > 1024 top-level keys, type mismatches, non-object targets, or any segment / top-level key matching `__proto__` / `constructor` / `prototype`. Successfully applied ops still reflect in `new_value`.
       </ResponseField>
--- a/docs/workers/iii-state.mdx
+++ b/docs/workers/iii-state.mdx
@@ -182,26 +182,33 @@ name: bridge
         | `merge` | `{ "type": "merge", "path": ["sessions", "abc"], "value": { "ts": "chunk" } }` | Shallow-merge an object at the root or at any nested path. |
         | `increment` | `{ "type": "increment", "path": "count", "by": 1 }` | Add `by` to a numeric field. |
         | `decrement` | `{ "type": "decrement", "path": "count", "by": 1 }` | Subtract `by` from a numeric field. |
-        | `append` | `{ "type": "append", "path": "events", "value": { "kind": "chunk" } }` | Push one element to an array or concatenate a string value. |
+        | `append` | `{ "type": "append", "path": ["sessions", "abc", "events"], "value": { "kind": "chunk" } }` | Push one element to an array (or concatenate a string) at the root, a first-level field, or any nested path. |
         | `remove` | `{ "type": "remove", "path": "status" }` | Remove a field from the current object. |
 
-        For `set`, `increment`, `decrement`, `append`, and `remove`, paths are first-level field names. For example, `user.name` updates the field named `user.name`; it does not traverse into `{ "user": { "name": ... } }`.
+        For `set`, `increment`, `decrement`, and `remove`, paths are first-level field names. For example, `user.name` updates the field named `user.name`; it does not traverse into `{ "user": { "name": ... } }`.
 
-        For `merge`, `path` accepts either a single string (legacy / first-level field) or an array of literal segments for nested merge:
+        For `merge` and `append`, `path` accepts either a single string (legacy / first-level field) or an array of literal segments for nested traversal:
 
         ```json
-        // Root merge (existing behavior, unchanged).
+        // Root merge / append (existing behavior, unchanged).
         { "type": "merge", "path": "", "value": { "status": "active" } }
+        { "type": "append", "path": "", "value": "first" }
 
-        // First-level merge into the field named "session-abc".
+        // First-level merge / append into the field named "session-abc".
         { "type": "merge", "path": "session-abc", "value": { "author": "alice" } }
+        { "type": "append", "path": "events", "value": { "kind": "chunk" } }
 
-        // Nested merge: walks "sessions" → "abc", auto-creating
-        // missing or non-object intermediates as it goes.
+        // Nested merge / append: walks the segments, auto-creating
+        // missing or non-object intermediates along the way.
         { "type": "merge", "path": ["sessions", "abc"], "value": { "ts": "chunk" } }
+        { "type": "append", "path": ["sessions", "abc", "events"], "value": { "kind": "chunk" } }
         ```
 
         Each array element is a *literal* key. `["a.b"]` writes a single key named `"a.b"`, not `a → b`.
+
+        **Append at a nested missing leaf is always an array.** When `append` walks to a missing leaf at the end of an array-form path, it creates `[value]` regardless of the value's type — including string values, which would be kept as a string under the legacy single-string path's string-concat tier. This is the core fix for [issue #1552](https://github.com/iii-hq/iii/issues/1552).
+
+        **Behavior change worth noting (issue #1552 case 2):** `append` at a path whose existing leaf is a JSON object or scalar (number/boolean) — i.e. a value that can't be appended to — now returns a structured `append.type_mismatch` error in the response `errors` array. Previously this would silently no-op when reached via the single-string path. Existing callers using `path: ""` or `path: "field"` against array, string, null, or missing-field leaves are unaffected.
 
         Validation: invalid update inputs are rejected with a structured error in the response's `errors` array. Reasons include path depth > 32 segments, segment > 256 bytes, value depth > 16, > 1024 top-level keys, type mismatches, non-object targets, or any segment / top-level key matching `__proto__` / `constructor` / `prototype`. Successfully applied ops still reflect in `new_value`.
       </ResponseField>
@@ -238,6 +245,8 @@ Each `state::update` op may add an entry to the response `errors` array. Operati
 | `<op>.path.segment_too_long` | A path segment is longer than 256 bytes | Shorten the field name or merge path segment. |
 | `merge.path.too_deep` | A nested merge path has more than 32 segments | Reduce the nested path depth. |
 | `merge.path.empty_segment` | A nested merge path array contains an empty segment | Remove the empty segment. |
+| `append.path.too_deep` | A nested append path has more than 32 segments | Reduce the nested path depth. |
+| `append.path.empty_segment` | A nested append path array contains an empty segment | Remove the empty segment. |
 | `merge.value.not_an_object` | `merge` value is not a JSON object | Pass an object as the merge value. |
 | `merge.value.too_deep` | `merge` value has JSON nesting deeper than 16 levels | Flatten the value. |
 | `merge.value.too_many_keys` | `merge` value has more than 1024 top-level keys | Split the write into smaller updates. |
--- a/docs/workers/iii-state.mdx
+++ b/docs/workers/iii-state.mdx
@@ -206,7 +206,11 @@ name: bridge
 
         Each array element is a *literal* key. `["a.b"]` writes a single key named `"a.b"`, not `a → b`.
 
+        For root operations (no path), the SDK encoders omit the `path` field from the wire payload entirely (e.g. `{ "type": "append", "value": "first" }`) rather than emitting `"path": null`. Servers accept either form on input — `null`, missing, and empty string all route to the root.
+
         **Append at a nested missing leaf is always an array.** When `append` walks to a missing leaf at the end of an array-form path, it creates `[value]` regardless of the value's type — including string values, which would be kept as a string under the legacy single-string path's string-concat tier. This is the core fix for [issue #1552](https://github.com/iii-hq/iii/issues/1552).
+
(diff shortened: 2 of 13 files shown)


deep-name:
all green from my side after a final local sweep:

- `cargo test -p iii-sdk --lib` → 71 passed
- `cargo test -p iii update_ops:: --lib` → 54 passed (9 new)
- `cargo test -p iii --test state_stream_update_e2e` → 20 passed (7 new)
- `cargo fmt --all --check` → clean
- `cargo clippy -p iii-sdk --all-targets --all-features -- -D warnings` → clean
- `pnpm --filter iii-sdk exec tsc --noEmit` + `pnpm --filter iii-browser-sdk exec tsc --noEmit` → clean
- `pnpm --filter iii-browser-sdk exec vitest run tests/exports.test.ts` → 7 passed
- `cd sdk/packages/python/iii && uv run pytest tests/test_stream_models.py` → 11 passed
- `uv run mypy src` → clean
- `uv run ruff check src/iii/stream.py tests/test_stream_models.py` → clean

two surface-level CI items here are first-time-fork artifacts, not code issues:

- `CI` and `License Header Check` workflows show as `action_required` — they're paused waiting for a maintainer "Approve and run workflows" click (standard github fork protection). i can't trigger them myself.
- `license-agreement` failed with `Resource not accessible by integration` while trying to upsert a sticky comment on the PR. that's a known fork-PR perm shape on `pull_request_target` workflows where the integration has read-only access to the PR's head fork — separate from my code. my apache-2 grant is checked in the body directly.

ready when you are.

deep-name:
thanks @coderabbitai — the actionable finding on `engine/src/workers/redis.rs` is real and load-bearing. fixed in fb693d12.

**root cause**: `is_array({})` returned `true` (`count == max` with both 0), so on the redis-backed adapter a brand-new empty document fell into the IMP-003 root-not-object gate at `redis.rs:385` and nested append failed with `append.target_not_object`. the in-memory rust path initializes `current` to `Value::Object({})` and proceeds — issue #1552's repro happens to test against `BuiltinKvStore`, so the divergence was invisible to the e2e suite.

**fix**: empty lua tables now report as objects (`return count == max and count > 0`). this aligns `is_array` with `json_type_name`'s existing convention and lets empty documents pass the nested-append gate. cjson collapses `[]` and `{}` to the same lua table across the encode/decode roundtrip, so single-segment append-into-existing-empty-leaf now reports `append.type_mismatch` consistently — which is what the rust path's `Value::Object({})` branch already does. previous behavior was non-deterministic with respect to the original wire shape.

**doc fixes**: also picked up the three outside-diff findings — `StreamUpdateResult.errors` in all three sdks now documents that `append` emits structured per-op errors too (it does, after this PR).

note that `engine/tests/state_stream_update_e2e.rs` runs against `BuiltinKvStore`, not the redis worker, so this regression isn't caught by the existing harness. flagging for the maintainers as a follow-up — a redis-backed integration variant for the state stream tests would catch the next round of lua/rust drift before review.

ytallo:
hey @deep-name 

can you check and add these tests scenarios?


```gherkin
Feature: state::update append op behaves identically across the in-memory
  and Redis (Lua) adapters

  Background:
    Given a running Redis instance
    And a `StateRedisAdapter` connected to it
    And a fresh state scope and key

  Scenario: Appending to an existing empty array at a single-segment path
    Given the state at the key is `{ "buffer": [] }`
    When I apply the op `append("buffer", "x")`
    Then the response contains no errors
    And the new value is `{ "buffer": ["x"] }`

  Scenario: Root append when the state itself is an empty array
    Given the state at the key is `[]`
    When I apply the op `append` with no path and value `"x"`
    Then the response contains no errors
    And the new value is `["x"]`

  Scenario: Nested append where an intermediate segment points to an array
    Given the state at the key is `{ "a": [1, 2, 3] }`
    When I apply the op `append` at path `["a", "b"]` with value `42`
    Then one of the following holds:
      | outcome           | errors            | new value                    |
      | Rust-parity       | empty             | { "a": { "b": [42] } }       |
      | Strict / explicit | non-empty         | { "a": [1, 2, 3] } unchanged |
    But the new value MUST NOT be a corrupted mixed-key object such as
      `{ "a": { "1": 1, "2": 2, "3": 3, "b": [42] } }`

  Scenario: Nested append happy path through the Lua script
    Given no state at the key
    When I apply the op `append` at path `["session", "abc", "events"]` with value `"first"`
    Then the response contains no errors
    And the new value is `{ "session": { "abc": { "events": ["first"] } } }`
```

```gherkin
Feature: Cross-SDK wire-format consistency for the `path` field on
  append ops

  Scenario: Root append serializes without an explicit `null` path
    Given a `UpdateOp::append_root("first")`
    When I serialize it to JSON
    Then the resulting object MUST NOT contain a `path` key
    And the resulting object equals `{ "type": "append", "value": "first" }`

  # Alternative if you prefer the TS-side fix:
  #   Then the TS type `UpdateAppend.path` accepts `null`
  #   And the type-test `const op: UpdateAppend = { type: "append", path: null, value: "x" }` compiles
```

---

deep-name:
on it @ytallo — these are sharp scenarios. quick read on each so you know what i'm planning:

**parity feature (5 scenarios):**
- A (`{"buffer": []}` + single-segment append): the cjson `[]` vs `{}` ambiguity i mentioned in the previous comment is exactly what bites here. fix: opt into `cjson.decode_array_with_array_mt(true)` so empty arrays carry the marker, then check `getmetatable(v) == cjson.array_mt` for the empty-table case in `is_array`. preserves the empty-document fix without re-breaking this case.
- B (root append on `[]`): falls out of the same fix.
- C (`{"a": [1,2,3]}` + nested `["a","b"]`): rust's `walk_or_create` already replaces non-object intermediates with `{}` (rust-parity outcome `{"a": {"b": [42]}}`, destructive). lua's `walk_or_create` currently walks INTO the array and produces the corrupted mixed-key form you flagged. fix: detect array intermediates in lua and replace with `{}` to match rust. opting for rust-parity over strict so we don't ship a behavior break for an already-shipped surface.
- D (lua nested happy path): covered by my last commit — adding the redis-backed test now.

**wire-format feature:**
- yes, `{"type":"append","path":null,...}` is what rust/python currently emit. fixing both sides to skip the field when `None` so the JSON matches the gherkin (`{"type":"append","value":"first"}`). will do the same for `merge` for consistency with the `MergePath` precedent in #1555.

infrastructure note: e2e tests so far are `BuiltinKvStore`-only (in-memory). spinning up a redis container locally to add a `state_stream_update_redis_e2e.rs` lane gated on `TEST_REDIS_URL` so future contributors can keep the rust/lua paths honest. should have a commit batch up shortly.

deep-name:
@ytallo all five scenarios are in (commit `b482902c`).

**parity feature:**
- **A** (`{"buffer": []}` + single-segment append): split the empty-table heuristic in two. `is_array` (used inside `append_to_target`) keeps the legacy "empty table is appendable as array" semantic so `{"buffer": []}` works on the cjson roundtrip; `is_object_shape` (used at the IMP-003 root gate and inside `walk_or_create`) follows `json_type_name`'s `value[1] ~= nil` convention so empty documents pass nested-append. The empty-`{}` existing leaf case is documented as a known cjson roundtrip limitation; redis cjson doesn't expose `array_mt` so we can't fully disambiguate. ✅
- **B** (root append on `[]`): falls out of the same fix. ✅
- **C** (`{"a": [1,2,3]}` + nested `["a","b"]`): rewrote lua's `walk_or_create` to replace any non-object intermediate (null, scalar, OR array) with `{}`, matching rust's `*entry = Value::Object(Map::new())` at `engine/src/update_ops.rs:231-233`. went with rust-parity over strict-error since that's already the merge precedent and would otherwise be a behavior break for an already-shipped surface. negative assertion against the corrupted mixed-key form is in both test lanes. ✅
- **D** (lua nested happy path): covered by reordering `is_object_shape` ahead of `walk_or_create` so the upvalue resolves at definition time. ✅

**wire-format feature:**
- rust SDK: `#[serde(default, skip_serializing_if = "Option::is_none")]` on `UpdateOp::Append.path` and `UpdateOp::Merge.path`. root ops now serialize as `{"type":"append","value":"first"}`. existing `path: null` payloads still deserialize via `#[serde(default)]`. ✅
- python SDK: `@model_serializer(mode="wrap")` on `UpdateAppend` and `UpdateMerge` drops `path` when `None`. byte-identical wire to rust. ✅
- node + browser SDKs: already used `path?: MergePath`, so the new wire format (no `path` key) is what their types already model — no change needed.

**tests:**
- `engine/tests/state_stream_update_e2e.rs`: 3 new scenarios through `BuiltinKvStore` (rust path).
- `engine/tests/state_stream_update_redis_e2e.rs` (new): redis-backed parity lane gated on `TEST_REDIS_URL` (default `redis://localhost:6379`). same 4 scenarios against `StateRedisAdapter` so future drift is caught locally. verified against `redis:7-alpine`.
- python: `path: null` round-trip + back-compat-on-input tests for both `UpdateAppend` and `UpdateMerge`.
- rust SDK: serialization tests now assert no `path` key in encoded output for root ops.

**total sweep:** 71 SDK + 54 update_ops + 23 in-memory e2e + 4 redis-parity + 7 browser exports + 11 python pydantic, all green. cargo fmt + clippy clean for touched files; mypy + ruff clean.

ready when you are.

ytallo:

Please reply with:

deep-name:
I agree that my contributions in this PR to the engine code are licensed under Apache 2.0.

ytallo (approved):

PR #1618, opened 2026-05-11
fix(observability): preserve OTLP resource attributes on forwarder hop

### LINKED ISSUE #1617: OTLP span forwarding drops resource attributes, causing service.name to appear as <nil>
### Summary

When SDK spans are forwarded through `iii-engine` to an OTLP collector, the resource attributes from the original OTLP payload appear to be dropped. As a result, collectors such as SigNoz show the service name as `<nil>`, even though spans and logs are otherwise collected successfully.

### Problem

SDK spans sent through the engine are received by the OTLP collector, but the service name is missing.

For example, when running with:

```bash
OTEL_SERVICE_NAME=my-service
```

and forwarding spans through the engine to SigNoz, the traces are visible, but the service name is displayed as:

```text
<nil>
```

### Suspected Cause

The issue appears to be in:

```text
engine/src/workers/observability/otel.rs
```

In `ingest_otlp_json`, incoming OTLP JSON payloads are converted using `convert_otlp_to_span_data` before being forwarded.

However, the resulting `SpanData` structure does not appear to include the OTLP resource information from:

```text
resource_spans[].resource
```

This resource object is where attributes such as `service.name` are stored. Because that resource data is dropped during conversion, the forwarded spans no longer carry the original service metadata.

Additionally, `SDK_SPAN_FORWARDER` appears to be a bare `SpanExporter` rather than being attached to a `TracerProvider`, so there does not seem to be a later opportunity to reattach or reconstruct the missing resource.

### Steps to Reproduce

1. Configure an SDK/application with a service name, for example:

   ```bash
   OTEL_SERVICE_NAME=my-service
   ```

2. Send SDK spans through `iii-engine`.

3. Configure the engine to forward spans to an OTLP collector such as SigNoz.

4. Open the traces in the collector UI.

### Expected Behavior

The forwarded spans should preserve the original OTLP resource attributes, including:

```text
service.name = my-service
```

The collector should display the service name correctly.

### Actual Behavior

The traces are collected successfully, but the service name appears as:

```text
<nil>
```

### Suggested Fix

Instead of converting the incoming OTLP JSON payload into `SpanData` and then exporting it again through `SpanExporter::export`, consider forwarding the original OTLP payload directly to the configured collector endpoint as a raw HTTP relay.

This would preserve the original OTLP structure, including:

```text
resource_spans[].resource
```

and avoid losing resource-level attributes such as `service.name`.

### Additional Context

I am not completely sure whether the current conversion behavior is intentional, but it seems surprising because resource attributes are part of the OTLP span payload and are required by collectors to identify the service that produced the spans.

deep-name:
## Summary

Fixes iii-hq/iii#1617 — OTLP span forwarding drops `resource_spans[].resource`, so SDK spans reach the collector with `service.name=<nil>` even when `OTEL_SERVICE_NAME` is set at the SDK.

## Why

`SpanData` carries no resource — in the OTel SDK contract, the owning `TracerProvider` supplies one before the exporter serialises. `SDK_SPAN_FORWARDER` was a bare `opentelemetry_otlp::SpanExporter` with no `TracerProvider` wrapping it, so `convert_otlp_to_span_data → SpanExporter::export` wrote an empty resource onto the wire. Collectors rendered the missing service identity as `<nil>` (SigNoz/Grafana/etc).

@alsruf36's diagnosis on the issue thread was exactly right; the fix shape follows their suggestion.

## What changed

`engine/src/workers/observability/otel.rs`
- `SDK_SPAN_FORWARDER`: `OnceLock<Arc<SpanExporter>>` → `OnceLock<tonic::Channel>`. Cheap to clone (Arc internally), lazy connect, no init-time blocking.
- `init_sdk_span_forwarder`: rewritten to build a `tonic::Endpoint` + `connect_lazy()`. For `https://` endpoints, applies `ClientTlsConfig::new().with_native_roots()` so the previous deployment contract (handled implicitly by `opentelemetry-otlp`'s `tls-roots` feature) is preserved.
- Forwarder branch in `ingest_otlp_json` (lines 1607-1635): translates the parsed `OtlpExportTraceServiceRequest` into `opentelemetry_proto::tonic::collector::trace::v1::ExportTraceServiceRequest` via field-for-field helpers (`otlp_json_request_to_proto` and friends, lines 1640-1834), then sends via a freshly-cloned `TraceServiceClient::new(channel.clone())`. Resource block survives byte-for-byte.
- Deleted `convert_otlp_to_span_data` and `otlp_kv_to_key_value` (truly dead post-replacement; in-memory storage path does its own per-field conversion) and their unit tests.
- Trimmed one assertion in `test_ingest_otlp_json_skips_spans_with_invalid_identifiers` — malformed-span filtering was a side effect of the old SpanData path, not a contract. New forwarder relays and lets the collector validate. Memory-storage behaviour is unchanged.

`engine/Cargo.toml`
- Promote `opentelemetry-proto = "0.31"` (features `["gen-tonic"]`), `tonic = "0.14"` (features `["transport", "tls-native-roots"]`), `tokio-stream = "0.1"` to direct deps. All three were already transitively present via `opentelemetry-otlp`'s `grpc-tonic` feature — promotion adds nothing new to the dep tree.

`engine/tests/otel_forwarder_resource_test.rs` (new)
- Integration test against a tonic gRPC mock collector. Owns its own process so the global `SDK_SPAN_FORWARDER` `OnceLock` is fresh.
- Asserts `service.name=my-service`, `service.namespace=production`, `service.version=1.4.2` survive the hop (three resource attributes locked in, not just one).
- Asserts `replica.count=7` (intValue) and `telemetry.enabled=true` (boolValue) — covers the proto-conversion's non-string AnyValue branches.
- Asserts `flags & 0x01 == 0x01` for an inbound payload that omits `flags` — pins the SAMPLED default that mirrors the storage path at `otel.rs:1521-1524`.

## Behaviour change worth noting

The new forwarder relays spans with malformed trace/span IDs (the old SpanData path pre-filtered them via `TraceId::from_hex`/`SpanId::from_hex` bailout). Downstream collectors validate and reject — engine-side pre-filtering hid that diagnostic signal. The in-memory storage path's filtering is unchanged. Called out in the source comment at `engine/src/workers/observability/otel.rs:1607-1622`.

## Test plan

- [x] `cargo test -p iii --test otel_forwarder_resource_test` — 1 passed (with extended assertions for string + int + bool resource attrs + flags=SAMPLED default)
- [x] `cargo test -p iii --lib` — 1542 passed, 0 failed, 6 ignored (KVM-gated)
- [x] `cargo test -p iii --lib workers::observability` — 289 passed
- [x] `cargo test -p iii --lib test_ingest_otlp_json` — 10 passed (including the trimmed `test_ingest_otlp_json_skips_spans_with_invalid_identifiers`)
- [x] `cargo fmt --check` — clean
- [x] `cargo clippy -p iii --lib --tests` — no new warnings on touched files
- [ ] Manual: run against a real collector with `OTEL_SERVICE_NAME=my-service`, confirm UI shows the service name (rather than `<nil>`)
- [ ] Manual: run against an `https://` collector with valid certs, confirm TLS handshake succeeds (verified locally against a self-signed loopback; full prod-collector cert validation deferred to CI / reviewer)

## Credits

@alsruf36 — diagnosis + suggested fix shape on the issue thread.

Closes #1617.

## Summary by CodeRabbit

* **Bug Fixes**
  * Preserve full incoming OTLP resource attributes when forwarding traces to collectors.

* **Improvements**
  * Forward OTLP traces as native proto over gRPC with injected OTLP headers.
  * Add native TLS handling for HTTPS collector endpoints.
  * Drop spans with malformed trace/span IDs from forwarded batches while retaining in-memory ingestion.

* **Tests**
  * Added integration tests verifying resource preservation, header propagation, and partial-batch behavior.

* **Chores**
  * Updated build/dev dependencies for OTLP and async streaming.

[Review Change Stack](https://app.coderabbit.ai/change-stack/iii-hq/iii/pull/1618)


deep-name (commented):
## Summary

Analytical review of iii-hq/iii#1618. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Heavy new dependencies (tonic, opentelemetry-proto, tokio-stream) added at top-level Cargo.toml for test-only use",
      "severity": "MEDIUM",
      "category": "dependencies",
      "file": "engine/Cargo.toml",
      "description": "The PR adds `opentelemetry-proto`, `tonic` (with `transport` and `tls-native-roots`), and `tokio-stream` to the main `[dependencies]` section. From the visible diff, these are only consumed by an integration test (`engine/tests/otel_forwarder_resource_test.rs`). Putting them in `[dependencies]` instead of `[dev-dependencies]` enlarges the production binary closure (tonic with TLS pulls in rustls/native-roots, prost, hyper) and lengthens compile times for every consumer of the engine crate. If the forwarder source itself doesn't actually import these (the otel.rs section is truncated so this can't be fully verified), they should be `[dev-dependencies]`.",
      "suggestion": "Confirm whether the forwarder implementation in `engine/src/workers/observability/otel.rs` directly uses `opentelemetry-proto`, `tonic`, or `tokio-stream`. If not (i.e. only the test does), move all three to `[dev-dependencies]`. If `opentelemetry-proto` is genuinely needed in production for proto conversion, keep just that one in `[dependencies]` and move `tonic`/`tokio-stream` to dev.",
      "confidence": 0.6
    },
    {
      "id": "F2",
      "title": "Process-global OnceLock forwarder state makes integration test order-dependent",
      "severity": "MEDIUM",
      "category": "testing",
      "file": "engine/tests/otel_forwarder_resource_test.rs",
      "description": "The header comment states the test relies on `SDK_SPAN_FORWARDER` being a process-global `OnceLock`, fresh per integration test file. This is true today but fragile: if a future test file in the same `tests/` directory also calls `init_sdk_span_forwarder`, or if cargo test harness changes (e.g., a future shared test binary, or someone adds a `#[test]` to the same file pointing at a different endpoint), the OnceLock will silently bind the wrong endpoint and the second test will export to the wrong collector. There is no documented protection nor a test-only reset hook.",
      "suggestion": "Either (a) add a `#[cfg(test)] pub fn reset_sdk_span_forwarder_for_tests()` that swaps the OnceLock contents, or (b) refactor the forwarder to accept an injected endpoint/exporter so tests don't depend on process-global state, or (c) at minimum add a code comment on `SDK_SPAN_FORWARDER` itself warning future integration tests not to call `init_sdk_span_forwarder` in the same binary.",
      "confidence": 0.7
    },
    {
      "id": "F3",
      "title": "50ms sleep for collector startup is a flake vector",
      "severity": "LOW",
      "category": "testing",
      "file": "engine/tests/otel_forwarder_resource_test.rs",
      "description": "`spawn_mock_collector` sleeps 50ms unconditionally before returning, with a comment acknowledging this exists to avoid races with tonic's accept loop. Loaded CI runners (especially under heavy parallel cargo test) can easily exceed 50ms before the spawned task is polled. The downstream poll loop (5s deadline) does provide a fallback safety net for the actual export, but the initial race could still cause connection-refused on the first export attempt and rely on tonic's internal retry — and that retry is not guaranteed across tonic versions.",
      "suggestion": "Replace the sleep with an active readiness check: in a small loop, attempt to `TcpStream::connect(addr)` until success or a bounded timeout (~1s). Alternatively, bind the listener synchronously *before* the spawn and only spawn the `serve_with_incoming` future — `TcpListener::bind` already guarantees the port is listening, so the race is solely about the tonic accept-loop being polled. A `tokio::task::yield_now().await` followed by a connect-probe loop would be deterministic.",
      "confidence": 0.6
    },
    {
      "id": "F4",
      "title": "Mock collector server task swallows panics on `unwrap()`",
      "severity": "LOW",
      "category": "testing",
      "file": "engine/tests/otel_forwarder_resource_test.rs",
      "description": "The spawned `Server::builder()...serve_with_incoming(...).await.unwrap()` runs in a detached `tokio::spawn`. If the server fails to start or errors out, the `.unwrap()` will panic inside the spawned task, but the test harness will see the export simply never arrive and surface as `captured.len() == 0` with a misleading message. The task handle is not awaited or stored.",
      "suggestion": "Either retain the JoinHandle and check it on test failure, or replace `.unwrap()` with `.expect(\"mock collector server failed to start\")` plus a `tracing::error!` — at minimum so the panic surfaces in test output rather than silently disappearing into a detached task.",
      "confidence": 0.5
    },
    {
      "id": "F5",
      "title": "Test asserts SAMPLED default but contract is undocumented in PR scope",
      "severity": "LOW",
      "category": "testing",
      "file": "engine/tests/otel_forwarder_resource_test.rs",
      "description": "The final assertion (`span.flags & 0x01 == 0x01`) pins the forwarder to default missing trace-flags to SAMPLED, with rationale that it 'mirrors the in-memory storage path'. This is a meaningful behavioural contract, but it is being introduced alongside a fix titled 'preserve OTLP resource attributes' — the trace-flags default is a separate concern. If this default is new behaviour added in this PR, it deserves explicit call-out in the PR description; if it's pre-existing behaviour now being locked in, that's fine but worth noting. The OTLP spec leaves trace flags semantics to the sender; a forwarder defaulting unsampled→SAMPLED could subtly amplify sampled trace volume downstream.",
      "suggestion": "Confirm in the PR description whether the SAMPLED default is (a) pre-existing and being regression-locked, or (b) new in this PR. If (b), consider whether defaulting to whatever the inbound payload says (0 when absent) is more correct per OTLP semantics, since collectors typically interpret missing flags as 'unspecified' rather than 'unsampled'.",
      "confidence": 0.5
    },
    {
      "id": "F6",
      "title": "Test only asserts forwarded payload, not absence of duplication or modification",
      "severity": "LOW",
      "category": "testing",
      "file": "engine/tests/otel_forwarder_resource_test.rs",
      "description": "The test asserts `captured.len() == 1` and that 5 attributes are correctly preserved. It does not assert that *no extra* attributes were injected by the forwarder (e.g. a hop-level `forwarder.id` attribute), nor that `scope_spans`/spans are passed through unmodified beyond the flags check. A forwarder bug that *adds* attributes or rewrites span names would pass this test.",
      "suggestion": "Consider asserting `resource.attributes.len() == 5` and `req.resource_spans[0].scope_spans[0].spans[0].name == \"forwarder.preserves.resource\"` to lock in pass-through fidelity, not just preservation of the named subset.",
      "confidence": 0.55
    },
    {
      "id": "F7",
      "title": "Net deletion of ~374 lines in otel.rs alongside 'fix' is non-obvious",
      "severity": "SPECULATION",
      "category": "review-visibility",
      "file": "engine/src/workers/observability/otel.rs",
      "description": "The truncated diff shows `engine/src/workers/observability/otel.rs` with +323 -697, a net deletion of ~374 lines, in a PR framed as a bug fix for resource-attribute preservation. A fix of this magnitude typically suggests a refactor or rewrite of the forwarder path rather than a surgical fix. Without seeing the source diff, I cannot evaluate whether the deletion removes dead code, consolidates a parser, replaces a hand-rolled protobuf encoder with `opentelemetry-proto`, or removes functionality. If it removes the hand-rolled OTLP/JSON→proto translation in favour of using `opentelemetry-proto` types directly, that's likely a good change but is much larger than the PR title implies.",
      "suggestion": "Request the full diff or a summary in the PR description explaining the structural change: is this replacing a custom proto encoder with `opentelemetry-proto::tonic` types? Are any code paths (e.g. backpressure, retry, batching) being removed in the process? Reviewers should not approve a -374-line change on the strength of one regression test without seeing what was removed.",
      "confidence": 0.7
    },
    {
      "id": "F8",
      "title": "Regression test design with full AnyValue branch coverage and clear failure messages",
      "severity": "PRAISE",
      "category": "testing",
      "file": "engine/tests/otel_forwarder_resource_test.rs",
      "description": "The test deliberately exercises three `AnyValue` discriminator branches (stringValue, intValue, boolValue), each with an assertion message that explains the bug class being guarded against rather than just the expected value. The header comment also explicitly states why the test is in a separate integration file (process-global OnceLock isolation) and why the assertion contract is 'full resource preservation, not just service.name' — preventing a future tempting single-field band-aid from passing this test. This is the right shape for a regression test tied to a specific incident.",
      "suggestion": "No action needed."
    }
  ]
}
```


## Callouts

_Enrichment unavailable for this review._

deep-name:
local self-review summary before maintainer queue:

ran our own bridgebuilder pass (claude-opus-4-7 via custom reviewer persona) plus two rounds of adversarial cross-model dissent (gpt-5.3-codex for both review + audit phases). 4 model-passes total, 3 distinct concerns surfaced + addressed before this hit your queue:

| Round | Tool | Model | Verdict |
|---|---|---|---|
| review-sprint round 1 | adversarial-review (review) | gpt-5.3-codex | BLOCKING: DISS-001 (TLS regression for `https://` collectors), DISS-002 (trace-flags default mismatch) |
| review-sprint round 2 | adversarial-review (review) | gpt-5.3-codex | clean — both blockers resolved |
| audit-sprint | adversarial-review (audit, no reviewer context) | gpt-5.3-codex | clean |
| local-self-review | bridgebuilder | claude-opus-4-7 | F1-F4 + F6 addressed in commit `1bc8af32` (dep placement, OnceLock test isolation, sleep→connect-probe, .unwrap→.expect, pass-through fidelity assertions) |

artifacts kept in our state zone for traceability:
- `grimoires/loa/a2a/sprint-bug-1/engineer-feedback.md`
- `grimoires/loa/a2a/sprint-bug-1/auditor-sprint-feedback.md`
- `grimoires/loa/a2a/sprint-bug-1/adversarial-review.json`
- `grimoires/loa/a2a/sprint-bug-1/adversarial-audit.json`
- the bridgebuilder review above (8 findings, of which 5 actionable, 1 praise, 2 already-addressed)

ready when you are. happy to drop the bridgebuilder bot comment above if it's noisy — it's preserved here as the audit trail showing the issues were caught locally, not surfaced by your team's review cycle.

deep-name:
status update for the maintainer queue:

- @alsruf36 (issue reporter) pulled this branch and validated end-to-end against their SigNoz instance: https://github.com/iii-hq/iii/issues/1617#issuecomment-4418061611. forwarder hop is doing exactly what it says on the tin.
- they surfaced a related but distinct bug — engine-originated spans (`SdkTracerProvider` export path, not the forwarder hop) also show empty `service.name`. that's out of scope here and filed as #1619 with their flame-graph evidence + a code-level read on the likely root cause (`OtelConfig::default()` short-circuits the env-var fallback when config has `service_name: ""`).

#1618 stays scoped to the forwarder. ready for review.

ytallo:
Hey @deep-name, thanks for your contribution one more time. Can you consider add these tests on your PR?


```gherkin
Feature: SDK span forwarder honors OTEL_EXPORTER_OTLP_HEADERS
  As an operator running iii behind an auth-required OTLP collector
    (Grafana Cloud, Honeycomb, SigNoz Cloud, or any API-keyed deployment)
  I want SDK-ingested spans forwarded by the engine to carry the same
    auth metadata as engine-originated spans
  So that worker-forwarded telemetry does not silently get rejected at
    the collector while engine telemetry keeps flowing

  Background:
    Given a mock OTLP gRPC collector listening on an ephemeral loopback port
    And the collector captures the tonic MetadataMap of every Export request

  Scenario: Authorization header from env var lands on the wire
    Given the environment variable OTEL_EXPORTER_OTLP_HEADERS is set to
      "authorization=Bearer-test-token-12345"
    And the SDK span forwarder has been initialized against the mock collector
    When a well-formed OTLP/JSON payload with one valid span is ingested via
      ingest_otlp_json
    Then the mock collector receives exactly one ExportTraceServiceRequest
    And that request carries an "authorization" metadata entry
    And
```

```gherkin
Feature: One bad span does not poison the rest of the forwarded batch
  As an operator whose SDKs occasionally emit malformed trace/span IDs
    (instrumentation bugs, truncated context propagation, custom emitters)
  I want the forwarder to drop only the malformed spans
  So that valid telemetry in the same payload still reaches the collector
    instead of being rejected wholesale when one span has empty ID bytes

  Background:
    Given a mock OTLP gRPC collector listening on an ephemeral loopback port
    And the collector captures every ExportTraceServiceRequest it receives

  Scenario: Mixed batch of valid and malformed spans is partially forwarded
    Given the SDK span forwarder has been initialized against the mock collector
    And an OTLP/JSON payload contains two spans:
      | name           | trace_id                          | span_id            | validity |
      | bad-trace-id   | not-hex                           | also-not-hex       | invalid  |
      | valid-span     | 0af7651916cd43dd8448eb211c80319c  | b7ad6b7169203331   | valid    |
    When the payload is ingested via ingest_otlp_json
    Then the mock collector receives at least one ExportTraceServiceRequest
    And no span on the wire has an empty trace_id
    And no span on the wire has an empty span_id
    And the captured stream contains exactly one span
    And that span has name "valid-span"
    And that span has trace_id 0af7651916cd43dd8448eb211c80319c
    And that span has span_id b7ad6b7169203331

```

deep-name:
both scenarios are spot-on @ytallo — these are the next two regression classes the raw-tonic switch introduced beyond TLS. on it.

quick read so you know what's landing:

**scenario 1 (OTEL_EXPORTER_OTLP_HEADERS)**: confirmed regression. `opentelemetry_otlp::SpanExporter::builder().with_tonic()` parses this env var automatically (per the OTel spec) and injects each `key=value` pair as a tonic `MetadataMap` entry on every export. our replacement channel doesn't. fix: parse the env var at `init_sdk_span_forwarder` time, stash the parsed pairs alongside the channel, attach a tonic interceptor that adds them to each `TraceServiceClient::export` request.

**scenario 2 (one bad span doesn't poison the batch)**: the audit-sprint feedback called out the "relay malformed spans, let collector validate" tradeoff as deliberate, but you're right that empty-byte IDs on the wire are a regression vs the pre-fix `convert_otlp_to_span_data` filtering, and "valid spans get rejected with the batch" is a worse outcome than "malformed spans dropped silently". fix: `otlp_span_to_proto` becomes fallible — `Option<ProtoSpan>`, returns `None` when `hex::decode` fails on `trace_id`/`span_id`. caller drops `None`s before building the `ExportTraceServiceRequest`. PR description gets updated to remove the "relays malformed spans" callout since we're restoring the pre-filter.

shipping both scenarios as runnable rust tests against the existing tonic mock + an HTTPS-fronted mock variant for header propagation. should have the commit batch up in the next hour or so.

deep-name (commented):
## Summary

Analytical review of iii-hq/iii#1618. Enrichment pass was unavailable; findings are unenriched.

## Findings


```json
{
  "schema_version": 1,
  "findings": [
    {
      "id": "F1",
      "title": "Integration tests rely on global SDK_SPAN_FORWARDER OnceLock and cannot coexist",
      "severity": "MEDIUM",
      "category": "testing",
      "file": "engine/tests/otel_forwarder_resource_test.rs",
      "description": "Both new integration test files (otel_forwarder_resource_test.rs and otel_forwarder_partial_batch_test.rs, plus the pre-existing otel_forwarder_headers_test.rs) call init_sdk_span_forwarder(&endpoint) which writes to a process-global OnceLock. The comment in each file asserts that 'Owns its own process via being an integration test file, so the process-global SDK_SPAN_FORWARDER OnceLock is fresh on every run.' This is true only because cargo compiles each integration test file in tests/ as a separate binary. However, if any test file in the future contains more than one #[tokio::test] that calls init_sdk_span_forwarder with a different endpoint, all subsequent tests will silently send to the first endpoint (OnceLock semantics) and the assertions about 'exactly one ExportTraceServiceRequest' will become non-deterministic across runs. There is no test-level guardrail (no serial_test marker, no assertion that the OnceLock was uninitialized) preventing future contributors from breaking this invariant.",
      "suggestion": "Either (a) add a doc-comment near init_sdk_span_forwarder warning that the OnceLock requires one-test-per-binary, with a brief explanation, or (b) expose a #[cfg(test)] reset helper so multiple tests per binary become safe. Option (b) is more robust.",
      "confidence": 0.75
    },
    {
      "id": "F2",
      "title": "Mock collector spawn task swallows panics on Server::builder errors after readiness",
      "severity": "LOW",
      "category": "testing",
      "file": "engine/tests/otel_forwarder_resource_test.rs",
      "description": "The tokio::spawn task that runs the tonic Server uses .expect() on serve_with_incoming. If the server errors mid-test (e.g., port reclaimed, OS resource exhaustion), the spawned task panics in the background but the main test continues, producing confusing 'received 0 requests' failures rather than the underlying transport error. Same pattern in otel_forwarder_partial_batch_test.rs.",
      "suggestion": "Capture the JoinHandle and abort()/check it at end-of-test, or use a oneshot channel to surface server startup errors to the main test thread.",
      "confidence": 0.6
    },
    {
      "id": "F3",
      "title": "Readiness probe leaves a stray connect socket against tonic accept loop",
      "severity": "LOW",
      "category": "testing",
      "file": "engine/tests/otel_forwarder_resource_test.rs",
      "description": "The TCP connect-then-drop readiness probe at line ~62 establishes a real TCP connection to the tonic server that is then dropped without an HTTP/2 handshake. tonic/h2 may log a spurious warning ('connection error: not a valid HTTP/2 connection preface' or similar) on stderr for each test. Not a correctness bug but adds noise to test output.",
      "suggestion": "Acceptable as-is; if the noise becomes problematic, switch to a lightweight in-process readiness signal (oneshot sent from inside the spawn before serve_with_incoming).",
      "confidence": 0.4
    },
    {
      "id": "F4",
      "title": "Excellent regression-test design with explicit contract assertions",
      "severity": "PRAISE",
      "category": "testing",
      "file": "engine/tests/otel_forwarder_resource_test.rs",
      "description": "The resource-preservation test pins not only the originally-broken service.name field but also tests three AnyValue discriminator branches (string, int, bool), span-name pass-through, attribute count (catches injection), and SAMPLED-flag default behavior. The partial-batch test goes further: it pins both hex-decode-failure AND hex-decodes-but-wrong-length cases, citing the spec lengths (16 bytes / 8 bytes) explicitly. Failure messages reference issue numbers and explain the intended invariant, making future debugging significantly easier.",
      "suggestion": "No change required."
    },
    {
      "id": "F5",
      "title": "Production tonic dependency added with tls-native-roots — confirm intent",
      "severity": "LOW",
      "category": "dependencies",
      "file": "engine/Cargo.toml",
      "description": "tonic is added to the [dependencies] section (not [dev-dependencies]) with features = [\"transport\", \"tls-native-roots\"]. The new test files use tonic, but they only need tonic at dev-time. If the forwarder implementation in the truncated otel.rs needs tonic directly (not just transitively through opentelemetry-otlp which already pulls tonic), that is fine. Otherwise, this is dependency bloat and should move to [dev-dependencies]. Cannot fully verify without the truncated otel.rs diff.",
      "suggestion": "Verify whether engine/src/workers/observability/otel.rs imports tonic directly. If only the integration tests need it, move tonic + opentelemetry-proto to [dev-dependencies].",
      "confidence": 0.55
    },
    {
      "id": "F6",
      "title": "Typo in test comment",
      "severity": "LOW",
      "category": "nit",
      "file": "engine/tests/otel_forwarder_resource_test.rs",
      "description": "'preceeding' is misspelled (should be 'preceding') in the comment block at the attribute-count assertion.",
      "suggestion": "Fix spelling.",
      "confidence": 0.95
    },
    {
      "id": "F7",
      "title": "Duplicated mock-collector helper across three test files",
      "severity": "LOW",
      "category": "maintainability",
      "file": "engine/tests/otel_forwarder_partial_batch_test.rs",
      "description": "CapturingTraceService, TraceService impl, and spawn_mock_collector are duplicated verbatim across otel_forwarder_resource_test.rs, otel_forwarder_partial_batch_test.rs, and (per the truncated diff) otel_forwarder_headers_test.rs. Cargo integration tests can share code via a tests/common/mod.rs module; this would also enforce that the readiness/spawn semantics stay in lockstep across the suite.",
      "suggestion": "Extract the mock collector into engine/tests/common/mod.rs and have each integration test `mod common;` it.",
      "confidence": 0.85
    },
    {
      "id": "F8",
      "title": "Truncated otel.rs change is the actual fix — cannot verify behavior without it",
      "severity": "REFRAME",
      "category": "review-scope",
      "file": "engine/src/workers/observability/otel.rs",
      "description": "The substantive fix (449 added, 696 removed lines in engine/src/workers/observability/otel.rs) is truncated from the diff. The tests pin behavior but the production change itself — including the `otlp_span_to_proto` returning Option, the resource conversion path, and the SAMPLED-default logic — is unreviewed. Findings here cover only the test/dependency surface; the core fix needs a separate pass.",
      "suggestion": "Request the full otel.rs diff for the next review pass to verify: (1) hex-decode length checks on trace_id (16) and span_id (8), (2) resource attribute conversion preserves all AnyValue variants, (3) SAMPLED-default flag handling, (4) no unwrap()/expect() on adversarial input.",
      "confidence": 0.9
    }
  ]
}
```


## Callouts

_Enrichment unavailable for this review._

deep-name:
both gherkin scenarios closed end-to-end @ytallo. round summary across 2 commits:

**a3c37fb0 — scenarios proper**

scenario 1 (OTEL_EXPORTER_OTLP_HEADERS): new `SdkForwarderState` struct wraps the tonic channel + the parsed `Vec<(MetadataKey<Ascii>, MetadataValue<Ascii>)>` pairs. `parse_otlp_headers_env()` reads `OTEL_EXPORTER_OTLP_TRACES_HEADERS` first (signal-specific precedence per the OTel SDK spec), falls back to `OTEL_EXPORTER_OTLP_HEADERS`; comma-split + `=`-split + warn-and-skip on invalid pairs (matches `opentelemetry-otlp`'s tolerance). forwarder dispatch attaches every parsed pair to `tonic::Request::metadata_mut()` before `TraceServiceClient::export`. test: `engine/tests/otel_forwarder_headers_test.rs` sets `authorization=Bearer-test-token-12345,x-tenant=acme` before init, asserts both metadata entries land on the captured request.

scenario 2 (one bad span doesn't poison the batch): `otlp_span_to_proto` is now fallible — returns `None` for spans whose `trace_id`/`span_id` fail to hex-decode OR decode to non-spec byte lengths (16 / 8). `otlp_scope_spans_to_proto` `filter_map`s them out. caught one gap mid-review: my initial filter was hex-only, so `traceId: "01"` would have decoded to a 1-byte trace_id and slipped through. spec-length check (`OTLP_TRACE_ID_BYTES = 16` / `OTLP_SPAN_ID_BYTES = 8`) added — test now covers all three failure modes (non-hex, hex-but-wrong-length, valid) in one mixed batch.

**380d6841 — test maintainability nits from local bridgebuilder pass**

extracted the duplicated mock-collector helpers into `engine/tests/common/forwarder_mock.rs` (`spawn_body_capturing_collector` + `spawn_metadata_capturing_collector`, plus the shared readiness probe). net tests/ shrinks by ~50 lines and keeps the three forwarder regression tests in lockstep when the forwarder shape changes in the future.

cross-model adversarial dissent (gpt-5.3-codex) was clean against the final state.

sweep: 1542 lib tests + 3 forwarder integration tests pass, cargo fmt + clippy clean on touched files. ready when you are.

ytallo:

Please reply with:

PR #1620, opened 2026-05-11
fix(cli): tighten iii project init UX

guibeira:
## Summary

Four user-visible fixes to `iii project init`:

- **Drop `-l` short alias and `--yes` flag.** Both were scaffolder-TUI carryovers exposed on every `iii project init`. `--languages` keeps its long form; the template flow always runs interactively now (`yes: false` hard-coded).
- **Scope engine flags to top-level only.** Removed `global = true` from `--config`, `--version`, `--use-default-config`, `--no-update-check`, and the three hidden `--install-*` flags so they stop polluting every subcommand's `--help` output. They still parse before a subcommand (`iii --config foo.yaml worker add x` works).
- **Refuse re-init on already-initialized dirs.** A pre-existing `.iii/project.ini` used to silently trigger idempotent re-init, which could clobber user edits to scaffold files. Now it's a hard error directing the user to delete the marker or pick a different `--directory`. `iii project generate-docker` remains the path for adding Docker assets to an existing project.
- **Bake `device_id` into the generated Dockerfile.** The fetched Dockerfile ships with a `__III_DEVICE_ID__` placeholder that the engine substitutes with the real device_id before writing. `.env` drops `III_HOST_USER_ID` and `docker-compose.yml` drops its `args:` block + env propagation. The image's identity no longer depends on whatever sits in `.env` at build time.

## Coordination

**Templates repo (`iii-hq/templates`) must land matching changes before this ships.** `apply_docker` calls `substitute_device_id`, which bails with a clear engine/template drift error if the fetched Dockerfile lacks `__III_DEVICE_ID__`. Update the canonical Dockerfile + `docker-compose.yml` + `template.yaml` comment to match the fixtures under `engine/tests/fixtures/templates/docker/`.

## Test plan

- [x] `cargo test -p iii --test project_init_e2e` — 15/15 pass (5 new tests across the four fixes)
- [x] `cargo test -p iii --bin iii` — 103/103 pass (2 new unit tests for top-level flag scoping)
- [x] `cargo build -p iii` — clean
- [x] `cargo clippy -p iii` — no new warnings introduced
- [x] `iii project --help` — Options block contains only `-h, --help`
- [x] `iii project init --help` — no `-y/--yes`, no `-l` short on `--languages`, `--allow-non-empty` help text reflects new reject-on-existing-project.ini behavior
- [ ] Verify against canonical `iii-hq/templates` repo once the matching PR there lands (currently exercised only against the local fixture under `engine/tests/fixtures/templates/docker/`)
- [ ] Manual smoke: fresh `iii project init demo --docker`, inspect generated `Dockerfile`/`docker-compose.yml`/`.env`, then `docker compose build` to confirm the baked `ENV III_HOST_USER_ID=<uuid>` actually sticks

## Summary by CodeRabbit

* **New Features**
  * Enhanced append operation to support nested paths alongside root-level appends with improved error handling for type mismatches.

* **Documentation**
  * Reorganized architecture documentation; consolidated content into feature-focused guides. Updated internal links for consistency.

* **Chores**
  * Removed interactive confirmation prompt from project initialization; updated language flag behavior for consistency.
  * Refreshed website metadata and SEO improvements; updated website sitemap reference.


guibeira:
Reopening with correct base from the start.

PR #1632, opened 2026-05-13
fix(http): stream JSON raw request bodies

### LINKED ISSUE #1631: Reading raw request body in an HTTP-triggered function
I couldn't find a way to read the raw request body inside a function triggered by an HTTP trigger via the `iii-http` worker. The SDK exposes `ApiRequest.request_body: ChannelReader` but that only seems to work for some content types. For the `application/json` case e.g., the channel is opened but the worker never writes bytes to it when it has already parsed the body itself.

Is something like this already supported? If not, would it be possible to make the HTTP worker pipe raw bytes into `request_body` regardless of content-type while still populating body when it's parseable. Maybe via a per-trigger config like `raw_body: true`? Also, when `request_body` has no bytes, `readAll()` should resolve immediately.

rohitg00:
## Summary
- Write buffered JSON request bytes into the `request_body` channel before invoking the handler.
- Close the request body sender for empty JSON bodies so `readAll` / `read_all` resolves.
- Add HTTP handler regressions for raw JSON and empty JSON request body channels.

## Tests
- `cargo fmt --package iii`
- `cargo test -p iii workers::rest_api::views::tests::test_dynamic_handler -- --nocapture`

Closes #1631

## Summary by CodeRabbit

* **Bug Fixes**
  * JSON request bodies are streamed to handlers' raw request channels and channels close correctly for empty JSON bodies.

* **Tests**
  * Added end-to-end tests (Node, Python, Rust) verifying handlers receive both parsed JSON and the original raw request bytes, including empty bodies.

* **Chores**
  * Telemetry download request now times out after 2 seconds and emits a clear timeout message.

[Review Change Stack](https://app.coderabbit.ai/change-stack/iii-hq/iii/pull/1632)

### DIFF
--- a/engine/src/workers/rest_api/views.rs
+++ b/engine/src/workers/rest_api/views.rs
@@ -403,7 +403,6 @@ pub async fn dynamic_handler(
                     if !body_bytes.is_empty() {
                         let _ = req_tx.send(ChannelItem::Binary(body_bytes.clone())).await;
                     }
-                    drop(req_tx);
                 }
                 serde_json::from_slice(&body_bytes).unwrap_or(Value::Null)
             } else if let Some(req_tx) = req_tx {
--- a/engine/src/workers/rest_api/views.rs
+++ b/engine/src/workers/rest_api/views.rs
@@ -399,6 +399,12 @@ pub async fn dynamic_handler(
                     }
                     axum::body::Bytes::from(buf)
                 };
+                if let Some(req_tx) = req_tx {
+                    if !body_bytes.is_empty() {
+                        let _ = req_tx.send(ChannelItem::Binary(body_bytes.clone())).await;
+                    }
+                    drop(req_tx);
+                }
                 serde_json::from_slice(&body_bytes).unwrap_or(Value::Null)
             } else if let Some(req_tx) = req_tx {
                 let mut body = body;
@@ -1477,6 +1483,33 @@ mod tests {
             .unwrap();
     }
 
+    async fn drain_request_body_channel(
+        engine: &Engine,
+        request: &HttpRequest,
+        timeout_message: &'static str,
+    ) -> Vec<u8> {
+        let mut rx = engine
+            .channel_manager
+            .take_receiver(
+                &request.request_body.channel_id,
+                &request.request_body.access_key,
+            )
+            .await
+            .expect("request body receiver");
+
+        tokio::time::timeout(std::time::Duration::from_secs(1), async {
+            let mut raw = Vec::new();
+            while let Some(item) = rx.recv().await {
+                if let ChannelItem::Binary(bytes) = item {
+                    raw.extend_from_slice(&bytes);
+                }
+            }
+            raw
+        })
+        .await
+        .expect(timeout_message)
+    }
+
     #[tokio::test]
     async fn test_dynamic_handler_route_not_found() {
         let engine = make_test_engine();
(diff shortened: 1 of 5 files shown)


engine/src/workers/rest_api/views.rs
```diff
@@ -399,6 +399,12 @@ pub async fn dynamic_handler(
                     }
                     axum::body::Bytes::from(buf)
                 };
+                if let Some(req_tx) = req_tx {
+                    if !body_bytes.is_empty() {
+                        let _ = req_tx.send(ChannelItem::Binary(body_bytes.clone())).await;
+                    }
+                    drop(req_tx);
```
sergiofilhowz: there's no need to drop, it will be dropped anyway in the next line

sergiofilhowz:
we need to have tests in sdk to ensure this is working, currently most of our test suites are in the sdk side, make sure we write in python, rust and node

rohitg00:
Addressed the review feedback in `c3c9f100e`:

- Removed the explicit `req_tx` drop in the JSON branch.
- Added SDK-side raw JSON `request_body` coverage for Node, Python, and Rust.

Local checks run:

```bash
cargo test -p iii workers::rest_api::views::tests::test_dynamic_handler -- --nocapture
cargo test -p iii-sdk --test api_triggers raw_json_request_body -- --nocapture
uv run --extra dev pytest -q tests/test_api_triggers.py -k raw_json_request_body
pnpm --dir sdk/packages/node/iii exec vitest run tests/api-triggers.test.ts -t "raw JSON request body"
pnpm --filter iii-sdk exec tsc --noEmit
uv run --extra dev ruff check tests/test_api_triggers.py
cargo clippy -p iii-sdk --test api_triggers --all-features -- -D warnings
git diff --check
```

rohitg00:
Addressed the latest inline comments in `ae2b9c18c`:

- Node raw JSON test now registers `test::api::json::raw` and uses `/test/json/raw` for the HTTP trigger path.
- Python raw JSON test now registers a matching `test::api::json::raw::py` function ID and uses `/test/py/json/raw`.
- Rust raw JSON test now uses `/test/rs/json/raw`.

All three findings were still valid; none were skipped.

Validated locally:

```bash
cargo fmt --all
uv run --extra dev ruff format tests/test_api_triggers.py
pnpm --filter iii-sdk exec tsc --noEmit
uv run --extra dev ruff check tests/test_api_triggers.py
cargo clippy -p iii-sdk --test api_triggers --all-features -- -D warnings
git diff --check
III_URL=ws://localhost:49199 III_HTTP_URL=http://localhost:3199 cargo test -p iii-sdk --test api_triggers raw_json_request_body -- --nocapture
III_URL=ws://localhost:49199 III_HTTP_URL=http://localhost:3199 uv run --frozen --extra dev pytest -q tests/test_api_triggers.py -k raw_json_request_body
III_URL=ws://localhost:49199 III_HTTP_URL=http://localhost:3199 pnpm --dir sdk/packages/node/iii exec vitest run tests/api-triggers.test.ts -t "raw JSON request body"
```

rohitg00:
Fixed the failed `Worker Tests (VM) - linux` check in `15b2149ea`.

Root cause: `iii worker add iii-http --wait` correctly skipped worker readiness waiting for built-ins, but still awaited best-effort engine-worker telemetry using the shared registry HTTP client timeout. On Linux CI that outbound telemetry request took ~15s and tripped `handle_managed_add_wait_on_builtin_returns_without_hanging`.

Fix: bound only the built-in telemetry request with a 2s timeout. Real registry/download requests still use the existing registry client behavior.

Validated locally:

```bash
cargo fmt --all
cargo test -p iii-worker --features integration-vm handle_managed_add_wait_on_builtin_returns_without_hanging --test config_managed_integration -- --nocapture
cargo test -p iii-worker --features integration-vm --test config_managed_integration -- --nocapture
cargo test -p iii-worker fire_engine_telemetry_percent_encodes_version_query -- --nocapture
git diff --check
```

Note: I could not run the exact `cargo nextest` wrapper locally because `cargo-nextest` is not installed in this environment; the equivalent affected `cargo test` target passes.

sergiofilhowz (approved):

PR #1635, opened 2026-05-13
feat(engine): expose generated HTTP functions as worker groups

rohitg00:
## Summary

- track generated worker groups for externally registered functions without starting generated worker processes
- expose converted OpenAPI/MCP sources as normal engine-runtime worker groups in `engine::workers::list`
- keep generated functions directly triggerable through normal iii function ids
- consume and strip the private `metadata.iii.generatedWorker` hint before public function metadata is stored
- accept the older private `metadata.iii.virtualWorker` key as a compatibility alias
- trust local generated-function bridges only when both the registering worker session and invocation URL are loopback
- enable `iii-http-functions` by default so converted OpenAPI/MCP calls can register without extra config
- cover the generated HTTP worker-group path from the Node, Python, and Rust SDK test suites

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p iii --test generated_worker_bridge_e2e`
- `cargo test -p iii generated_worker --lib`
- `cargo test -p iii test_list_worker_infos_includes_generated_group_without_private_public_shape --lib`
- `cargo test -p iii http_functions --lib`
- `cargo check -p iii --tests`
- `cargo test -p iii --test http_e2e_security`
- `cargo test -p iii --test http_e2e_error_handling`
- `cargo fmt -p iii-sdk -- --check`
- `pnpm install --filter iii-sdk --frozen-lockfile`
- `pnpm --filter iii-sdk exec tsc --noEmit`
- `python3 -m compileall -q sdk/packages/python/iii/src sdk/packages/python/iii/tests/test_http_external_functions_integration.py`
- `cargo test -p iii-sdk --test http_external_functions exposes_generated_http_functions --no-run`
- spec-to-worker repo: `cargo fmt`, `cargo test`, `cargo clippy --all-targets --all-features -- -D warnings`

## Spec-to-Worker E2E Contract

The paired worker PR is iii-experimental/spec-to-worker#13. It now uses generic fixtures and docs instead of Context7-specific examples. The expected contract is:

- OpenAPI conversion registers normal iii functions for each operation
- MCP HTTP conversion registers normal iii functions for each discovered tool
- MCP stdio conversion registers normal iii functions for each discovered tool
- `engine::workers::list` shows the generated source as a normal engine-runtime worker group
- public worker/function payloads do not expose private routing metadata or generated process details
- no generated worker process is started; routing remains engine-owned through HTTP invocation

## Notes

`spec-to-worker::convert` registers normal iii functions backed by HTTP invocation. The engine exposes a user-facing worker grouping for discovery, while routing remains engine-owned and no backing worker process is started for that group. Other workers see ordinary worker/function shapes, not private generated routing details.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2544,6 +2544,7 @@ dependencies = [
  "tracing",
  "tracing-opentelemetry",
  "tracing-subscriber",
+ "url",
  "uuid",
  "which",
  "winapi",
--- a/engine/Cargo.toml
+++ b/engine/Cargo.toml
@@ -120,6 +120,7 @@ uuid = { version = "1", features = ["v4", "serde"] }
 dashmap = "6"
 indexmap = { version = "2", features = ["serde"] }
 redis = { version = "1.0.1", features = ["tokio-comp", "connection-manager"] }
+url = { workspace = true }
 lapin = { version = "3", optional = true }
 reqwest = { version = "0.12", default-features = false, features = ["json", "rustls-tls", "stream"] }
 async-trait = "0.1.89"
--- a/engine/src/engine/mod.rs
+++ b/engine/src/engine/mod.rs
@@ -4,7 +4,10 @@
 // This software is patent protected. We welcome discussions - reach out at team@iii.dev
 // See LICENSE and PATENTS files for details.
 
-use std::{net::SocketAddr, sync::Arc};
+use std::{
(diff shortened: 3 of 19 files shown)


rohitg00:
Superseded by #1640 after renaming the head branch to generated-http-functions and adding the worker CI cache fix.

PR #1665, opened 2026-05-20
docs(stream): add agent skills bundle for iii-stream worker

rohitg00:
## Summary

First per-worker **skills bundle** for the engine's built-in workers, mirroring the layout used in [`iii-hq/workers`](https://github.com/iii-hq/workers) (e.g. [`iii-database/skills/...`](https://github.com/iii-hq/workers/tree/main/iii-database/skills)) and following the contract in [`DOCUMENTATION_GUIDELINES.md`](https://github.com/iii-hq/workers/blob/main/DOCUMENTATION_GUIDELINES.md).

This is **PR 1 of an N-PR plan** to add skills bundles to each engine built-in worker. `iii-state` is being authored separately by Sergio to test the pipeline; the remaining workers (`cron`, `rest_api`, `bridge_client`, `observability`, `shell`, `pubsub`, `http_functions`, ...) follow as independent PRs using the same template.

### What's added

```
engine/src/workers/stream/skills/
├── index.md                  # type: index — worker overview, surfaced in agent bootstraps
└── stream/
    ├── set.md                # how-to + Side effects (persist + WS broadcast + trigger fan-out)
    ├── get.md                # how-to (pure read by triple)
    ├── delete.md             # how-to + Side effects (no-op when item absent)
    ├── update.md             # how-to + Side effects (atomic op-list, UpdateOpError shape)
    ├── send.md               # how-to + Side effects (transient broadcast, no persist)
    └── listing.md            # multi-function file: list + list_groups + list_all
```

## Test plan

- [ ] Skills directory structure validates against `DOCUMENTATION_GUIDELINES.md` (frontmatter, required sections in order, optional sections placed before `# Worked example`, `iii://` link shape, no emoji). Validated locally with a small linter; all seven files pass.
- [ ] Frontmatter `function_id` values match the real ids registered in `engine/src/workers/stream/stream.rs` (`stream::set`, `stream::get`, `stream::delete`, `stream::list`, `stream::list_groups`, `stream::list_all`, `stream::send`, `stream::update`).
- [ ] JSON examples parse as valid JSON shape (placeholder strings, real keys, plausible values, inline `// ...` comments per guidelines).
- [ ] Cross-links between siblings (`set` <-> `update` <-> `delete` <-> `send`, all four <-> `get`, all five <-> `listing`) all resolve to files that exist in this PR.
- [ ] CI on this PR is green (no Rust changes, so the engine/SDK test matrix should skip; only docs/path filters apply).

## Summary by CodeRabbit

* **Documentation**
  * Added comprehensive docs for stream operations (set, get, update, delete, send) including behaviors, side‑effects, worked examples, and error/response shapes.
  * Added discovery docs for listing functions (list, list_groups, list_all) with ordering and metadata guarantees.
  * Documented reactive triggers (stream, stream:join, stream:leave), per‑connection auth, subscription lifecycle, broadcast semantics, adapter constraints, and configuration defaults.

[Review Change Stack](https://app.coderabbit.ai/change-stack/iii-hq/iii/pull/1665?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


rohitg00:
Pushed a self-audit fix-up after re-reading the source. Five concrete things changed; one assumption is still open and called out below.

## What changed

**1. Side-effects ordering corrected in `set.md`, `update.md`, `delete.md`, `send.md`.**

`engine/src/workers/stream/stream.rs:485-489` does:
```rust
self.invoke_triggers(message.clone()).await;        // spawns a task, doesn't await handlers
if let Err(e) = adapter.emit_event(message).await { … }
```

And `invoke_triggers` itself is `tokio::spawn(async move { … })` (`stream.rs:338`). So the actual runtime ordering for every write is:

1. **Synchronous** — adapter persists the new value (or removes the entry, for `delete`).
2. **Asynchronous, fire-and-forget** — trigger task is spawned. Handlers run on the spawned task **after** the originating call returns. Slow or failing handlers do not delay or fail the originating write.
3. **Synchronous** — `adapter.emit_event` broadcasts the `StreamWrapperMessage` to WebSocket subscribers, in the caller's task, before the call returns.

The original docs had this both numbered wrong (1=persist, 2=broadcast, 3=triggers) **and** prose-contradicted (saying "triggers fire before broadcast completes"). Both are gone; the new prose says "the trigger task is the only thing that can still be running when the call returns."

**2. Reactive triggers documented as first-class skills.**

The original bundle treated `stream` / `stream:join` / `stream:leave` as a one-sentence footnote in `index.md`. Reactive triggers are how you build reactive backends in iii (`skills/iii-reactive-backend`, `skills/iii-state-reactions`, `skills/iii-realtime-streams`), so this PR now ships per-trigger how-tos:

- `triggers/stream.md` — `stream_name` is required (the worker indexes triggers by it and unwraps in `register_trigger`); `group_id` and `item_id` use empty-string-as-wildcard semantics; full `StreamWrapperMessage` payload shape with the four `event.type` discriminants (`create`/`update`/`delete`/`event`).
- `triggers/stream-join.md` — documents the **authorization-gate semantics** the original bundle missed entirely. Returning `{ unauthorized: true }` from a join handler rejects the subscription before any data flows (`engine/src/workers/stream/connection.rs:135-152`). Pairs with the worker config's `auth_function`, whose return value lands in every join/leave event's `context` field.
- `triggers/stream-leave.md` — paired cleanup hook; `subscription_id` joins to the matching `stream:join` event; return value is ignored (no auth gate here).

Each trigger how-to follows the same five-section shape as the function how-tos (`# When to use` → `# Inputs` → `# Outputs` → `# Worked example` → `# Related`) with the semantic mapping that "Inputs = trigger config", "Outputs = handler event payload". Frontmatter uses a new `trigger_type:` field in place of `function_id:` to identify the trigger; `type: how-to` is preserved so existing tooling indexes them as how-tos.

**3. `UpdateOp` paths documented accurately.**

The original `update.md` said `path` was a string for every op. Source (`sdk/packages/rust/iii/src/types.rs:23-107`) actually distinguishes:

- `FieldPath` = single string (used by `set`, `increment`, `decrement`, `remove`).
- `MergePath` = enum of `Single(String)` or `Segments(Vec<String>)` or omitted (used by `merge` and `append` only).

The op section now spells out the difference and includes the engine's path normalization rules (omitted / `""` / `[]` → root; `"foo"` ≡ `["foo"]`; `["a.b"]` is one literal key, not nested `a → b`).

**4. `index.md` restructured to give triggers equal billing.**

The opening paragraph now names both surfaces (functions and reactive triggers) explicitly. The `## How-tos` section has two `###` sub-categories — `### stream::* functions` and `### Reactive triggers` — each with bullet links to the per-skill files. This is a minor extension to `DOCUMENTATION_GUIDELINES.md` section 2 (which prescribes only `### <ns>::*` sub-namespace headings under `## How-tos`). I kept everything under `## How-tos` rather than splitting into separate top-level sections because that's the heading the directory reader's bootstrap renderer keys off — splitting would risk dropping triggers from agent bootstrap output.

**5. Cross-link integrity verified.**

A small validator confirms every `iii://stream/<rest>` link resolves to a file in the bundle under both interpretations of the deployment layout (collapsed primary-namespace and uncollapsed). 10/10 files pass.

## Still-open assumption

**The engine has no `skills_folder` plumbing for built-in workers yet.**

`iii-directory`'s `fs_source.rs` resolves `iii://<ns>/<rel>` to `<skills_folder>/<ns>/<rel>.md`. The `<ns>` segment in the iii-directory precedent is the **function namespace** (`directory`), not the worker name (`iii-directory`); the deployment process strips the inner duplicate `iii-directory/` directory so source `iii-directory/skills/directory/skills/list.md` ends up at URI `iii://directory/skills/list`.

This bundle's URIs follow that same convention: `iii://stream/set` for the function `stream::set`, `iii://stream/triggers/stream` for the reactive trigger. For these to resolve, the engine needs to mount `engine/src/workers/stream/skills/` at `<skills_folder>/stream/` and apply the same inner-namespace collapse.

That plumbing doesn't exist in `engine/` today. Two ways forward:

- (A) The engine adds a `skills_folder` configuration that walks built-in workers' `skills/` dirs at startup. Until then, these URIs are aspirational.
- (B) Adjust the URIs in this bundle to whatever scheme the engine ultimately uses (e.g. `iii://iii-stream/stream/set` if the namespace is the worker name from `iii.worker.yaml`, or `iii://stream/stream/set` if the inner namespace is preserved).

Happy to flip to either as soon as the resolution rule is decided.

engine/src/workers/stream/skills/triggers/stream.md
```diff
@@ -0,0 +1,187 @@
...
+Watch the `presence` stream for any change in `room-1` and project a
+running attendee count into a `metrics` stream:
+
+```typescript
+import { registerWorker, TriggerAction } from 'iii-sdk'
+
+const iii = registerWorker('ws://localhost:49134')
+
+const onChange = iii.registerFunction(
+  { id: 'presence::onChange' },
```
sergiofilhowz:
```suggestion
  id: 'presence::onChange',
```

engine/src/workers/stream/skills/skills/stream/reactive-triggers.md
```diff
@@ -0,0 +1,172 @@
...
+- **Drive a derived projection.** From inside a `stream` handler, call `stream::set` / `stream::update` against a different `(stream_name, group_id)` to maintain counters, search indexes, or audit logs; the originating write has already returned, so a slow projection cannot delay or block writers.
+
+For runnable scaffolds covering these patterns end-to-end (TypeScript, Python, and Rust), see the stream worker source and the SDK usage examples in [the iii main repo](https://github.com/iii-hq/iii).
+
+# Related
+
+- [`stream::set`](iii://iii-stream/stream/set), [`stream::update`](iii://iii-stream/stream/update), [`stream::delete`](iii://iii-stream/stream/delete), [`stream::send`](iii://iii-stream/stream/send) — the four functions whose calls the `stream` trigger reacts to.
+- [`stream::get`](iii://iii-stream/stream/get) — read on demand instead of registering a trigger.
+- `auth_function` (worker config; see [the README](../../../README.md)) — the once-per-connection hook whose return value lands in every `stream:join` / `stream:leave` event's `context` field.
```
sergiofilhowz: make sure to document the this in a skill file, not in readme

engine/src/workers/stream/skills/index.md
```diff
@@ -0,0 +1,51 @@
...
+
+The `stream:join` trigger doubles as an **authorization gate**:
+returning `{ unauthorized: true }` from a join handler rejects the
+incoming subscription before any data flows. Pair this with the
+worker's `auth_function` config (which runs once per WebSocket
+handshake and stamps a `context` value into every join/leave event) to
+build per-connection authorization. For the `port`/`host`/`adapter`/
+`auth_function` config block, the WebSocket URL shape, and the
+per-adapter compatibility table (`kv` is single-instance, `redis` fans
+real-time delivery across instances), see [the README](../README.md).
```
sergiofilhowz: make sure to document the this in a skill file, not in readme

sergiofilhowz (dismissed):

sergiofilhowz (approved):

PR #1742, opened 2026-06-03
fix(iii-worker): replace existing bundle install on cross-project add

andersonleal:
## Problem

`iii worker add <bundle>` fails when the same bundle worker is already installed in another project on the machine:

```
✓ Downloaded bundle harness (0.0s)
✓ Extracted bundle (0.0s)
error: worker "harness" already exists
error: [W900] internal: iii worker add exited with rc 1
```

Once any project installs a bundle worker, no other project on the same machine can add it.

## Root cause

Bundle workers install to a **global, machine-wide** directory keyed by name only — `~/.iii/workers-bundle/{name}/` (`config_file.rs:bundle_worker_path`) — shared across every project. `handle_bundle_add` always ran the full pipeline and called `atomic_install`, which hard-failed if the target dir exists (W111 `AlreadyExists`). So a second project's add aborted at the install step, **before** the worker was ever registered in that project's `config.yaml`. The "already exists" was about the shared on-disk bundle, not a duplicate in your project.

## Fix

Every add now runs the full pipeline and **replaces** the on-disk bundle with the freshly resolved version (rather than reusing whatever the first project installed, which would pin every project to a possibly-stale payload):

- `atomic_install` replaces instead of refusing: an existing install is parked aside as a unique `{name}.old.<unique>` sibling (same-fs atomic rename), the new tree is installed (`install_fresh`: rename, with the existing `.partial` cross-fs fallback), and the parked dir is removed. If the new install fails, the previous install is **restored** — it's never lost. Stale `.old.*` siblings from a replacement that crashed mid-swap are swept on the next install. The whole swap runs under the existing per-worker staging fslock.
- `config_file::bundle_is_installed(name)` — manifest-gated detection of the global install, reused by `check_install_fallback` and to decide rollback behavior.
- On `append_worker` failure, only **fresh** installs are rolled back. A replaced install is kept: the previous payload is already gone and other projects resolve the bundle by name from disk — deleting it would break all of them, keeping the new one breaks nobody.
- Side effect: `iii worker update` funnels bundles through `handle_bundle_add`, so it now actually refreshes the on-disk install (previously it was a no-op once any install existed).

This covers both direct adds and transitive dependency installs — both route through `handle_bundle_add`, the sole caller of `atomic_install`.

## Known limitation

The global cache is keyed by name with no version, so two projects can't independently pin different bundle versions — the last add wins machine-wide (always the freshly resolved version, so it can't go stale). True per-project version pinning is a separate, larger change (version-keyed install dirs touching resolver/list/start/update).

## Test plan

- [x] `atomic_install_replaces_existing_target` — existing install fully replaced, no `.old.*`/`.partial.*` leftovers
- [x] `atomic_install_restores_previous_install_on_failure` — failed install restores the previous payload
- [x] `atomic_install_sweeps_stale_old_siblings` — crash-leaked `.old.*` dirs swept on next install
- [x] `atomic_install_concurrent_both_safe` — unserialized racing installs end with a complete install, no leaks
- [x] `bundle_add_replaces_existing_install_in_second_project` — full pipeline offline (sha256 cache pre-seeded): second project's add replaces the stale payload and registers a name-only `config.yaml` entry
- [x] `bundle_is_installed_detects_shared_global_install`
- [x] `cargo test -p iii-worker` — 49 integration + 793 lib tests pass; clippy clean for changed files

## Summary by CodeRabbit

* **New Features**
  * Faster global bundle reuse: installed bundles are detected by their manifest and can be reused across projects.
  * Crash-safe replace installs: replacements use a safe rename-and-sweep flow to avoid corruption.

* **Bug Fixes**
  * Improved atomic install and rollback: prevents leaked .old/.partial siblings, handles cross-filesystem moves, and ensures concurrent installs remain safe.

* **Tests**
  * New regression and integration tests for detection, reuse, replace semantics, atomic installs, and concurrency.

* **Documentation**
  * Updated install docs describing replacement and cleanup behavior.

### DIFF
--- a/crates/iii-worker/src/cli/bundle_download.rs
+++ b/crates/iii-worker/src/cli/bundle_download.rs
@@ -33,12 +33,17 @@
 //!   slipping into a partially-cleaned-up state.
 //! - The final step is an atomic `std::fs::rename` from the staging dir
 //!   to `~/.iii/workers-bundle/{name}/`. Cross-filesystem renames fall
-//!   back to copy-then-delete.
+//!   back to copy-then-delete. A previous install of the same name is
+//!   parked aside as a `{name}.old.<unique>` sibling first and restored
+//!   if the new install fails (see `atomic_install`).
 //!
 //! Process-restart safety:
 //! - The `Drop` guard cannot run if the process is killed mid-install.
 //! - `sweep_orphans()` runs at engine/CLI startup and removes any
 //!   leftover staging directories. See T18.
+//! - `{name}.old.*` siblings leaked by a replacement that crashed
+//!   between rename-aside and cleanup are swept on the next install of
+//!   the same worker (`sweep_stale_old_siblings`).
 
 use std::path::{Path, PathBuf};
 use std::sync::LazyLock;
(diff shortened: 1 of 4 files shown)


sergiofilhowz (commented):
it should override when using a clean iii instance (when iii.lock is not present) and use the latest or the selected version (ex `iii worker add harness@0.1.0`)

sergiofilhowz (dismissed):

PR #1750, opened 2026-06-03
fix: issue with unregister trigger in custom triggers

sergiofilhowz:
## Summary by CodeRabbit

* **New Features**
  * SDKs now receive and handle trigger unregistration notifications so providers observe when consumer bindings are removed and cleanup can run; provider reconnect still re-registers triggers.

* **Tests**
  * Added cross-language integration suites validating trigger lifecycle: firing, provider reconnect, and consumer disconnect behavior.

* **Documentation**
  * Added/versioned many docs and a Quickstart to the 0.17.0 site.

* **Changelog**
  * Documented inbound unregister-trigger behavior in the 0.19.0 changelog.


sdk/packages/rust/iii/src/iii.rs
guibeira, line 1460: Will be great to include a logger here, this way we can check what kind message we are not handling.

docs/0-17-0/index.mdx.skill.md
```diff
@@ -0,0 +1,96 @@
+<!-- generated by iii-skill-render. DO NOT EDIT (changes here are overwritten on the next render). Edit docs/0-17-0/index.mdx. -->
```
guibeira, line 1: We need to check if this file is necessary.

docs/0-17-0/install.mdx.skill.md
```diff
@@ -0,0 +1,44 @@
+<!-- generated by iii-skill-render. DO NOT EDIT (changes here are overwritten on the next render). Edit docs/0-17-0/install.mdx. -->
```
guibeira, line 1: We need to check if this file is necessary.

docs/0-17-0/quickstart.mdx.skill.md
```diff
@@ -0,0 +1,254 @@
+<!-- generated by iii-skill-render. DO NOT EDIT (changes here are overwritten on the next render). Edit docs/0-17-0/quickstart.mdx. -->
```
guibeira, line 1: We need to check if this file is necessary.

guibeira (approved):
Looks like we are missing handling this message type; it will be great to include a logger for any unexpected type of message the SDK receives.

About the docs, there are some new files at the archives 0.17.0, we need to check if they are valid or not.

Overall looks good.

PR #1771, opened 2026-06-04
refactor(rust-sdk)!: remove orphan TriggerTypeInfo

guibeira:
Removes the `TriggerTypeInfo` struct from the Rust SDK. It is an orphan: no producer or consumer in any SDK, engine, or example — its endpoint `engine::trigger-types::list` was removed (trigger types are now served by `engine::triggers::list` -> `TriggerInfo`). Deletes the struct and its crate-root re-export. Hard delete, no shim (breaking: `iii_sdk::TriggerTypeInfo` no longer exported).

## Summary by CodeRabbit

* **Breaking Changes**
  * Removed a public API component from the Rust SDK. Code that depends on this component for response data will require updates.

### DIFF
--- a/sdk/packages/rust/iii/src/iii.rs
+++ b/sdk/packages/rust/iii/src/iii.rs
@@ -86,17 +86,6 @@ pub struct TriggerInfo {
     pub metadata: Option<Value>,
 }
 
-/// Trigger type information returned by `engine::trigger-types::list`
-#[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct TriggerTypeInfo {
-    pub id: String,
-    pub description: String,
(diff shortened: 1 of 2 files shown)


sdk/packages/rust/iii/src/iii.rs
sergiofilhowz, line 104: i know this is out of scope of this PR, but where do we store the configuration_format?

sergiofilhowz (approved):

sdk/packages/rust/iii/src/iii.rs
guibeira: Sorry, what do you mean with `configuration_format`?

PR #1780, opened 2026-06-05
feat(worker-api)!: self-describing worker::* API + builtin worker descriptions

andersonleal:
## Summary

Makes the entire worker surface discoverable by LLM and automation callers: the `worker::*` management ops self-describe with real schemas, and every builtin worker now carries a description surfaced in `engine::workers::list` / `engine::workers::info`.

### Worker-op introspection (`2230f52c2`)
- All 8 `worker::*` registrations attach typed request schemas, descriptions, and metadata (`default_timeout_ms`, `idempotent`) via a shared `describe_op` helper; `op_description` is the single source of truth shared with `worker::schema`. Schema table built once via `LazyLock`.
- `iii trigger <fn> --help` fetches `engine::functions::info` and renders a parameter table; degrades gracefully against older engines.
- Workers self-describe: `description` added to the Rust/Node/Go/Python SDK `WorkerMetadata`, capped and sanitized at the engine ingest boundary.
- New `WorkerOpError::BadRequest` (W105) for malformed payloads with a `details.hint` naming `worker::schema`; W101 retained as a reserved wire code. Worker-op failures map to `IIIError::Remote` so `ErrorBody.code` carries the stable W-code.
- `workers::info` trigger-type attribution resolves by owner name, fixing trigger types vanishing when several connections share one worker name; DashMap re-entrancy deadlock avoided in owner resolution.
- Console DLQ tab unwraps the W-code JSON envelope and caps message paths against DOM flooding.
- `--help` sanitizes all worker-controlled fields (incl. Unicode bidi / Trojan-Source overrides); worker-reported names sanitized at ingest.

### Builtin worker descriptions (`94349acac`)
- `WorkerRegistration` gains a required `description: &'static str`; the `register_worker!` macro mandates `description = "..."` in all three variants, so future builtins cannot register without one.
- All 14 in-process builtins described (cron, queue, state, stream, pubsub, http, http-functions, observability, telemetry, configuration, exec, bridge, worker-manager, engine-functions), reusing `iii.worker.yaml` text where it existed; `configuration` and `queue` manifests synced.
- `RuntimeWorkerInfo` carries the description (inventory lookup), replacing the hardcoded `description: None` in both `workers::list` and `workers::info`. Programmatically registered workers stay `None`.
- SDK-connected first-party daemons `iii-sandbox` and `iii-console` now set `WorkerMetadata.description` (`iii-worker-ops` already did).

## Breaking changes

- `worker::add { source: { kind: "local", path } }` over the trigger now succeeds (previously rejected with W102). **Security posture change (intentional, explicitly decided):** a local install makes the daemon execute manifest setup/install/start scripts on the engine host with daemon privileges, reachable by any connected worker. Treat the daemon as a host-level code-execution surface when reachable by untrusted callers.
- Malformed `worker::*` payloads return W105 (was W101); worker-op failures surface the W-code as `ErrorBody.code` (was `invocation_failed`).
- `WorkerMetadata.description` is source-breaking for Rust SDK consumers constructing the struct without `..Default::default()`.
- `register_worker!` now requires the `description` argument (compile error otherwise).

## Test plan

- [x] `cargo test -p iii --lib` — 1682 passed
- [x] `cargo test -p iii --test builtin_functions_e2e` — 6 passed (null-description assertion flipped to pin the real text)
- [x] `cargo check -p iii -p iii-worker --all-targets` and `cargo check -p iii-console` clean
- [x] New guard test `every_builtin_worker_registration_has_description` (inventory coverage + lookup)
- [x] `worker_trigger_integration` (W105 envelope, op metadata/description), SDK `error_wire`, help.rs sanitizer coverage, Python description present/absent
- [x] Multi-model pre-landing review on the first commit; 15-agent adversarial review on the second (caught the missing `iii-console` description)

## Summary by CodeRabbit

* **New Features**
  * Workers can publish an optional one-line description shown in listings/info.
  * Local worker installs via triggers are now supported.
  * Worker schemas now include description, default timeout, and idempotency metadata.
  * Help output and function descriptions are sanitized for terminal safety.
  * Improved dead-letter queue error message extraction and truncation.

* **Bug Fixes**
  * Malformed request payloads now return W105 (BadRequest) with actionable details.

* **Documentation**
  * Updated guides and SDK references to document metadata, schema, and trigger behavior.

### DIFF
--- a/crates/iii-worker/src/cli/worker_manager_daemon.rs
+++ b/crates/iii-worker/src/cli/worker_manager_daemon.rs
@@ -18,10 +18,10 @@ use std::sync::{Arc, Mutex};
 
 use crate::core::{
     AddOptions, AddOutcome, ClearOptions, ClearOutcome, EventSink, ListOptions, ListOutcome,
-    NullSink, ProjectCtx, RemoveOptions, RemoveOutcome, StartOptions, StartOutcome, StopOptions,
-    StopOutcome, UpdateOptions, UpdateOutcome, WorkerOpError, add as core_add, clear as core_clear,
-    list as core_list, remove as core_remove, start as core_start, stop as core_stop,
-    update as core_update,
+    LogsOptions, LogsOutcome, NullSink, ProjectCtx, RemoveOptions, RemoveOutcome, StartOptions,
+    StartOutcome, StopOptions, StopOutcome, UpdateOptions, UpdateOutcome, WorkerOpError,
+    WorkerOpErrorKind, add as core_add, clear as core_clear, list as core_list, logs as core_logs,
+    remove as core_remove, start as core_start, stop as core_stop, update as core_update,
 };
 use iii_observability::OtelConfig;
 use iii_sdk::{
@@ -29,7 +29,7 @@ use iii_sdk::{
     register_worker,
 };
 use schemars::{JsonSchema, schema_for};
-use serde_json::Value;
+use serde_json::{Value, json};
 
 use crate::cli::app::WorkerManagerDaemonArgs;
 use crate::cli::host_shim::CliHostShim;
@@ -53,6 +53,12 @@ pub async fn run(args: WorkerManagerDaemonArgs) -> i32 {
             otel: Some(OtelConfig::default()),
             metadata: Some(WorkerMetadata {
                 name: "iii-worker-ops".to_string(),
+                description: Some(
+                    "Manages installed workers: add/remove/update/start/stop/list/clear/logs, \
+                     plus worker::schema introspection and the `worker` lifecycle \
+                     trigger type."
+                        .to_string(),
+                ),
                 ..Default::default()
             }),
             ..Default::default()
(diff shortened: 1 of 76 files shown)


crates/iii-worker/src/cli/worker_manager_daemon.rs
```diff
@@ -94,33 +100,81 @@ pub fn err_payload(e: &WorkerOpError) -> String {
...
+        "worker::stop" => "Stop a running worker",
+        "worker::list" => "List installed workers",
+        "worker::clear" => "Wipe worker artifacts",
+        "worker::schema" => {
+            "Introspect request/response schemas for worker::* triggers. \
+             Optional `function_id` filters to a single trigger."
+        }
+        _ => "",
+    }
+}
```
sergiofilhowz, line 162: why don't we use latest rust sdk features?

crates/iii-worker/src/cli/worker_manager_daemon.rs
```diff
@@ -264,144 +327,152 @@ fn sink_ref<'a>(sink: &'a Arc<IIIEventSink>) -> &'a dyn EventSink {
...
-                    .map_err(|e| bad_request_error("worker::stop", &e))?;
-                let ctx = ProjectCtx::open(project_root).map_err(|e| op_error(&e))?;
-                core_stop::run(opts, &ctx, sink_ref(&sink), &CliHostShim)
-                    .await
-                    .map_err(|e| op_error(&e))
-            }
-        })
-        .description("Stop a running worker"),
+        describe_op::<StopOptions>(
+            RegisterFunction::new_async(move |payload: Value| {
```
sergiofilhowz: we should avoid using serde::Value and define our own types, it creates the json schema out of the box

ytallo (approved):

ytallo (approved):

PR #1855, opened 2026-06-13
fix(observability): stop shutdownOtel() hanging when the engine is unreachable

### LINKED ISSUE #1835: iii-sdk: shutdownOtel() hangs forever when the engine is unreachable (forceFlush never resolves)
**Summary**

`sdk.shutdown()` never resolves if the iii engine has become unreachable by the time shutdown runs. The OTel shutdown path calls `tracerProvider.forceFlush()`, which queues spans onto a WebSocket exporter that is stuck in an **infinite** reconnect loop, so the flush promise never settles. The process hangs until something force-kills it.

**Impact**

Any long-lived process that embeds the SDK and shuts down via SIGTERM. In my case ([agentmemory](https://github.com/rohitg00/agentmemory), which bundles `iii-sdk`), this manifested as a **90-second hang on every OS shutdown/reboot** — systemd waits out its stop timeout and then SIGKILLs the process. On a desktop this looks like the machine freezing on a black screen during shutdown.

**Root cause**

Traced through the published dist of `iii-sdk@0.11.2` and `@iii-dev/observability@0.19.2` (latest). The relevant symbols:

1. `DEFAULT_RECONNECTION_CONFIG.maxRetries = -1` (infinite retries by default).
2. `SharedEngineConnection` therefore **never reaches the `failed` terminal state** — it reconnects forever with exponential backoff.
3. `EngineSpanExporter` drains its pending queue via an `onFailed()` callback that only fires in the `failed` state. With `maxRetries = -1`, that callback is **unreachable**, so in-flight exports never fail out. (The code comment on `onFailed` even says it exists *"so in-flight forceFlush() calls do not hang"* — but the default config disables the only path that triggers it.)
4. `shutdownOtel()` calls `tracerProvider.forceFlush()` / `meterProvider.forceFlush()` / `loggerProvider.forceFlush()` **before** `sharedConnection.shutdown()`. So when the engine is gone, the flush enqueues spans, the exporter waits on a reconnect that never succeeds and never fails, and `sharedConnection.shutdown()` is never reached.

**Evidence** (real shutdown captured via journald; engine had already exited):

```
[agentmemory] Shutting down...
[OTel] Disconnected from engine, will reconnect...
[OTel] Reconnecting in 1200ms (attempt 1)...
[OTel] Reconnecting in 2479ms (attempt 2)...
[OTel] Reconnecting in 3913ms (attempt 3)...
... (continues; only stops when the process is SIGKILLed)
```

In a previous run without an external timeout, the loop ran to attempt 8 (~30s backoff each) and was killed after 90s.

**Suggested fix directions:**

- Make the OTel export/flush bounded (e.g. `exportTimeoutMillis` on the batch processor) so `forceFlush()` fails fast instead of waiting on the queue.
- Default `maxRetries` to a finite value so the existing `failed` → `failPending()` drain becomes reachable.
- In `shutdownOtel()`, tear down / signal `sharedConnection` **before** flushing providers, so pending sends fail and the flush resolves.

I can put together a minimal standalone repro (init OTel → nothing listening on the engine WS → emit a span → call `shutdownOtel()` → it hangs) if that would help

If you want me to open a PR with your preferred approach, let me know.

**Versions:** `iii-sdk@0.11.2` (as bundled in agentmemory 0.9.27) and `@iii-dev/observability@0.19.2` (verified the same ordering + default in latest). Node 26.2.0, Arch Linux (kernel 7.0.10).

CryptoQuaternions:
Fixes #1835. Thanks for confirming, @guibeira — here's the fix.

## Problem

When the engine is unreachable at shutdown, `shutdownOtel()` `forceFlush()`es the OTel providers while `SharedEngineConnection` is still in its (infinite) reconnect loop. Each exporter's `doExport()` sees a non-`connected` state and **queues** the export without invoking its result callback — that callback only fires on reconnect, on the `failed` state, or on exporter shutdown. With the default `maxRetries: -1` the `failed` state is unreachable, the engine never returns, and the queue is only drained *after* the flush — so `forceFlush()` never resolves and the process hangs until it's force-killed (~90s under systemd).

## Fix

An explicit draining phase, no timeouts:

- `SharedEngineConnection.beginShutdown()` stops reconnecting and marks the connection as shutting down, but leaves the open socket in place so buffered telemetry can still flush.
- Exporters **fail-fast** (drop, like the `failed` state) when there's no live connection *and* the connection is shutting down, instead of queueing.
- `shutdownOtel()` calls `beginShutdown()` before the provider flushes.

Net: a final `forceFlush()` still **sends** buffered telemetry when the engine is reachable (healthy shutdown unchanged), but **fails fast** instead of hanging when it isn't.

## Testing

New regression test (`span-exporter.test.ts`): an export queues while disconnected, then fails fast once shutting down. `pnpm test` (18 pass), `type-check`, `build` all green; biome clean on changed files. The hang affected traces, metrics and logs, so all three exporters get the symmetric change.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved graceful shutdown behavior for telemetry exporters to prevent process hangs
  * Enhanced connection state handling during shutdown to properly fail pending telemetry exports with clear error messages
  * Optimized shutdown sequence to stop reconnection attempts and drain queued metrics, logs, and spans

### DIFF
--- a/sdk/packages/node/observability/src/telemetry-system/connection.ts
+++ b/sdk/packages/node/observability/src/telemetry-system/connection.ts
@@ -214,11 +214,32 @@ export class SharedEngineConnection {
    * be flushed. Call before flushing, then call shutdown() to close fully.
    */
   beginShutdown(): void {
+    if (this.shuttingDown) {
+      return
+    }
     this.shuttingDown = true
 
     if (this.reconnectTimeout) {
       clearTimeout(this.reconnectTimeout)
       this.reconnectTimeout = null
+    }
+
+    // With no live connection there's nothing to flush to, so fail queued work
+    // now instead of leaving callbacks unresolved for the flush below to wait on
+    if (this.state !== 'connected') {
+      const pending = this.pendingMessages.splice(0, this.pendingMessages.length)
+      const shutdownError = new Error('Connection shutdown before message could be sent')
+      for (const { callback } of pending) {
+        callback?.(shutdownError)
+      }
+
+      for (const cb of this.onFailedCallbacks) {
+        try {
+          cb()
+        } catch (err) {
+          console.error('[OTel] onFailed callback threw:', err)
+        }
+      }
     }
   }
 
--- a/sdk/packages/node/observability/src/telemetry-system/connection.ts
+++ b/sdk/packages/node/observability/src/telemetry-system/connection.ts
@@ -200,6 +200,29 @@ export class SharedEngineConnection {
   }
 
   /**
+   * Whether the connection is shutting down. While shutting down, exporters
+   * fail-fast instead of queueing when there is no live connection, so a final
+   * forceFlush() can't hang waiting for a reconnect that will never happen.
+   */
+  isShuttingDown(): boolean {
+    return this.shuttingDown
+  }
+
+  /**
+   * Begin shutdown: stop reconnecting and stop accepting new queued exports,
+   * while leaving an open connection in place so buffered telemetry can still
+   * be flushed. Call before flushing, then call shutdown() to close fully.
+   */
+  beginShutdown(): void {
+    this.shuttingDown = true
+
+    if (this.reconnectTimeout) {
+      clearTimeout(this.reconnectTimeout)
+      this.reconnectTimeout = null
+    }
+  }
+
+  /**
    * Shutdown the connection.
    */
   async shutdown(): Promise<void> {
--- a/sdk/packages/node/observability/src/telemetry-system/index.ts
+++ b/sdk/packages/node/observability/src/telemetry-system/index.ts
@@ -229,21 +229,26 @@ export async function shutdownOtel(): Promise<void> {
   // Stop reconnecting/queuing before flushing so a final forceFlush() can't hang on a dead engine
   sharedConnection?.beginShutdown()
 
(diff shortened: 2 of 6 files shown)


CryptoQuaternions:
Pushed a follow-up (5bdb139) addressing review feedback + end-to-end validation:

1. **@coderabbitai's catch** — `beginShutdown()` now drains already-queued exporter callbacks when there's no live connection, so a queue built up before shutdown can't leave callbacks unresolved for the flush to wait on.

2. **Found while validating end-to-end** — with fail-fast now reachable, `BatchSpanProcessor.forceFlush()` *rejects* on the dropped export, and `shutdownOtel()` propagated that as an uncaught exception (i.e. the original hang would have become a crash). `shutdownOtel()` now treats the final flush as best-effort and swallows those rejections so teardown always completes.

Validated against the real OTel pipeline (init → emit span → `shutdownOtel()`):

- engine unreachable → resolves in ~0 ms (no hang, no crash)
- engine unreachable with a span queued before shutdown → resolves in ~0 ms
- engine reachable → the span is still flushed and received by the engine, then resolves

`pnpm test` (19), `type-check`, and `build` all green.

guibeira (approved):

guibeira:
Thanks @CryptoQuaternions for finding the issue and fixing this.

PR #1877, opened 2026-06-17
docs: guide users through the "KVM not accessible" worker error

### LINKED ISSUE #1854: WSL2: iii worker add fails with "KVM not accessible" during VM startup
## Summary

I am unable to start local workers when running iii inside WSL2. The worker sandbox is created successfully, dependencies are installed/cached, but the worker process fails during VM startup with a KVM-related error.

I'm unsure whether this is:

* a bug,
* an unsupported environment,
* a WSL2 compatibility limitation,
* or an issue with how iii detects KVM availability.

I would appreciate clarification on the expected level of WSL2 support for local workers.

---

### Host

* Windows 11
* Docker Desktop with WSL integration enabled

### WSL

* Ubuntu 26.04
* WSL Version: 2.7.8.0
* Kernel: 6.18.33.1-1

### iii

* Installed via the official installation method
* Using the Quickstart template from the documentation

---

## Steps to Reproduce

Create and enter the Quickstart project:

```bash
iii project init quickstart --template quickstart
cd quickstart
```

Start the engine:

```bash
iii --config config.yaml
```

In a second terminal:

```bash
iii worker add ./workers/math-worker
```

---

## Expected Behavior

The worker should start successfully and register its functions with the engine as described in the Quickstart documentation.

Example:

```text
✓ Worker math-worker added
✓ Worker auto-started
```

and:

```bash
iii trigger math::add a=2 b=3
```

should return a result.

---

## Actual Behavior

The worker never becomes ready and eventually exits.

Output:

```text
→ Adding math-worker (local)...

sandbox: prepared (rootfs + deps cached)
process: dead pid=2695 (stale pidfile)

error: VM execution failed: KVM not accessible
```

Worker logs report:

```text
error: VM execution failed: KVM not accessible -- /dev/kvm exists but current user lacks permission
```

The worker remains queued and never registers with the engine.

---

## Additional Observations

* The engine starts successfully.
* The worker configuration is detected correctly.
* Sandbox creation succeeds.
* Dependency installation succeeds.
* Cached dependencies are reused on subsequent runs.
* Failure occurs specifically during VM startup.

This suggests the failure is occurring after dependency resolution and during the virtualization stage.

---

## Questions

1. Is WSL2 officially supported for running local iii workers?
2. Are VM-backed workers expected to work inside WSL2?
3. Does iii require native Linux KVM support for local workers?
4. Is there a non-KVM worker runtime available for development environments?
5. If WSL2 is unsupported, could the documentation explicitly mention this?
6. If WSL2 is supported, could the CLI provide a more specific diagnostic than "current user lacks permission" when KVM is unavailable due to platform limitations?

---

## Suggested Improvement

If WSL2 is not a supported environment for VM-backed workers, it would be helpful if iii could detect this scenario and emit a clearer message, for example:

```text
WSL2 detected.
VM-backed workers require native Linux KVM support and are not currently supported in this environment.
```

This would make troubleshooting significantly easier for Windows developers.

guibeira:
## Problem

Closes #1854.

On Linux hosts where the current user is not in the `kvm` group (common on WSL2), starting a worker fails with `KVM not accessible`. Workers boot inside a microVM that needs read/write access to `/dev/kvm`, but nothing in the docs or the error output tells the user how to recover.

## Fix

- **Docs:** add a `Warning` callout at Quickstart step 3 ("Start the Python worker") — the exact point where the worker boots its microVM and the error surfaces. It explains the microVM/KVM requirement and the recovery steps: add the user to the `kvm` group and restart the session (`wsl --shutdown` on WSL2).
- **Worker:** both KVM pre-flight failures in `crates/iii-worker/src/cli/vm_boot.rs` (`KVM not available` and `KVM not accessible`) now point to that doc section, so the error is self-resolving from the terminal.

## Notes

- Applied to both the `next` (0.20.x) and current (0.19.x) doc versions, plus the rendered `.skill.md` files.
- Verified the existing `check_kvm_at_path` test still passes (asserts on `"does not exist"`, unchanged).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Linux “KVM not accessible” error messaging by appending a direct troubleshooting link when `/dev/kvm` is missing or access is denied.

* **Documentation**
  * Added a new Troubleshooting page (and “next” variant) with a dedicated “KVM not accessible” section, including `/dev/kvm` access guidance and WSL2 recovery steps (`wsl --shutdown`).
  * Updated the Quickstart to show a Linux warning callout after starting the Python worker, pointing to the troubleshooting section.
  * Updated documentation navigation so the new troubleshooting entry appears in the Getting Started flow.


crates/iii-worker/src/cli/vm_boot.rs
```diff
@@ -212,14 +212,16 @@ fn check_kvm_available() -> Result<(), String> {
 fn check_kvm_at_path(kvm: &std::path::Path) -> Result<(), String> {
     if !kvm.exists() {
         return Err("KVM not available -- /dev/kvm does not exist. \
-             Ensure KVM is enabled in your kernel and loaded (modprobe kvm_intel or kvm_amd)."
+             Ensure KVM is enabled in your kernel and loaded (modprobe kvm_intel or kvm_amd). \
+             See https://iii.dev/docs/quickstart#3-start-the-python-worker"
```
sergiofilhowz: why python?

sergiofilhowz: i think we should have a section for troubleshooting in the docs

sergiofilhowz: it's much less likely to be outdated

guibeira: It's because it happens at that moment:
image

Okay, i agree, we should have and section for this.

sergiofilhowz (dismissed):

crates/iii-worker/src/cli/vm_boot.rs
```diff
@@ -212,14 +212,16 @@ fn check_kvm_available() -> Result<(), String> {
 fn check_kvm_at_path(kvm: &std::path::Path) -> Result<(), String> {
     if !kvm.exists() {
         return Err("KVM not available -- /dev/kvm does not exist. \
-             Ensure KVM is enabled in your kernel and loaded (modprobe kvm_intel or kvm_amd)."
+             Ensure KVM is enabled in your kernel and loaded (modprobe kvm_intel or kvm_amd). \
```
anthonyiscoding, line 215:
```suggestion
             Ensure KVM is enabled in your kernel and loaded (modprobe kvm_intel or kvm_amd). \
             This can also occur on Windows with WSL. \
```

docs/next/troubleshooting.mdx
```diff
@@ -0,0 +1,24 @@
...
+Workers boot inside a microVM that needs read/write access to `/dev/kvm`. On Linux hosts where your
+user is not in the `kvm` group (common on WSL2), starting a worker fails with `KVM not accessible`.
+
+Add your user to the `kvm` group:
+
+```bash
+sudo usermod -aG kvm $USER
+```
+
+Then restart your session so the new group membership applies. On WSL2, run `wsl --shutdown` from a
```
anthonyiscoding, line 21: A search shows some other things that may need to be done. Can anyone with Windows test? https://serverfault.com/questions/1043441/how-to-run-kvm-nested-in-wsl2-or-vmware

anthonyiscoding (changes requested):

docs/next/troubleshooting.mdx
guibeira: During my setup for Windows, enabling Hypervisor was a requirement to install WSL2. I think we can assume if the user already has WSL2, we don't need to specify this.

anthonyiscoding (dismissed):

PR #1897, opened 2026-06-22
Preserve state keys

### LINKED ISSUE #1896: iii-console states view should preserve semantic state keys instead of item-N labels
## Summary

The iii-console `/states` table can render rows as positional labels such as `item-0`, `item-1`, etc. when browsing state groups, even when the underlying `iii-state` entries were written with semantic keys.

This makes operational state hard to inspect from the console. For stateful runtime data, the row label should preserve keys like:

- `worker`
- `health`
- `last_test_suite`
- `last_enqueued_job`
- `source-watch.worker-source`
- `files/worker/config.yaml`

Instead, the UI can show uninformative `item-N` rows, which hides the state contract and forces users to open every row to discover what it is.

## Expected behavior

When the state API returns a key/value map or item list that includes the key, the console should display the real key as the primary row label. If the backend response lacks keys for a list shape, the console should display a clear fallback like `(missing key)` and surface a console/backend issue, rather than silently inventing `item-N` labels.

## Why this matters

For runtime systems using iii-state as an operator-facing source of truth, semantic keys are part of the contract. Positional labels make it look like the worker wrote bad state even when the backend has meaningful keys.

## Suggested fix

Audit the `/states` view normalization path and prefer explicit fields in this order:

1. `item.key`
2. tuple/map key from `Object.entries(...)`
3. `item.state_key` embedded in the value
4. only as a final fallback, a clear missing-key label

McHersheys:
Fixes #1896.

Tests:
- `corepack pnpm --filter console-frontend test -- src/api/state/state.test.ts`
- `corepack pnpm --filter console-frontend lint`
- `corepack pnpm --filter console-frontend build`

## Summary by CodeRabbit

* **Improvements**
  * Enhanced state listing to normalize multiple response shapes and reliably extract keys/values, including clear placeholder keys when identifiers are missing.
  * State listing results now consistently return keyed entries (`key` + `value`) instead of value-only arrays.
* **New Features**
  * Added a key/value index listing capability in the key-value store layer.
* **Tests**
  * Expanded and updated normalization and listing tests to cover multiple formats and verify keyed output.

### DIFF
--- a/console/packages/console-frontend/src/api/state/state.test.ts
+++ b/console/packages/console-frontend/src/api/state/state.test.ts
@@ -1,0 +1,66 @@
+import { afterEach, describe, expect, it, vi } from 'vitest'
+import { fetchStateItems } from './state'
+
+function mockStateItems(payload: unknown): void {
+  vi.stubGlobal(
+    'fetch',
+    vi.fn(async () => ({
+      ok: true,
+      json: async () => payload,
+    })),
+  )
+}
+
+describe('fetchStateItems', () => {
+  afterEach(() => {
+    vi.unstubAllGlobals()
+  })
+
+  it('uses explicit item keys', async () => {
+    mockStateItems({ items: [{ key: 'worker', value: { status: 'ready' } }] })
+
+    const { items } = await fetchStateItems('runtime')
+
+    expect(items[0]).toMatchObject({
+      groupId: 'runtime',
+      key: 'worker',
+      value: { status: 'ready' },
+      type: 'object',
+    })
+  })
+
+  it('uses map keys for object-shaped item responses', async () => {
+    mockStateItems({ items: { 'files/worker/config.yaml': { raw: true } } })
+
+    const { items } = await fetchStateItems('runtime')
+
+    expect(items[0].key).toBe('files/worker/config.yaml')
+    expect(items[0].value).toEqual({ raw: true })
+  })
+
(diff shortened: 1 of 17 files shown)


anthonyiscoding:
Thanks for the PR! We'll take a look and get back to you

guibeira:
Hey @McHersheys do you mind to update your branch? I tested you code and got this result:
This is the main:
image
This is from your branch:
image


The reason this is happening is that the engine is not returning the key. You can confirm this by running this:
`iii trigger state::list scope=runtime | jq`

Do you mind to check this, since the main issue was not addressed?

McHersheys:
Thanks, you were right — the engine list path was still returning values without their state keys.

I updated the branch so `state::list` now returns key/value entries from the built-in, Redis, and bridge adapters, and updated the SDK/console types/tests to expect `[{ key, value }]`.

Verified locally:
- `cargo check -p iii --lib`
- `pnpm --filter console-frontend test -- src/api/state/state.test.ts`
- `pnpm --filter console-frontend lint`
- `pnpm --filter console-frontend build`
- `pnpm --dir sdk/packages/node/iii build`
- `pnpm --dir sdk/packages/node/iii-browser build`
- `uv run --extra dev ruff check src/iii/state.py tests/test_state.py`

One caveat: `cargo test -p iii state --lib --no-default-features` exceeded the 600s local command limit while compiling/linking the engine test binary, so I could not complete that Rust test run in this environment.

PR #1910, opened 2026-06-25
Fix Docker data directory permissions

### LINKED ISSUE #1883: docker compose up crash-loops on Linux: configuration worker can't create ./data (permission denied)
## Summary

`iii project init --docker && docker compose up` crash-loops on a **Linux host** because the `configuration` worker cannot create its data directory. The generated `docker-compose.yml` mounts no writable data volume, and the published image runs as a non-root user (`65532`) with no writable `./data`.

Tested on Linux with engine `0.19.4`. (Often not reproducible on macOS Docker Desktop, whose file-sharing layer ignores UID ownership.)

## Reproduction

```bash
iii project init --docker
docker compose up
```

## Actual

The engine exits on boot:

```
failed to create configuration directory './data/configuration': Permission denied
```

…and restarts in a loop.

## Expected

A fresh `init --docker` + `up` should boot and persist state with no manual edits.

## Root cause

1. The `configuration` worker's `fs` adapter defaults to `./data/configuration`:

   `engine/src/workers/configuration/adapters/fs.rs:38`
   ```rust
   pub(crate) const DEFAULT_DIRECTORY: &str = "./data/configuration";
   ```
   With `WORKDIR /app` (set by the generated Dockerfile), this resolves to `/app/data/configuration`.

2. `/app` is owned by root and the image runs as UID `65532` (base: `gcr.io/distroless/cc-debian12:nonroot`, `engine/Dockerfile`). The non-root user cannot create `/app/data`.

3. The generated compose mounts only `./config.yaml:ro` — there is **no writable data volume** (templates live in `iii-hq/templates`; mirrored at `engine/tests/fixtures/templates/docker/docker-compose.yml`).

4. Even adding a plain named volume does not fix it: the stock image does not pre-create the mount path, so Docker creates the named volume as `root:root` and `65532` still cannot write.

## Proposed fix

- **`engine/Dockerfile`**: pre-create `/app/data` (and `/data`) owned by `65532` via a small build stage (distroless has no shell), so an empty named volume mounted there inherits writable ownership.
- **`iii-hq/templates` `docker/docker-compose.yml`** (and the mirrored fixture): mount a named `iii_data` volume at `/app/data` and declare it under `volumes:`.

After both, `init --docker && up` works and persists out of the box.

## Related minor cleanup

The generated assets and `engine/Dockerfile` still `EXPOSE`/map port `9464` (former Prometheus endpoint) which nothing listens on. Worth dropping in the same change.

## Workaround (for users today)

Point the `configuration` worker at a writable, mounted path and bind-mount it (or use an image that creates the dir owned by `65532`):

```yaml
# config.yaml
workers:
  - name: configuration
    config:
      adapter:
        name: fs
        config:
          directory: /data/configuration
```

snvtac:
## What

- Pre-create `/app/data` and `/data` as UID/GID 65532 in the published distroless engine image.
- Mirror the Docker template fixture to mount `iii_data` at `/app/data`.
- Add project-init e2e assertions so generated Docker assets keep the writable data volume.
- Drop the stale `9464` exposure from the engine image and generated Docker fixture.

## Why

Fixes #1883. On Linux, `iii project init --docker && docker compose up` can crash-loop because the configuration worker defaults to `./data/configuration` under `/app`, while the image runs as non-root UID 65532 and has no writable data directory.

## Notes

- The live generated template is fetched from `iii-hq/templates`; companion template PR: https://github.com/iii-hq/templates/pull/31
- Validation run: `rustfmt --check engine/tests/project_init_e2e.rs`, YAML parse for the compose fixture, semantic check for `iii_data:/app/data`, and diff whitespace check.
- Full `cargo test` / Docker smoke were not run here: full Git clone repeatedly stalled in this environment, and Docker CLI is unavailable.

## Summary by CodeRabbit

* **New Features**
  * Updated the container build to use a safer multi-stage approach.
  * Updated the generated Docker Compose setup to persist application data using a named volume mounted at `/app/data`, improving retention across restarts.
* **Bug Fixes**
  * Reduced the container’s exposed ports by removing the previously exposed `9464` port, while keeping `49134`, `3111`, and `3112`.


guibeira:
Hey @snvtac thanks for another contribution, tested your changes, and they worked well. Do you mind fixing the conflicts?

PR #1919, opened 2026-06-29
fix(configuration): numeric fields can be used with env variables

### LINKED ISSUE #1916: Env-driven config for the new format is giving errors
Hello, I am migrating the config from iii-config.yaml to data/configuration with ${ENV} placeholders and pushing those as expected from version 0.16.7 (included). Env-driven config works for string fields (host, service_name, etc..). But it fails for typed fields,  in my case, it fails for "port" field.  Maybe there is a bug in this new config structure with environment-driven pattern:
```invalid type: string "8080", expected u16  →  engine drops the value, falls back to defaults```
basically I set port: ${HTTP_PORT:3111} in data/configuration

sergiofilhowz:
## Summary

Fixes configuration worker reliability around HTTP port changes, on-disk layout, file-based hot-reload, and env-var expansion in numeric fields.

- **`${VARIABLES}` in numeric fields** — lone placeholders like `port: ${HTTP_PORT:3111}` are env-expanded and YAML-coerced to the correct scalar type (int/bool/float), so schema validation passes instead of treating them as strings. Fixes #1916.
- **Config directory** — default store moved from `./data/configuration` → `./iii-config`. Existing entries are auto-migrated on first boot; explicit `directory:` overrides are untouched.
- **File-edit hot-reload** — direct edits to `./iii-config/*.yaml` are picked up via the fs adapter watcher (canonicalized watch path fixes macOS FSEvents on relative paths). External edits fan out the same rebind/apply path as `configuration::set`; internal writes are deduped to avoid reload loops.
- **HTTP port binding** — `iii-http` now gracefully shuts down the old listener on host/port rebind (closes idle keep-alive connections, not just the accept socket), with a hard-abort safety net so a stuck connection can’t pin the old port indefinitely.

## Test plan

- [ ] `cargo test -p iii http_configuration_e2e` — port rebind via `configuration::set` and via direct file edit
- [ ] Edit `./iii-config/iii-http.yaml` port on disk; confirm new port serves and old port stops accepting
- [ ] Set `port: ${HTTP_PORT:3111}` (or similar) in a persisted config entry; confirm it resolves to an integer
- [ ] Upgrade from an install with `./data/configuration/` entries; confirm one-time migration to `./iii-config/`
- [ ] Smoke: `iii` boots, HTTP worker listens, config changes apply without restart

## Summary by CodeRabbit

* **New Features**
  * Configuration files now use `./config` by default, with a one-time migration from the previous location when possible.
  * Templated values like `${VAR:default}` are now resolved more reliably, with numeric values converted automatically when appropriate.
  * HTTP server updates now rebind more gracefully, including cleaner shutdown of old listeners and open connections.

* **Bug Fixes**
  * External config edits are now validated before applying, reducing broken or invalid settings from taking effect.
  * Missing environment variables without defaults now fall back to the last valid or seed configuration instead of causing failures.

### DIFF
--- a/engine/config.yaml
+++ b/engine/config.yaml
@@ -154,7 +154,7 @@ workers:
       adapter:
         name: fs
         config:
-          directory: ./iii-config
+          directory: ./config
       # 0 disables TTL cleanup. Set >0 (in seconds) to delete a configuration
       # entry whose last subscriber trigger has been unregistered for that long.
       ttl_seconds: 0
--- a/engine/config.yaml
+++ b/engine/config.yaml
@@ -154,7 +154,7 @@ workers:
       adapter:
         name: fs
         config:
-          directory: ./data/configuration
+          directory: ./iii-config
       # 0 disables TTL cleanup. Set >0 (in seconds) to delete a configuration
       # entry whose last subscriber trigger has been unregistered for that long.
       ttl_seconds: 0
--- a/engine/src/logging.rs
+++ b/engine/src/logging.rs
@@ -796,7 +796,7 @@ mod tests {
     fn boot_read_path_constants_align_with_fs_adapter() {
         use crate::workers::configuration::adapters::fs;
         assert_eq!(fs::ADAPTER_NAME, "fs");
-        assert_eq!(fs::DEFAULT_DIRECTORY, "./iii-config");
+        assert_eq!(fs::DEFAULT_DIRECTORY, "./config");
         assert_eq!(fs::FILE_EXTENSION, "yaml");
         assert_eq!(
             format!(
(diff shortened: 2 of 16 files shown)


anthonyiscoding:
We should make this a rule in our editors to never use `iii-`: `./iii-config` should be `./config`

anthonyiscoding:
Will configuration for workers be stored locally to the worker or all stored at the iii engine itself?

My belief is that it should be wherever the worker is so there's no drift and less coupling between a worker and a specific iii instance.

sergiofilhowz:
> Will configuration for workers be stored locally to the worker or all stored at the iii engine itself?
> 
> My belief is that it should be wherever the worker is so there's no drift and less coupling between a worker and a specific iii instance.

in iii engine globally

guibeira (approved):

PR #1928, opened 2026-07-01
feat: engine::register_trigger + first-class invocation metadata

ytallo:
## Summary

Adds `engine::register_trigger` / `engine::unregister_trigger` and makes per-invocation **metadata** a first-class argument that travels as a distinct channel through the entire invocation path — engine, wire protocol, and all four SDKs — instead of being folded into the payload.

Implements the spec in [MOT-3747](https://linear.app/motia/issue/MOT-3747/engineregister-trigger-unregister-trigger-with-metadata-round-trip).

## What's here

**`engine::register_trigger` / `engine::unregister_trigger`**
- Register a trigger whose fire is delivered to `function_id`, with arbitrary `metadata` recoverable at fire time. Returns the trigger id; unregister is idempotent (reports whether it existed). Honors a session's `allowed_trigger_types` (RBAC parity with the message path).

**Metadata as a first-class argument**
- `metadata: Option<Value>` is threaded through `EngineTrait::call` (the third argument), `Invocations::handle_invocation`, `Function::call_handler`, and the `HandlerFn` / `FunctionHandler` signatures.
- It rides on the wire as a new field on `Message::InvokeFunction` — additive with `#[serde(default, skip_serializing_if = "Option::is_none")]`, so peers that neither send nor expect the field are unaffected.
- The target function receives metadata as a separate argument; the payload is left untouched (no `__metadata` injection).

**No proxy function**
- `register_trigger` binds the trigger directly to `function_id` with `metadata` stored on the `Trigger`. Every trigger-fire path (cron, state, stream, http, configuration, and the engine's own `fire_triggers`) passes `trigger.metadata` at fire time. This removes the previous per-registration proxy function and its derived-id garbage collection on unregister and on worker disconnect.

**SDKs (all four)** — no new registration functions, no signature changes; metadata flows through each language's existing idiom:

| SDK | Handler reads metadata | Invocation-time send |
| --- | --- | --- |
| Rust | 2-arg closures accepted by the existing `RegisterFunction::new` / `new_async` (`\|data, meta: Option<Value>\|`; 1-arg closures unchanged) | `TriggerRequest { … }.metadata(json!(…))` builder |
| Node | optional trailing param: `(data, metadata?)` | `trigger({ …, metadata })` |
| Python | `def handler(data, metadata=None)` or `*, metadata=None` — forwarded **only** when the handler declares a parameter named `metadata` | `trigger({ …, "metadata": … })` |
| Go | `iii.MetadataFromContext(ctx)` — rides on the handler's ctx; `Handler` / `TypedHandler` signatures untouched | `TriggerRequest.Metadata` |

Registration-level metadata (stored with the function/trigger, distinct from the per-invocation sidecar) is attached the same way: Go `RegisterFunctionOptions{Metadata: …}` variadic option, Rust `RegisterFunction::metadata(…)` builder, and the existing `TriggerTypeRef::register_trigger_with_metadata`.

## Backward compatibility

- **No public API removed or reshaped.** Go registration functions keep their exact pre-existing signatures (including `RegisterFunctionTyped`'s generic inference — verified by compiling all pre-branch call patterns against this branch). Rust keeps the released `TriggerTypeRef::register_trigger_with_metadata` (v0.20.0) and the single-arg `RemoteFunctionHandler` alias; the two-arg dispatch shape is the new `RemoteFunctionHandlerWithMetadata` alias.
- **Existing handlers behave identically.** Python forwards metadata only to handlers that explicitly declare `metadata`, so handlers with unrelated optional parameters, `*args`, or `**kwargs` keep their exact old call shape. Node/Go handlers are untouched by design.
- **Wire format is additive in both directions**: the field is omitted when absent (serde `default` + `skip_serializing_if`, Go `omitempty`, Node `undefined` omission, Python `exclude_none`), so old SDK ↔ new engine and new SDK ↔ old engine both work.
- One semver note: the Rust SDK's public `Message::InvokeFunction` variant gains an optional `metadata` field, so external exhaustive matches on that enum need a `..` — the next release should be a minor bump (0.21.0).

## Test plan

- [x] `cargo check --workspace --tests` — clean
- [x] Engine unit tests for touched modules (protocol, function, trigger, condition, engine_fn, cron, rest_api, worker_connections) pass, including new tests: wire round-trip + backward-compat, direct trigger binding + clean-payload delivery, and metadata forwarded over `WorkerConnection`
- [x] Rust SDK: `cargo build` / `clippy` / `fmt` clean; 53 lib tests + 17 doc-tests pass (incl. metadata-aware handlers, builder-optional metadata, legacy struct-literal `trigger()`, restored `register_trigger_with_metadata`, and a compile-time guard pinning the 1-arg `RemoteFunctionHandler` shape)
- [x] Go SDK: `go build` / `go vet` (incl. `integration` tag) / `gofmt` clean; 75 unit tests pass (incl. `MetadataFromContext` present/absent delivery for raw and typed handlers)
- [x] Go source-compatibility harness: all pre-branch call patterns (2-arg closures, `Handler` vars, 5-arg `RegisterTrigger`, explicit and fully-inferred `RegisterFunctionTyped`) compile against this branch
- [x] Node SDK: `pnpm build` + metadata test suite — 10/10
- [x] Python SDK: metadata suite (17 tests: mode detection incl. unrelated-second-param/`*args`/`**kwargs` back-compat, handler delivery, outbound trigger) plus registration/dispatch unit suites — pass
- [ ] Cross-process end-to-end (live engine + worker) delivering metadata to a worker handler

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end per-invocation metadata support across function invocations, triggers, queues, HTTP routing, streams, and cron.
  * Function handlers can optionally receive metadata as a separate argument.
  * Added engine trigger registration with metadata; trigger unregistration is now idempotent and reports whether a trigger existed.
* **Bug Fixes**
  * Preserved backward compatibility when metadata is absent (legacy handlers still work).
* **Documentation**
  * Updated SDK docs and examples (Go/Node/Python/Rust) to send and read invocation metadata.


docs/next/api-reference/sdk-rust.mdx
```diff
@@ -657,7 +669,7 @@ the invocation.
 
 #### TriggerRequest
 
-Request object for `trigger()`. Matches the Node/Python SDK signature:
+Request object for `trigger()`. Matches the previous Rust SDK shape:
```
anthonyiscoding: I realize it was here already but this note isn't that useful for people. The matching of shapes should just be an inherent thing of our SDK.


```suggestion
Request object for `trigger()`.
```

docs/next/api-reference/sdk-node.mdx
```diff
@@ -382,10 +382,15 @@ in invocation payloads to pass channel endpoints between workers.
...
+payload and an optional per-invocation `metadata` value, and returns the
+result.
+
+`metadata` is arbitrary JSON travelling on a separate channel from the
+payload. It is `undefined` when the caller did not attach any. Existing
+single-argument handlers keep working -- they ignore the extra argument.
 
 ```typescript
-type RemoteFunctionHandler = (data: TInput) => Promise<TOutput>
+type RemoteFunctionHandler = (data: TInput, metadata: unknown) => Promise<TOutput>
```
anthonyiscoding: Since it's arbitrary json would type be object here?

docs/next/api-reference/sdk-rust.mdx
```diff
@@ -136,41 +136,52 @@ The routing behavior depends on the `action` field of the request:
...
 | --- | --- | --- | --- |
-| `request` | impl Into&lt;[`TriggerRequest`](#triggerrequest)&gt; | Yes | - |
+| `request` | impl Into&lt;[`TriggerRequestWithMetadata`](#triggerrequestwithmetadata)&gt; | Yes | - |
 
 #### Example
 
 ```rust
 // Synchronous
-let result = worker.trigger(TriggerRequest {
+let result = iii.trigger(TriggerRequest {
```
anthonyiscoding: Our standard is always worker.trigger() now. Please change all of these back (obviously in the source code examples and not here).
```suggestion
let result = worker.trigger(TriggerRequest {
```

docs/next/using-iii/triggers.mdx
```diff
@@ -85,6 +85,16 @@ You bind triggers to functions via the `function_id`. The trigger declares its `
...
+### Trigger metadata
+
+The optional `metadata` field on a trigger registration (`null` / `None` in the examples above)
+is arbitrary JSON stored with the trigger and delivered to the bound function on every fire as a
+distinct argument alongside the payload, never folded into it. A target function shared by many
+triggers can use it to recover which registration fired and with what context. See
+[Receive per-invocation metadata](../creating-workers/functions#receive-per-invocation-metadata)
+for handler-side access in each SDK. Callers can attach the same sidecar on direct invocations
+via `trigger({ …, metadata })`.
+
```
anthonyiscoding, line 103:
```suggestion
The optional `metadata` field on a trigger registration (`null` / `None` in the examples above)
is arbitrary JSON stored with the trigger and delivered to the receiving function as a
distinct argument alongside the payload. It is useful for providing contextual information about
the trigger or execution context to the receiving function.

A target function shared by many triggers can use it to recover which registration fired and 
with what context. See
[Receive per-invocation metadata](../creating-workers/functions#receive-per-invocation-metadata)
for handler-side access in each SDK.

<Note>
  `metadata` can be provided both via `registerTrigger` and direct `trigger()` invocations
</Note>
```

docs/next/sdk-reference/engine-sdk.mdx
```diff
@@ -124,15 +124,21 @@ trigger fires.
...
 `invocation_id` is omitted on `Void` invocations (the worker has no result channel to reply on).
-`traceparent` and `baggage` carry W3C trace context. `action` is the routing flag (see
-[Trigger actions](#trigger-actions) below); absent / `null` means synchronous.
+`metadata` is optional per-invocation metadata (arbitrary JSON) that travels as a distinct
+channel alongside `data`, never folded into the payload, and is surfaced to the target
+handler as a separate argument. It is set at fire time from a trigger's registered `metadata`
+or attached by an invocation-time caller, and omitted entirely when absent, so peers that
+neither send nor expect it are unaffected. `traceparent` and `baggage` carry W3C trace
+context. `action` is the routing flag (see [Trigger actions](#trigger-actions) below);
+absent / `null` means synchronous.
```
anthonyiscoding, line 144:
```suggestion
The optional `metadata` field on a trigger registration (`null` / `None` in the examples above)
is arbitrary JSON stored with the trigger and delivered to the receiving function as a
distinct argument alongside the payload. It is useful for providing contextual information about
the trigger or execution context to the receiving function. A target function shared by many triggers can use it to recover which registration fired and with what context.

`metadata` can be provided both via `registerTrigger` and direct `trigger()` invocations

`traceparent` and `baggage` contain W3C trace
context. `action` is the routing flag (see [Trigger actions](#trigger-actions) below);
absent / `null` means synchronous.
```

docs/next/creating-workers/functions.mdx
```diff
@@ -228,6 +228,50 @@ you gate access on the fields you define.
   </Tab>
 </Tabs>
 
+## Receive per-invocation metadata
+
+Distinct from the registration metadata above, each invocation can carry its own optional
+`metadata` sidecar: arbitrary JSON that travels as a separate channel from the payload, never
+folded into it. The engine sets it at fire time from a trigger's registered `metadata` (see
+[trigger metadata](../using-iii/triggers#trigger-metadata)), or a caller attaches it directly on
+`trigger({ …, metadata })`. Handlers opt in; existing handlers keep working unchanged.
```
anthonyiscoding:
```suggestion
Distinct from the registration metadata above, each function can also receive an optional
`metadata` object that is sent as arbitrary JSON. When a trigger invokes a function it can
use this `metadata` object to provide trigger or execution context to the receiving function (see
[trigger metadata](../using-iii/triggers#trigger-metadata)) for more information.
```

anthonyiscoding (changes requested):
Looks good overall, some suggested changes for the docs to make in the rustdoc/jsdoc/etc.

API Doc generation is working.

docs/next/api-reference/sdk-node.mdx
ytallo: `unknown` is intentional here, not an oversight. `metadata` is arbitrary JSON per the wire contract (Rust's `Option<Value>` in `protocol.rs`), so it isn't guaranteed to be an object — it could be a string, array, number, or bool. `object` would incorrectly reject those.

This is actually the mirror image of another finding on this PR: the Python SDK's `metadata: dict[str, Any] | None` in `iii_types.py` was flagged as *too strict* for the same reason. Keeping Node's `unknown` (and loosening Python's type) is the more consistent fix across SDKs — let me know if you'd rather standardize on object-only metadata instead, but that would mean tightening the Rust/wire contract too.

ytallo: Follow-up: went one better than `unknown` — added an exported `JsonValue` type (the TS equivalent of `serde_json::Value`) and the handler is now `(data: TInput, metadata?: JsonValue)`. Unlike `object` it still admits the scalars/arrays the wire allows, and unlike `unknown` it lets handlers index into metadata after a normal object check. `TriggerRequest.metadata` stays `unknown` on the input side so interface-typed values remain assignable.

engine/src/trigger.rs
```diff
@@ -329,33 +329,35 @@ impl TriggerRegistry {
...
 
         if let Some(tt) = self.trigger_types.get(&trigger.trigger_type) {
-            let result: Result<(), anyhow::Error> =
-                tt.registrator.unregister_trigger(trigger.clone()).await;
-
-            result?
+            tt.registrator.unregister_trigger(trigger.clone()).await?;
         }
 
         self.triggers.remove(&id);
```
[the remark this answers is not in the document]
ytallo: Deferring this one: the unregister race requires a same-id re-registration overlapping an in-flight registrator call (caller-stable ids on the message path), which is narrow enough to split out. Keeping this PR scoped to the register_trigger/metadata feature; the remove_if fix will land as a follow-up.

anthonyiscoding (approved):

PR #1929, opened 2026-07-02
fix(iii-queue): add per-queue dispatch_timeout_ms to bound lost dispatches

### LINKED ISSUE #1927: fn-queue invocation can wedge unacked forever — no per-invocation dispatch timeout to requeue a lost engine→worker dispatch
## Summary

When the engine dequeues a **queued function invocation** from a RabbitMQ fn-queue and dispatches it to a worker over the WebSocket bus, a **lost or never-returned dispatch leaves the AMQP message in-flight (`unacked`) indefinitely**. The handler is never invoked and the invocation is never retried, because the queue's `max_retries`/`backoff` (retry→DLQ) only fires on a **nack/failure**, not on a message that was delivered to the engine's consumer but never completed. The only thing that eventually frees it is RabbitMQ's broker-level `consumer_timeout` (~30 min default).

Because the **engine** (not the worker) is the AMQP consumer, **restarting the worker does not recover it** — the engine holds the unacked grip.

## Environment

- Engine/SDK **0.20.0**, `docker compose` runtime (engine + Python SDK workers).
- `iii-queue` with the **RabbitMQ adapter**.
- Example queue config (a `standard` parallel fan-out queue):
  ```yaml
  my-dim-queue:
    type: standard
    max_retries: 2
    concurrency: 5
    backoff_ms: 1000
  ```

## Repro signature

A queued function (`type: standard`) fanned out to N parallel invocations; **one invocation's result never lands**. On the broker:

```
iii.__fn_queue::my-dim-queue.queue   messages=1  messages_ready=0  messages_unacknowledged=1  consumers=1
```

- The registered handler **never logs / never runs** for that invocation.
- The message **survives a worker restart** (stays `unacked=1`, not requeued to `ready`).
- The corresponding `::retry.queue` and `::dlq.queue` stay empty — no retry was ever attempted.
- It self-heals only after ~30 min (RabbitMQ's broker `consumer_timeout` closes the channel and requeues).

## Analysis

1. Workers connect to the engine purely over the WS bus; they have **no AMQP client**. So the `consumers=1` on the fn-queue is the **engine's** RabbitMQ consumer.
2. The engine dequeues the invocation (now broker-side `unacked`), then dispatches it to the worker over WS.
3. If that engine→worker dispatch is lost or the worker never returns a completion, the engine appears to have **no per-invocation dispatch/ack deadline** that would `nack`→requeue the message. It stays `unacked` until the broker's `consumer_timeout`.
4. `max_retries`/`backoff_ms`/retry→DLQ are **failure-driven** (they fire when a handler nacks/errors), so a *held-but-never-processed* message never enters that path.

Net effect: the queue's "at-least-once with retry→DLQ" guarantee does **not** hold for the *dispatched-but-uncompleted* case — only for the *handler-failed* case. A single transient WS-dispatch hiccup silently strands one invocation for up to the broker timeout (and consumes one of the queue's `concurrency` slots for that whole window).

## Ask

A **per-invocation dispatch/ack deadline** (ideally configurable per queue, e.g. `dispatch_timeout_ms`) that, when an invocation has been dispatched to a worker but not completed within the deadline, **nacks→requeues** it through the normal retry→DLQ path — so at-least-once delivery actually holds for a lost dispatch without waiting on RabbitMQ's 30-min broker `consumer_timeout`.

If there is already a recommended way to bound this (a queue option I've missed, or a heartbeat-based liveness requeue), a pointer would be very welcome.

## Workaround (worker-side)

For now we recover at the application layer: when a fan-out result is still missing past a deadline safely above the slowest legitimate handler runtime (and below the overall barrier timeout), we **re-enqueue that invocation once** to get a fresh dispatch — which near-always succeeds, since the wedge is a transient delivery race. The original message stays `unacked` until the broker timeout, then redelivers harmlessly (our handlers are idempotent). This unsticks us but doesn't address the underlying engine behavior, hence this report.

ytallo:
## What

Adds an opt-in per-queue `dispatch_timeout_ms` to `queue_configs` entries. When set, a fn-queue invocation that has been dispatched but not completed within the deadline is abandoned and nacked back through the normal retry→DLQ path, releasing its concurrency slot. When unset (the default), behaviour is unchanged.

- New `Engine::call_with_timeout(function_id, input, Option<Duration>)`: bounds the invocation, and on elapse evicts the abandoned invocation from **both** tracking structures — the engine-global invocations map and the owning worker's in-flight set — then returns a `dispatch_timeout` error. `EngineTrait::call` now delegates to it with `None` (no behaviour change for other callers).
- The fn-queue consumer (`QueueWorker::spawn_consumer`) dispatches through it, so a timeout lands in the existing nack→retry→DLQ arm.
- `dispatch_timeout_ms: 0` is rejected at both the JSON schema (`range(min = 1)`) and the `config.yaml` seed path (`QueueModuleConfig::validate`), mirroring the existing `concurrency`/`max_priority` guards.
- Docs (config field, README, SKILL.md) document the new knob and its caveat: the timeout does **not** cancel work already running on a worker, so a retry can overlap the original attempt — bounded handlers must be idempotent.

## Why

Fixes #1927. When the engine dequeues a fn-queue message and the engine→worker dispatch is lost (or the worker never returns a result), the message's ack/nack was gated behind an unbounded await: the fn-queue consumer awaited the invocation with no deadline, and for a worker-routed (deferred) invocation the engine parked on a oneshot that only resolves on a worker reply or a detected disconnect. A half-open connection resolves neither. Since `max_retries`/`backoff_ms`/DLQ are nack-driven, a dispatched-but-uncompleted message was never retried — it stayed `unacked` until RabbitMQ's broker `consumer_timeout` (~30 min), pinned one of the queue's `concurrency` slots the whole time, and survived worker restarts because the engine holds the AMQP consumer.

## Notes

- **Scope:** covers named fn-queues for all adapters (their consumption funnels through the single dispatch site in `queue.rs`). The topic/subscriber (`durable:subscriber`) delivery path has the same unbounded-await exposure and is tracked separately (MOT-3801), along with centralizing the deadline in `InvocationHandler` and active worker-liveness detection.
- **At-least-once semantics unchanged:** a timed-out dispatch redelivers, same as a broker-timeout redelivery today — the knob just makes recovery configurable per queue instead of fixed at ~30 min.
- Also fixes an eager `unwrap_or(Uuid::new_v4())` in `handle_invocation` that would otherwise draw a discarded UUID per invocation now that the id is generated up front.

## Test plan

- [x] `cargo test -p iii --lib` — 1900 passed (includes 3 new `call_with_timeout` unit tests: deferred-eviction from the invocations map, worker in-flight set cleanup, `None` stays unbounded)
- [x] `cargo test -p iii --test queue_integration` — 15 passed (new: wedged handler dead-letters via timeout; timeout releases the concurrency slot for a queued healthy message; unset timeout leaves a slow handler unbounded)
- [x] Config: parse/default/`Some(0)`-rejection unit tests
- [x] `cargo fmt --check`, `cargo clippy` clean on touched files
- [ ] Manual repro against a real RabbitMQ broker (docker compose) with a wedged worker — verify `messages_unacknowledged` drains at the configured deadline instead of ~30 min

## Summary by CodeRabbit

* **New Features**
  * Added optional per-queue dispatch timeouts for function jobs via `dispatch_timeout_ms` (JSON: `dispatchTimeoutMs`), applied to dispatched work.
* **Bug Fixes**
  * Timed-out dispatches now follow the normal retry → dead-letter path instead of staying in-flight.
  * Improved timeout cleanup so queue capacity/concurrency is freed for subsequent jobs.
  * Long-running handlers remain unaffected when timeouts are unset (bounded only when configured).
* **Tests**
  * Added integration coverage for wedged dispatch recovery, retry/dead-letter behavior, capacity freeing, and unset-timeout success.

### DIFF
--- a/engine/src/engine/mod.rs
+++ b/engine/src/engine/mod.rs
@@ -1818,6 +1818,115 @@ impl Engine {
             .remove_if(function_id, |_, owner| owner == worker_id)
             .is_some()
     }
+
+    /// Like [`EngineTrait::call`], but bounds how long the caller waits for the
+    /// invocation to complete.
+    ///
+    /// When `timeout` is `Some`, a dispatch that does not complete within the
+    /// deadline is abandoned: the pending (worker-routed) invocation is evicted
+    /// from the invocations map AND from the owning worker's in-flight set —
+    /// so a lost or never-returning worker cannot leak either entry — and a
+    /// `dispatch_timeout` error is returned. `None` preserves the unbounded
+    /// behaviour of [`EngineTrait::call`], which simply delegates here.
+    ///
+    /// The timeout does NOT cancel work already running on a worker (the WS
+    /// protocol has no cancel message): the worker keeps executing and its
+    /// eventual result is discarded, while the caller may retry the message.
+    /// In-process handlers, by contrast, ARE drop-cancelled at their next
+    /// await point. Either way, handlers invoked under a timeout must be
+    /// idempotent.
+    ///
+    /// The fn-queue consumer passes the queue's `dispatch_timeout_ms` so a lost
+    /// engine→worker dispatch is nacked back through the normal retry→DLQ path
+    /// instead of wedging the underlying broker message (RabbitMQ leaves it
+    /// `unacked` until its server-side `consumer_timeout`, ~30 min by default).
+    pub async fn call_with_timeout(
+        &self,
+        function_id: &str,
+        input: impl Serialize + Send,
+        timeout: Option<std::time::Duration>,
+        metadata: Option<Value>,
+    ) -> Result<Option<Value>, ErrorBody> {
+        let input = serde_json::to_value(input).map_err(|e| ErrorBody {
+            code: "serialization_error".into(),
+            message: e.to_string(),
+            stacktrace: None,
+        })?;
+
+        let Some(function) = self.functions.get(function_id) else {
+            return Err(ErrorBody {
+                code: "function_not_found".into(),
+                message: format!("Function {} not found", function_id),
+                stacktrace: None,
+            });
+        };
+
+        // Inject current trace context and baggage to link spans as parent-child
+        // Use the tracing span's context directly to ensure proper propagation in async code
+        let ctx = tracing::Span::current().context();
+        let traceparent = inject_traceparent_from_context(&ctx);
+        let baggage = inject_baggage_from_context(&ctx);
+
+        // Generate the invocation id up front so a timeout can evict the
+        // deferred (worker-routed) invocation from the map by id.
+        let invocation_id = Uuid::new_v4();
+        let invocation_fut = self.invocations.handle_invocation(
+            Some(invocation_id),
+            None,
+            function_id.to_string(),
+            input,
+            function,
+            traceparent,
+            baggage,
+            None,
+            metadata,
+        );
+
+        let result = match timeout {
+            Some(duration) => match tokio::time::timeout(duration, invocation_fut).await {
+                Ok(result) => result,
+                Err(_elapsed) => {
+                    // Deadline hit before the handler returned. Evict any
+                    // deferred invocation so a late (or never-arriving) worker
+                    // result cannot leak the map entry, then surface a failure
+                    // so callers route the message back through retry/DLQ.
+                    // A worker-routed dispatch is tracked in TWO places: the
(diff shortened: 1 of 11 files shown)


guibeira (dismissed):
LGTM

PR #1932, opened 2026-07-03
fix(sdk/node): reconnect ChannelWriter after a dropped WebSocket

chrisayl:
## Problem

`ChannelWriter.sendRaw` guards on a tracked `wsReady` boolean instead of the socket's live `readyState`:

- `wsReady` is set `true` in the `'open'` handler but is never reset in the `'close'` or `'error'` handlers.
- `ensureConnected()` early-returns whenever `this.ws` is non-null, regardless of the socket's actual state.

After any connection drop the flag stays stuck `true`, `ensureConnected()` never reconnects, and `sendRaw` calls `.send()` on a dead socket — which throws (or silently drops data) instead of queueing to `pendingMessages` and reconnecting. Both the frame-by-frame `sendChunked` path and direct `sendMessage` calls are affected.

## Fix

- **`sendRaw`** — guard on `this.ws.readyState === WebSocket.OPEN` rather than the `wsReady` flag, and wrap `.send()` in a `try/catch` that requeues to `pendingMessages` if the socket dies between the readiness check and the call.
- **`ensureConnected()`** — reconnect when the existing socket is `CLOSING`/`CLOSED` instead of pinning a terminated socket forever, and reset `wsReady` on the fresh connection.
- **`'close'` / `'error'` handlers** — reset `wsReady = false`.

Net effect: a send issued after a drop is queued and the writer transparently reconnects, flushing the queue once the new socket opens — instead of throwing or losing data.

## Tests

New `tests/channel-writer-readiness.test.ts` (mocks `ws`, no engine required), 6 cases:

- never sends on a dead socket when `wsReady` is stuck true; reconnects and queues instead
- requeues instead of throwing when `send()` throws on a socket that died underneath us
- resets `wsReady` on `'close'`
- resets `wsReady` on `'error'`
- `ensureConnected` reuses a live socket but replaces a dead one
- reconnects after a drop and flushes messages queued against the dead socket

## Verification

`pnpm install`, then the node `iii` package's vitest suite. All mock-based unit suites pass, including the pre-existing `channel-close-delay.test.ts` as a regression guard for the same class. The engine-dependent integration suites require a live engine and were not exercised in this environment.

---

Opening as a **draft** for review.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved connection handling to avoid reusing closed or closing WebSocket connections.
  * Messages are now safely queued and retried if a send fails during a connection race.
  * Connection status is reset more reliably after socket errors and disconnects.
  * After a drop, queued messages are flushed once a new connection is available.

* **Tests**
  * Added coverage for WebSocket readiness, reconnection, and message replay behavior under failure conditions.

### DIFF
--- a/sdk/packages/node/iii/src/channels.ts
+++ b/sdk/packages/node/iii/src/channels.ts
@@ -100,7 +100,18 @@ export class ChannelWriter {
   }
 
   private ensureConnected(): void {
-    if (this.ws) return
+    // A non-null socket is only reusable while it is still connecting or open.
+    // A CLOSING/CLOSED socket is dead and must be replaced — guarding solely on
+    // `this.ws` being non-null would pin a terminated socket forever, so every
+    // subsequent send would target a connection that can never deliver it.
+    if (
+      this.ws &&
+      this.ws.readyState !== WebSocket.CLOSING &&
+      this.ws.readyState !== WebSocket.CLOSED
+    ) {
+      return
+    }
+    this.wsReady = false
     this.ws = new WebSocket(this.url)
 
     this.ws.on('open', () => {
@@ -112,10 +123,12 @@ export class ChannelWriter {
     })
 
     this.ws.on('error', (err) => {
+      this.wsReady = false
       this.stream.destroy(err)
     })
 
     this.ws.on('close', () => {
+      this.wsReady = false
       if (!this.stream.destroyed) {
         this.stream.destroy()
       }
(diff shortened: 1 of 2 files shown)


sergiofilhowz (approved):

sergiofilhowz (approved):

sergiofilhowz:
hi @chrisayl, this is good, thanks for capturing it and fixing it, it'll be pushed in the next release

PR #1941, opened 2026-07-08
docs: update configuration management details and add new configuration page

sergiofilhowz:
- Added "next/using-iii/configuration" to the documentation structure.
- Expanded changelog for version 0.21.0 to include details on real-time configuration management and migration of worker settings to the new configuration worker.
- Updated various documentation pages to reflect the new configuration management approach, emphasizing live edits and the transition from `config.yaml` to individual worker configuration files in `./config/`.


guibeira (dismissed):

docs/next/tutorials/linkly/durable-execution.mdx
```diff
@@ -46,6 +46,13 @@ workers:
   a given queue.
 </Note>
 
+<Note>
```
anthonyiscoding: Tutorials teach new users current behavior so they shouldn't reference previous/changed things. They need to just show the current behavior (ie. what does a user need to do to complete each step).

docs/next/tutorials/linkly/foundations.mdx
```diff
@@ -49,6 +49,12 @@ workers:
 #         file_path: ./data/state_store.db
 ```
 
+<Note>
+  A worker's `config.yaml` block seeds its settings on that worker's first boot. After that boot,
+  settings live in one file per worker under `./config/` and are edited there or from the console —
+  a later `config.yaml` edit won't apply. See [Configuration](/using-iii/configuration).
+</Note>
```
anthonyiscoding: Same here, can drop this note entirely.


```suggestion
```

docs/next/tutorials/linkly/persistence.mdx
```diff
@@ -29,6 +29,13 @@ mkdir -p data
 
 Adjust your database worker in `config.yaml` so that it looks like the below.
 
+<Note>
+  If the `database` worker already booted with its default settings, the `config.yaml` block has
+  been consumed as a seed (and may have been replaced with a breadcrumb comment). In that case make
+  the same change under `value:` in `./config/database.yaml` — it applies live. See
+  [Configuration](/using-iii/configuration).
+</Note>
+
```
anthonyiscoding:
```suggestion
```

docs/next/understanding-iii/engine.mdx
```diff
@@ -46,6 +46,11 @@ commits the new config. Workers that did not change in the diff stay running thr
 only added, removed, or changed Workers are restarted. An invalid config (parse error or validation
 failure) causes the Engine to exit rather than enter an indeterminate state.
 
+Worker *settings* are a separate layer: a Worker that registers a configuration schema reads its
+`config:` block once, as a first-boot seed, and its runtime settings live in the
+[configuration worker](../using-iii/configuration) from then on — updated live, without an engine
+reload.
+
```
anthonyiscoding: This introduces "worker settings" but I'm not sure what is meant here. Should this instead say something like this?


```suggestion
Workers bootstrap their configuration via a local `config.yaml` file once at initialization
after which the local `config.yaml` entry is migrated to the configuration worker and 
can be updated dynamically via the [configuration worker](../using-iii/configuration).

An invalid bootstrap config (parse error or validation failure) causes the Engine to exit.
```

docs/next/understanding-iii/engine.mdx
```diff
@@ -46,6 +46,11 @@ commits the new config. Workers that did not change in the diff stay running thr
 only added, removed, or changed Workers are restarted. An invalid config (parse error or validation
 failure) causes the Engine to exit rather than enter an indeterminate state.
```
anthonyiscoding: I believe this is no longer true and covered by the below. May need to mention if configuration worker's config validation if it does any.

```suggestion
```

docs/next/using-iii/configuration.mdx
```diff
@@ -0,0 +1,175 @@
...
+  disk, the console, or configuration::set."
+owner: "devrel"
+type: "how-to"
+---
+
+## Settings live in a worker, not in config.yaml
+
+iii splits configuration across two layers:
+
+- **`config.yaml`** declares _which workers run_. See
```
anthonyiscoding:
```suggestion
- **`config.yaml`** declares _which workers run_ and provides bootstrap values for configuration. See
```

docs/next/using-iii/configuration.mdx
```diff
@@ -0,0 +1,175 @@
...
+  [Engine configuration](./engine#engine-configuration).
+- **The `configuration` worker** owns _each worker's runtime settings_. It is a schema-validated,
+  reactive registry that ships enabled by default: every worker registers its settings schema under
+  its own entry (`iii-http`, `iii-state`, `iii-queue`, ...), and every change is validated against
+  that schema and applied to the running worker — no engine restart.
```
anthonyiscoding:
```suggestion
  that schema and applied to the running worker.
```

docs/next/using-iii/configuration.mdx
```diff
@@ -0,0 +1,175 @@
...
+the entry is the source of truth and the block is ignored. On the next boot the engine strips the
+dead block from `config.yaml` and leaves a one-line breadcrumb comment pointing at where the value
+now lives; the `- name:` line and everything else in the file are preserved as written.
+
+<Note>
+  This applies to workers that register a configuration schema — the built-in workers (`iii-http`,
+  `iii-state`, `iii-queue`, `iii-cron`, `iii-stream`, `iii-pubsub`, `iii-observability`) and
+  registry workers like `database` and `shell` all do. A worker that doesn't register one (for
+  example `iii-worker-manager`) keeps reading its `config.yaml` block directly.
+</Note>
```
anthonyiscoding, line 40:
```suggestion
<Note>
  This applies to workers that register a configuration schema. Some workers do not register
  a configuration schema. A worker that doesn't register one (for example `iii-worker-manager`) 
  keeps reading its `config.yaml` block directly.
</Note>
```

anthonyiscoding: Do we show what a configuration schema looks like somewhere/how to register one when creating a new worker?

docs/next/using-iii/configuration.mdx
```diff
@@ -0,0 +1,175 @@
...
+
+## One file per worker
+
+With the default `fs` adapter, every configuration entry is one YAML file named after the worker, in
+`./config` at your project root:
+
+```
+config/
+  iii-http.yaml
```
anthonyiscoding, line 49: Are we dropping all the iii- prefixes in this release? If so need to adjust these and anywhere else.

docs/next/using-iii/configuration.mdx
```diff
@@ -0,0 +1,175 @@
...
+  iii-queue.yaml
+  iii-state.yaml
+  ...
+```
+
+Each file carries the entry's identity plus its current value (the registered JSON Schema is not
+persisted — the worker re-registers it on every boot):
```
anthonyiscoding:
```suggestion
Each file carries the entry's identity plus its current value:
```

That other line is confusing and a bit off topic. However that information should go into a schema section.

docs/next/using-iii/configuration.mdx
```diff
@@ -0,0 +1,175 @@
...
+  [Engine configuration](./engine#engine-configuration).
+- **The `configuration` worker** owns _each worker's runtime settings_. It is a schema-validated,
+  reactive registry that ships enabled by default: every worker registers its settings schema under
+  its own entry (`iii-http`, `iii-state`, `iii-queue`, ...), and every change is validated against
+  that schema and applied to the running worker — no engine restart.
+
+The `config:` block under a worker in `config.yaml` is a **seed**: it is read once, on the first
+boot after the worker registers its schema, to create the worker's configuration entry. From then on
+the entry is the source of truth and the block is ignored. On the next boot the engine strips the
+dead block from `config.yaml` and leaves a one-line breadcrumb comment pointing at where the value
```
anthonyiscoding:
```suggestion
```

docs/next/using-iii/configuration.mdx
```diff
@@ -0,0 +1,175 @@
...
+now lives; the `- name:` line and everything else in the file are preserved as written.
```
anthonyiscoding:
```suggestion
```

docs/next/using-iii/configuration.mdx
```diff
@@ -0,0 +1,175 @@
...
+owner: "devrel"
+type: "how-to"
+---
+
+## Settings live in a worker, not in config.yaml
+
+iii splits configuration across two layers:
+
+- **`config.yaml`** declares _which workers run_. See
+  [Engine configuration](./engine#engine-configuration).
```
anthonyiscoding, line 16:
```suggestion
  [Engine configuration](./engine#engine-configuration). Your project's root `config.yaml` 
  file that iii reads at startup indicates which workers iii should start. It is the configuration 
  for iii. Each worker also has their own configuration controlled by the configuration 
  worker described below.
  
  <Note>
  In earlier versions of iii `config.yaml` contained all configuration information. That information is
  now in the configuration worker, and iii performs an automatic one time migration. Refer to the
  changelog for more information.
  </Note>
```

docs/next/using-iii/configuration.mdx
anthonyiscoding: Since this is a one time migration it should only exist in the migration docs, otherwise it's confusing for users who never experienced this behavior.

docs/next/using-iii/configuration.mdx
anthonyiscoding: Link to the changelog entry

docs/next/using-iii/configuration.mdx
```diff
@@ -0,0 +1,175 @@
...
+  host: 127.0.0.1
+```
+
+The directory is created on boot and watched from then on.
+
+<Note>
+  Before 0.21 the default directory was `./data/configuration`. When running on the default
+  location, the engine performs a one-time migration on boot and moves any entries it finds there
+  into `./config`. An explicit `directory:` override skips the migration.
+</Note>
```
anthonyiscoding: Could move this up and use in place of my other suggestion: https://github.com/iii-hq/iii/pull/1941/changes#r3545250713

docs/next/using-iii/configuration.mdx
```diff
@@ -0,0 +1,175 @@
...
+validated against the worker's registered schema and applied through exactly the same path as
+`configuration::set`. An edit that fails validation is rejected with a warning in the engine logs
+and the previous good value stays in effect, so a typo can't take a worker down.
+
+### Use the console
+
+The [console worker](https://workers.iii.dev/workers/console)'s **Configuration** page lists every
+registered entry and renders an editing form generated from its JSON Schema — typed fields,
+per-adapter variants, and validation before save. `${VAR:default}` templates are shown and saved
+verbatim, so env-driven values survive a console edit. Saving applies immediately.
```
anthonyiscoding: Confusing AI verbage here.
```suggestion
per-adapter variants, and validation before save. `${VAR:default}` templates are shown and saved
verbatim so environment variables can be recorded via the console. Saving applies immediately.
```

docs/next/using-iii/configuration.mdx
```diff
@@ -0,0 +1,175 @@
...
+schema's scalar type after expansion — `port: ${HTTP_PORT:3111}` validates as the integer `3111`,
+not a string. Pass `raw: true` to `configuration::get` to read the stored template form.
+
+## How changes apply
+
+Most settings hot-apply the moment they change: `iii-http` swaps CORS, timeout, and middleware
+without dropping the listener, and binds a new `host`/`port` before releasing the old one. A few
+fields are restart-tier — the change is recorded, logged, and takes effect at the next engine start
+(for example `iii-state`'s storage adapter). Each worker's page on
+[workers.iii.dev](https://workers.iii.dev/) documents its settings and how they apply.
```
anthonyiscoding, line 134: This contradicts earlier assertions that config hot reloads without a restart.

docs/next/using-iii/configuration.mdx
```diff
@@ -0,0 +1,175 @@
...
+
+## Changing where configuration lives
+
+The configuration worker is configured like any other worker — through its own block in
+`config.yaml`. To store the per-worker files somewhere else, set the `fs` adapter's `directory`:
```
anthonyiscoding: This is not true, now most workers are configured through configuration worker and one file per worker.

Is this a special case? Can configuration worker not store its own config?

docs/next/using-iii/configuration.mdx
```diff
@@ -0,0 +1,175 @@
...
+        name: fs
+        config:
+          directory: ./config
+```
+
+The `bridge` adapter replaces local files entirely and delegates every `configuration::*` call to a
+remote iii engine — useful when a fleet of engines should share one configuration source. An
+optional `ttl_seconds` cleans up entries for ephemeral workers after their last subscriber
+unregisters. See the [configuration worker docs](https://workers.iii.dev/workers/configuration) for
+both.
```
anthonyiscoding:
```suggestion
<Note>
The configuration worker also provides a `bridge` adapter for cases where multiple iii engines need to share the same configuration source. See the 
[configuration worker docs](https://workers.iii.dev/workers/configuration) for more details.
</Note>
```

docs/next/using-iii/configuration.mdx
```diff
@@ -0,0 +1,175 @@
...
+
+worker.registerTrigger({
+  type: "configuration",
+  function_id: "link::on-config-change",
+  config: {
+    configuration_id: "iii-http", // optional — omit to receive every id
+    event_types: ["configuration:updated"], // optional — registered | updated | deleted
+  },
+});
+```
```
anthonyiscoding: This gets into implementation details and belongs in creating-workers, this seems it would also be a good area to explain how to define and register a configuration schema.

docs/next/using-iii/engine.mdx
```diff
@@ -43,6 +43,13 @@ workers:
           file_path: ./data/state_store.db
 ```
 
+The `config:` block under a worker is a **first-boot seed**. A worker that registers a
+configuration schema reads it once to create its entry in the
+[configuration worker](./configuration); from then on its settings live in one file per worker
+under `./config/` — editable in real time from disk, the console, or `configuration::set` — and the
+engine strips the consumed block from `config.yaml`, leaving a breadcrumb comment in its place. See
+[Configuration](./configuration) for the full lifecycle.
+
```
anthonyiscoding, line 52: If this structure is no longer true it should be updated to be clear how the configuration worker works and that config.yaml is now only for telling iii which workers to start.

docs/next/using-iii/workers-registry.mdx
```diff
@@ -19,7 +19,8 @@ Workers can also be found on Docker and OCI-compatible registries.
 ## Adding a worker
 
 `iii worker add` accepts three sources. In every case the worker is added to your project's
-`config.yaml` and started automatically.
+`config.yaml` and started automatically. Its settings are seeded from `config.yaml` on first boot
+and managed through the [configuration worker](./configuration) afterwards.
```
anthonyiscoding: Earlier the docs mention that this is a one time migration, here it seems it happens on every run. Please adjust to the correct behavior and what `iii worker add` does differently (if anything)

docs/next/using-iii/workers.mdx
```diff
@@ -86,6 +86,10 @@ iii worker add ghcr.io/org/worker:tag # Pulls and adds a worker from a Docker or
 The worker is added to `config.yaml` and started automatically. To force a redownload of an existing
 worker, use `iii worker reinstall <name>` (equivalent to `add --force`).
 
+`iii worker add` also writes the worker's default `config:` block. That block seeds the worker's
+settings on its first boot; from then on, runtime settings are managed through the
+[configuration worker](./configuration), not `config.yaml`.
```
anthonyiscoding: Same comment here.

Also it seems that users don't have a say between whether they manage their own configuration locally or if they use the configuration worker?

anthonyiscoding (changes requested):

docs/next/using-iii/configuration.mdx
sergiofilhowz: no, not yet

docs/next/understanding-iii/engine.mdx
sergiofilhowz: Kept this sentence: the engine does still exit when `config.yaml` fails to parse or validate on reload (`reload.rs` propagates the error so `serve()` exits the process). Worker *settings* validation is a separate layer that never exits: an invalid change is rejected with a warning and the previous value stays in effect. Reworded the sentence to "An invalid `config.yaml`" and expanded the paragraph below to state the settings-validation behavior explicitly, so the two layers can't be conflated.

docs/next/using-iii/configuration.mdx
sergiofilhowz: No, the `iii-` prefixes stay in 0.21: the engine's configuration ids are unchanged (`iii-http`, `iii-state`, `iii-queue`, `iii-cron`, `iii-stream`, `iii-pubsub`, `iii-observability`). The only unprefixed one is the `configuration` worker itself. The docs match the code as-is.

docs/next/using-iii/configuration.mdx
sergiofilhowz: Not yet. Today registering a schema is a raw `configuration::register` call over the bus (there's no SDK sugar for it), and no creating-workers page covers it. Deferring that to a follow-up page under creating-workers (define a schema, register it, react to changes) rather than growing this operator-facing how-to.

docs/next/using-iii/configuration.mdx
sergiofilhowz: Agreed on placement: trimmed this section to a short pointer at the trigger and the worker docs, keeping the SDK snippet out of this operator-facing page. The full "define and register a configuration schema" guide belongs in creating-workers and is deferred to a follow-up (same follow-up as the schema question above), since registration today is a raw `configuration::register` call and deserves a proper page.

docs/next/using-iii/workers-registry.mdx
sergiofilhowz: Two different mechanisms were reading as one here. The one-time migration is the store's directory move (`./data/configuration` to `./config`), done once on boot for the default location. Seeding is per worker: each worker's `config.yaml` block is consumed once, on that worker's first boot, not on every run. `iii worker add` does nothing special beyond writing the default `config:` block that later serves as that seed. Reworded to "seeded from `config.yaml` once, on the worker's first boot" so it can't read as per-run.

docs/next/using-iii/workers.mdx
sergiofilhowz: On the choice question: correct, and by design. The configuration worker is mandatory (the engine injects it even when it's absent from `config.yaml`), so for any worker that registers a schema it is the settings surface; there's no per-worker opt-out. What stays in your hands is where the files are stored (the fs adapter's `directory`) and how you edit them: the per-worker files under `./config/` are plain YAML on disk and hand-editable, so managing configuration locally still works, edits just go through schema validation. Added a sentence here pointing that out, and applied the same "once, on its first boot" reword as the other thread.

docs/next/using-iii/configuration.mdx
anthonyiscoding: Alright let's get this added soon to the creating-workers section

anthonyiscoding (dismissed):

anthonyiscoding (approved):

PR #1942, opened 2026-07-08
opt-in ping_interval_secs keepalive on the listener

Atlas1225:
The worker-manager listener has no active liveness mechanism: the engine
only answers Ping frames, never initiates them, and a connection whose
peer died without FIN/RST (suspended device, NAT-dropped path) lingers in
the worker and trigger registries until a write to it exhausts TCP
retransmissions.

Add an opt-in `ping_interval_secs` to WorkerManagerConfig. When set, the
listener sends a WebSocket Ping frame to each connection every interval
and closes connections that produce no inbound frame (Pong included —
browsers and WS libraries auto-reply per RFC 6455) for two consecutive
intervals. Unset keeps the previous behavior unchanged.

## Summary by CodeRabbit

* **New Features**
  * Added optional WebSocket keepalive checks for worker connections, with configurable ping intervals.
  * Connections now stay alive during normal activity and are closed only after extended inactivity.

* **Bug Fixes**
  * Improved handling of unresponsive worker connections to prevent stale sessions from lingering.
  * Preserved existing behavior when keepalive checks are not configured.

* **Tests**
  * Added end-to-end coverage for silent, responsive, and disabled keepalive scenarios.

### DIFF
--- a/engine/src/engine/mod.rs
+++ b/engine/src/engine/mod.rs
@@ -1474,6 +1474,11 @@ impl Engine {
         tracing::debug!(peer = %peer, "Worker connected via WebSocket");
         let (mut ws_tx, mut ws_rx) = socket.split();
 
+        let ping_interval = config
+            .ping_interval_secs
+            .filter(|secs| *secs > 0)
+            .map(std::time::Duration::from_secs);
+
         let session =
             match rbac_session::handle_session(peer, Arc::new(self.clone()), config, uri, headers)
                 .await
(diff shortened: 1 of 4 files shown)


Atlas1225:
Do I need to do any change?

guibeira:
Hey @Atlas1225 thanks for your contribution! This really makes sense. Before I merge it, do you mind fixing the linter issue?

PR #1955, opened 2026-07-10
refactor: remove blog module and related assets, update workspace and…

sergiofilhowz:
… CI configurations to exclude blog paths


anthonyiscoding (changes requested):
- The navbar has lost some functionality and design. See the previous header below, is this intentional? If so can we get it aligned to the left?

Image

- Navbar doesn't narrow on scroll down like prior header, we should keep this behavior.

CleanShot 2026-07-13 at 10 42 13@2x

- Tech specs could be moved up to above the console section

sergiofilhowz:
hey @anthonyiscoding yea i asked to remove this because it was confusing to see links scrolling and other links to new pages when i was testing

sergiofilhowz:
but we can have this

image

anthonyiscoding:
> but we can have this
> 
> image

Looks good to me.

anthonyiscoding (dismissed):

PR #1959, opened 2026-07-14
fix(iii-worker,sdk): propagate managed worker name to SDKs via III_WORKER_NAME

sergiofilhowz:
## Why

The nightly [quickstart-validator broke](https://github.com/iii-hq/quickstart-validator/actions/runs/29297124845) on iii 0.21.5: `iii worker add ./workers/caller-worker` times out after 120s with **"not registered (engine has no worker under this name)"** while the worker VM is booted, connected, its function registered, and `iii trigger math::add_two_numbers a=10 b=20` returns `{"c":30}`.

## Root cause

#1949 (MOT-3931) made `iii worker add --wait` / `status` / `list` trust only the names reported by `engine::workers::list`. That's the right call — but nothing propagates the config.yaml entry name to the worker's SDK, so a worker whose code doesn't set `workerName` self-reports the SDK fallback `hostname:pid` (`(none):288` inside a libkrun VM) and can never match its config entry. The quickstart's TS caller-worker template sets no name → deterministic 120s timeout. The Python math-worker template happens to hardcode `worker_name="math-worker"` → step 3 passes, step 4 hangs.

Reproduced locally on macOS with the released 0.21.5 binaries; `engine::workers::list` showed the connected worker as `"(none):288"`.

## What

Plumb the managed identity end to end so worker code never has to repeat the config entry name:

- **iii-worker** injects `III_WORKER_NAME` (the config.yaml entry name) into every managed start path:
  - libkrun local/bundle (krun guest env, `build_local_env`)
  - OCI (`build_container_spec` env)
  - binary workers (spawned process env)

  The manifest `env:` block cannot override it (same treatment as `III_URL`/`III_ENGINE_URL`) — identity comes from the config entry.
- **Node / Python / Rust SDKs** default their reported worker name from `III_WORKER_NAME` when the caller doesn't set one. An explicit `workerName` / `worker_name` / `metadata.name` still wins; the `hostname:pid` fallback is unchanged when the env var is absent. (Flagging per AGENTS.md "Ask First": this changes default behavior on the published SDK surface, additively.)

## Verification

- Unit tests added in all four spots (env injection + non-overridability; SDK env default, explicit-option precedence, fallback, empty-env handling). `cargo test -p iii-worker/-p iii-sdk` targeted suites, Node vitest (12 passed), Python pytest (14 passed) all green.
- End-to-end on the local quickstart repro with a branch-built `iii-worker`: the VM boot cmdline carries `--env III_WORKER_NAME=caller-worker`, the guest workload sees it (`probe III_WORKER_NAME = caller-worker` from inside the VM), and with the SDK reading it the engine lists `caller-worker  connected` and `iii worker status caller-worker` reports `✓ ready / registered`.

## Interim / companion fixes

The published npm/PyPI SDKs (0.20.0) predate the env default, so until new SDKs ship the templates set `workerName` explicitly: iii-hq/templates#39 (unbreaks the quickstart + validator immediately) and iii-hq/quickstart-validator#7 (step-9 patch file).

## Summary by CodeRabbit

* **New Features**
  * Worker identity (`III_WORKER_NAME`) is now consistently set for managed, local, and binary workers, and included in worker environment/container specs.
  * Node.js, Python, and Rust SDKs use `III_WORKER_NAME` for default worker metadata when set and non-empty; otherwise fall back to `hostname:process ID`.
  * Explicitly configured worker names continue to take precedence over `III_WORKER_NAME`.
* **Bug Fixes**
  * Prevented project environment values from unintentionally overriding the worker identity, including ensuring the identity isn’t exported from local worker env settings.


guibeira (dismissed):

sergiofilhowz:
CI triage for the two failures on the previous runs:

1. **Worker Tests / Engine Coverage (compile error)** — my miss: `local_worker_integration.rs` still called the old 2-arg `build_local_env`. Fixed in 15cab66e (verified with `cargo check --workspace --all-targets`).
2. **Worker Tests (OCI) - linux: `kill_stale_worker_spares_recycled_pid`** — pre-existing flake in the #1949 test, not this PR's code: the test spawns its decoy and calls `kill_stale_worker` immediately, racing the exec transition where the kernel reports an empty cmdline. `pid_identity_matches` reads that as unjudgeable (`None`), which falls through to the kill path, so the decoy dies (the 0.51s test runtime matches the kill path's 500ms grace sleep exactly). Fixed in 5dc7edec by polling for the decoy's argv first, like the sibling tests already do. Product-level note for a follow-up discussion: the same `None` → kill fall-through means a *real* recycled PID caught mid-exec could still be killed — rare, but it's the exact hazard MOT-3931 guards against; a brief re-probe-on-`None` would close it.
3. **SDK Go Tests (`TestEnqueueRoutesThroughQueue`)** — unrelated to this PR and already failing on `main` (both runs since #1950 landed): the test engine no longer has an `engine::queue::enqueue` provider ("No queue provider is installed"). Needs a fix on the #1950 side; it will stay red here too until then.

guibeira (changes requested):
Please, replicate the changes for the Go SDK as well.

sergiofilhowz:
Follow-up on item 3 of the triage: took the `SDK Go Tests` fix into this PR after all (6e6e911c), since it blocks this pipeline and main's. `TestEnqueueRoutesThroughQueue` is deleted to match what #1950 did for the other SDKs — that PR removed the queue worker from `sdk/fixtures/config-test.yaml` (so the suite's engine has no enqueue provider) and dropped the Node/Python/Rust queue integration tests in favor of engine-free `TriggerAction` wire-shape contract tests; the Go suite was missed. The Go enqueue wire shape is already pinned by the in-package unit test in `protocol_test.go`, so no replacement test is needed. Verified locally: `gofmt`, `go vet` (plain + integration tag), unit tests with `-race`, and the full `go test -tags integration ./tests/...` suite against an engine started CI-style from `config-test.yaml` — all green.

guibeira (approved):

PR #1964, opened 2026-07-15
fix(browser-sdk): generate invocation UUIDs in insecure contexts

ytallo:
## Problem

The console (and any `iii-browser` consumer) breaks when served over plain `http://` from a non-localhost address — e.g. `http://<lan-ip>:3113`. Every `trigger()` call and `registerTrigger()` throws `TypeError: crypto.randomUUID is not a function`, because `crypto.randomUUID` is only defined in secure contexts (https / localhost).

The visible symptom: the console shows the **"install the harness worker"** setup screen and an empty model list even when the engine, harness, and providers are all healthy — the `engine::workers::list` probe fails client-side before it ever reaches the engine.

## Fix

Add a `randomUUID()` helper in `utils.ts` that uses `crypto.randomUUID` when present and otherwise builds an RFC 4122 v4 UUID from `crypto.getRandomValues` (available in insecure contexts, same CSPRNG). Both call sites in `iii.ts` (`registerTrigger`, invocation ids) now use it.

The fallback keeps the strict UUID format because the engine parses invocation ids as UUIDs and silently drops frames that don't parse (`json decode error: UUID parsing failed`).

## Test plan

- [x] `vitest run` — 54 tests pass
- [x] New `tests/utils.test.ts`: valid v4 UUID with and without `crypto.randomUUID` (simulated insecure context), uniqueness over 1000 ids
- [x] `npm run build` (tsdown) passes
- [x] Manually reproduced against a live engine: invocation with a well-formed v4 UUID from an insecure-origin page returns results; non-UUID ids are dropped by the engine

https://claude.ai/code/session_01Teg1GKop3sPUs9YvS3LqRm

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved compatibility when generating unique identifiers in browser environments where the preferred cryptographic UUID API is unavailable.
  * Trigger registrations and function invocations now consistently receive valid, unique UUIDs.

* **Tests**
  * Added coverage for UUID formatting, fallback generation, and uniqueness.

### DIFF
--- a/sdk/packages/node/iii-browser/src/utils.ts
+++ b/sdk/packages/node/iii-browser/src/utils.ts
@@ -80,3 +80,21 @@ const extractRefsRecursive = (
     extractRefsRecursive(value, path, refs)
   }
 }
+
+/**
+ * Generate an RFC 4122 v4 UUID.
+ *
+ * `crypto.randomUUID` is only defined in secure contexts (https / localhost),
+ * so pages served over plain http from a LAN address would throw on every
+ * invocation. Fall back to building the UUID from `crypto.getRandomValues`,
+ * which is available in insecure contexts and uses the same CSPRNG. The
+ * engine parses invocation ids as UUIDs, so the fallback must be a real one.
+ */
+export function randomUUID(): string {
+  if (typeof crypto.randomUUID === 'function') return crypto.randomUUID()
+  const bytes = crypto.getRandomValues(new Uint8Array(16))
+  bytes[6] = (bytes[6] & 0x0f) | 0x40
+  bytes[8] = (bytes[8] & 0x3f) | 0x80
+  const hex = Array.from(bytes, (b) => b.toString(16).padStart(2, '0')).join('')
+  return `${hex.slice(0, 8)}-${hex.slice(8, 12)}-${hex.slice(12, 16)}-${hex.slice(16, 20)}-${hex.slice(20)}`
+}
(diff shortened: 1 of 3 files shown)


sdk/packages/node/iii-browser/src/utils.ts
```diff
@@ -80,3 +80,21 @@ const extractRefsRecursive = (
...
+ * which is available in insecure contexts and uses the same CSPRNG. The
+ * engine parses invocation ids as UUIDs, so the fallback must be a real one.
+ */
+export function randomUUID(): string {
+  if (typeof crypto.randomUUID === 'function') return crypto.randomUUID()
+  const bytes = crypto.getRandomValues(new Uint8Array(16))
+  bytes[6] = (bytes[6] & 0x0f) | 0x40
+  bytes[8] = (bytes[8] & 0x3f) | 0x80
+  const hex = Array.from(bytes, (b) => b.toString(16).padStart(2, '0')).join('')
+  return `${hex.slice(0, 8)}-${hex.slice(8, 12)}-${hex.slice(12, 16)}-${hex.slice(16, 20)}-${hex.slice(20)}`
```
sergiofilhowz, line 99: probably better to just use uuid package

sergiofilhowz: it looks like uuid npm package uses crypto, let's just push this one, but make sure this doesn't have collisions

PR #1965, opened 2026-07-15
fix: SSRF bypass via IPv4-mapped IPv6 addresses in url_validator

Pranjal0410:
## What

`is_private_ip` in `engine/src/invocation/url_validator.rs` classified IPv4 and IPv6 addresses separately. An IPv4-mapped IPv6 address like `::ffff:169.254.169.254` was being run through the IPv6 checks only, none of which catch it, so it came back as "not private."

In practice that means an attacker could reach loopback, RFC1918 ranges, or the cloud metadata endpoint by writing the address in its `::ffff:a.b.c.d` form, sidestepping the SSRF guard entirely.

## Fix

`is_private_ip` now unwraps IPv4-mapped IPv6 addresses (`Ipv6Addr::to_ipv4_mapped`, stable since 1.63) to their underlying `Ipv4Addr` and classifies that instead. The IPv4 logic was pulled into a small `is_private_ipv4` helper so both branches use the same rules.

## Tests

Added to the existing `is_private_ip` unit tests:
- `::ffff:127.0.0.1` → private (loopback)
- `::ffff:169.254.169.254` → private (metadata IP, mapped form)
- `::ffff:10.0.0.1` → private (RFC1918, mapped form)
- `2606:4700:4700::1111` → still public (control case, unaffected by this change)

Ran locally on this branch:
```
cargo test -p iii --lib url_validator
```
All 3 new cases fail against the pre-fix code and pass after. Full `url_validator` suite (36 tests) is green post-fix.

## Not in this PR

There's a second, separate gap in this file: `validate()` resolves the hostname to vet its IP, then whatever makes the actual HTTP request resolves it again — nothing pins the request to the IP that was checked, so DNS rebinding (public IP at check time, private IP at request time) still gets through. Fixing that means threading the resolved IP through to the HTTP client (e.g. `reqwest`'s `resolve()` override) and making the resolver injectable for testing. Wanted to keep this PR scoped to the mapped-address bug; happy to follow up with that one separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved detection of private IPs when IPv4 addresses are provided in IPv6-mapped form.
  * IPv4-mapped loopback, cloud metadata, and RFC1918-mapped addresses are now consistently treated as private.
  * Public IPv6 addresses remain correctly allowed.

* **Tests**
  * Added coverage to ensure IPv4-mapped loopback is blocked when private IP blocking is enabled.
  * Expanded unit tests for private IP classification scenarios.

### DIFF
--- a/engine/src/invocation/url_validator.rs
+++ b/engine/src/invocation/url_validator.rs
@@ -347,6 +347,23 @@ mod tests {
     }
 
     #[tokio::test]
+    async fn test_url_validator_blocks_ipv4_mapped_ipv6_loopback() {
+        let validator = UrlValidator::new(UrlValidatorConfig {
+            allowlist: vec!["*".to_string()],
+            block_private_ips: true,
+            require_https: false,
+        })
+        .unwrap();
+
+        let result = validator.validate("http://[::ffff:127.0.0.1]/").await;
+        assert!(result.is_err());
+        assert!(matches!(
+            result.unwrap_err(),
+            SecurityError::PrivateIpBlocked
+        ));
+    }
+
+    #[tokio::test]
     async fn test_url_validator_blocks_private_rfc1918() {
         let validator = UrlValidator::new(UrlValidatorConfig {
             allowlist: vec!["*".to_string()],
--- a/engine/src/invocation/url_validator.rs
+++ b/engine/src/invocation/url_validator.rs
@@ -110,15 +110,16 @@ impl UrlValidator {
 
 fn is_private_ip(ip: &IpAddr) -> bool {
     match ip {
-        IpAddr::V4(ipv4) => {
-            ipv4.is_loopback()
-                || ipv4.is_private()
-                || ipv4.is_link_local()
-                || ipv4.is_broadcast()
-                || ipv4.is_documentation()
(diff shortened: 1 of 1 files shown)


guibeira:
Hey @Pranjal0410 , thanks for your contribution!

The new tests exercise the private is_private_ip helper directly, but not the public validation path. Could we add a regression test that calls:

  validator
      .validate("http://[::ffff:127.0.0.1]/")
      .await

  and asserts Err(SecurityError::PrivateIpBlocked)?

  This would cover the complete Url::parse → host_str() → DNS lookup → IP classification flow and ensure the SSRF bypass is blocked through the API actually used by callers.

Pranjal0410:
Good call -  I added \`test_url_validator_blocks_ipv4_mapped_ipv6_loopback\`, which calls \`validator.validate("http://[::ffff:127.0.0.1]/")\` and asserts \`Err(SecurityError::PrivateIpBlocked)\`. Confirmed it fails against the pre-fix \`is_private_ip\` (assertion failure on \`result.is_err()\`) and passes with the fix in place. Full \`url_validator\` suite is at 37/37.

guibeira (approved):

PR #1974, opened 2026-07-16
fix(engine,iii-worker): pass engine's actual WS address to spawned builtin daemons (MOT-3970)

andersonleal:
Fixes [MOT-3970](https://linear.app/motia/issue/MOT-3970/) (found smoke-testing MOT-3857).

## Problem

The engine spawns builtin daemons via the `KNOWN_EXTERNAL` table (`iii-sandbox` → `iii-worker sandbox-daemon`, `iii-worker-ops` → `iii-worker worker-manager-daemon`) with argv = subcommand + optional `--config` only — no engine URL. Both daemons' `--engine` args hardcode `ws://127.0.0.1:49134` with no env fallback.

An engine running its worker WS listener on any non-default port therefore spawned daemons that connected to **whatever sat on 49134** — a different engine instance (cross-engine registration clobbering) or nothing. Two engines on one host could not each get their own daemon.

## Fix

Env var rather than argv, because the same `external.rs` spawn path also runs legacy `iii.toml` externals (arbitrary binaries that accept no `--engine` flag), and `III_ENGINE_URL` is already the documented engine-set contract (`worker_manifest.rs`, `build_local_env`, libkrun path):

- **engine**: `ExternalWorker` now carries the effective `iii-worker-manager` port (`engine.worker_manager_port()`, same source as the already-fixed `registry_worker` sibling path) and exports `III_ENGINE_URL`/`III_URL` = `ws://127.0.0.1:<port>` to every spawned external worker, next to `III_ENGINE_PID`.
- **iii-worker**: `WorkerManagerDaemonArgs.engine` and `SandboxDaemonArgs.engine` gain `env = "III_ENGINE_URL"` (clap `env` feature added). Precedence: explicit `--engine` > env > default, so existing integration tests and manual invocations are untouched.

## Tests

- `argv_probe.sh` spawn-contract fixture now reports `engine_url=`; the engine spawn test constructs the worker with a distinctive port (50123) and asserts the child received `ws://127.0.0.1:50123` (fails with `engine_url=unset` without the fix).
- New serial parse test in `app.rs`: both daemons pick up `III_ENGINE_URL`, explicit flag beats env, default without either.

## Verification

- `cargo test -p iii external` (52), `--test external_known_sandbox` (2), `worker_manager_port` (3), `cargo test -p iii-worker cli::app` (4) — all green; `cargo fmt --all --check` clean; no new clippy warnings in touched files.
- Live smoke: engine with `iii-worker-manager: {port: 50123}` + `iii-worker-ops` → daemon spawned with bare argv held ESTABLISHED connections to `127.0.0.1:50123`, zero connections to 49134, and registered its `worker::*` functions on the spawning engine.

## Summary by CodeRabbit

* **New Features**
  * Worker daemon commands can now read the engine URL from the `III_ENGINE_URL` environment variable.
  * Explicit command-line engine settings continue to take precedence over environment variables.
  * External workers now connect using the configured worker manager port, including non-default ports.

* **Bug Fixes**
  * Fixed external worker processes receiving an outdated or hardcoded engine connection URL.

* **Tests**
  * Added coverage for environment-variable parsing, precedence, default values, and custom worker manager ports.

### DIFF
--- a/engine/src/workers/external.rs
+++ b/engine/src/workers/external.rs
@@ -180,6 +180,11 @@ pub struct ExternalWorker {
     name: String,
     binary_path: PathBuf,
     extra_args: Vec<String>,
+    /// Effective `iii-worker-manager` port, from `engine.worker_manager_port()`.
+    /// Exported to the child as III_ENGINE_URL/III_URL so builtin daemons
+    /// (sandbox-daemon, worker-manager-daemon) connect back to THIS engine
+    /// instead of their hardcoded ws://127.0.0.1:49134 default (MOT-3970).
+    worker_manager_port: u16,
     config: Option<Value>,
     child: Arc<Mutex<Option<Child>>>,
     config_file: Arc<Mutex<Option<PathBuf>>>,
@@ -193,7 +198,7 @@ pub struct ExternalWorker {
 }
 
 impl ExternalWorker {
-    pub fn new(info: ExternalWorkerInfo, config: Option<Value>) -> Self {
+    pub fn new(info: ExternalWorkerInfo, config: Option<Value>, worker_manager_port: u16) -> Self {
         let name = info.name.clone();
         let display_name = Box::leak(format!("ExternalWorker({})", &name).into_boxed_str());
         Self {
@@ -201,6 +206,7 @@ impl ExternalWorker {
             name,
             binary_path: info.binary_path,
             extra_args: info.extra_args,
+            worker_manager_port,
             config,
             child: Arc::new(Mutex::new(None)),
             config_file: Arc::new(Mutex::new(None)),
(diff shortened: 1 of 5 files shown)


engine/src/workers/external.rs
```diff
@@ -180,6 +180,11 @@ pub struct ExternalWorker {
     name: String,
     binary_path: PathBuf,
     extra_args: Vec<String>,
+    /// Effective `iii-worker-manager` port, from `engine.worker_manager_port()`.
+    /// Exported to the child as III_ENGINE_URL/III_URL so builtin daemons
+    /// (sandbox-daemon, worker-manager-daemon) connect back to THIS engine
+    /// instead of their hardcoded ws://127.0.0.1:49134 default (MOT-3970).
```
guibeira, line 186: Can remove the ticket mention; also, should we include the host as well?

guibeira (approved):

PR #1984, opened 2026-07-16
feat: namespace as a runtime routing dimension (engine + SDKs)

guibeira:
## Summary

Adds **namespace** as a runtime routing dimension across the engine and all four SDKs. A function keeps its id verbatim (`state::get` is `state::get` in any namespace); the namespace is a separate dimension carried alongside it, declared **per connection**. Two compose files that both declare a `state` worker can now coexist on one engine.

## What it does

- **Function id stays verbatim.** Namespace is a separate routing dimension, never folded into the name.
- **Strict routing, no best-fit.** `trigger("state::get")` resolves **only** in `default`; `trigger("state::get", namespace="analytics")` resolves **only** in `analytics`. The caller's own namespace never influences resolution. A miss returns `function_not_found` listing the namespaces where the id *does* exist.
- **Collision = rejection.** A second worker claiming a live `(namespace, name)` is rejected with a structured `RegistrationRejected { code: "WORKER_NAMESPACE_CONFLICT", namespace, worker_name, owner_worker_id }` and its connection is closed. A function-id collision from another live worker returns `FUNCTION_NAMESPACE_CONFLICT` **without** closing the connection.
- **Backward compatible.** Workers that send no namespace land in `default` and behave exactly as before. Every new wire field is optional.
- **All seven trigger types respect namespace.** state, cron, configuration, stream, http, queue, pubsub — target *and* condition functions resolve in the namespace the trigger declares: the registering worker's own namespace for the typed helpers, or `default` when the `RegisterTrigger` message carries none. The engine's own `functions-available` lifecycle path is namespaced too.
- **Four SDKs mirror the surface.** `namespace` init option (resolved explicit-option > `III_NAMESPACE` env > default), trigger targeting, and fatal `RegistrationRejected` handling (no reconnect) in Node, Rust, Python, and Go (`WithNamespace` / `FatalError`).

## How it works (load-bearing details)

- **Namespace arrives per connection.** It is not a protocol message — it rides on the `engine::workers::register` call. Because SDKs historically flush `RegisterFunction` *before* that call, the engine **buffers** a connection's registrations until its namespace is known (via `workers::register` or a 5s grace), then drains them in arrival order. This is why old SDKs keep working unchanged.
- **Registries keyed by `(namespace, name)`.** `FunctionsRegistry`, `ServicesRegistry`, and the function-ownership maps are all re-keyed; lookups, removals, RBAC, and introspection all thread the connection's real namespace.
- **Fast-restart safe.** Rejecting *any* duplicate `(ns, name)` (including the default `hostname:pid`) risks rejecting a worker restarting against its own not-yet-cleaned connection. The liveness check treats a tearing-down connection (`namespace_states == Aborted`, set as the first step of `cleanup_worker`) as not-live, so a restart reclaims its name immediately.

## Testing

Every behavior is covered by a test that goes **red** when the implementing line is neutered (delete-the-line discipline), driving the real WS / registration / trigger paths rather than hand-built objects. Engine e2e suites: `namespace_routing_e2e`, `namespace_conflict_e2e`, `namespace_introspection_e2e`, `namespace_trigger_e2e`, `worker_namespace_ws_e2e`, plus a real-broker rabbitmq namespaced-dispatch test. Each SDK extends its existing `III_WORKER_NAME` metadata test. Lib suite green; the namespace work is mutation-verified across the routing, rejection, buffering, introspection, and trigger paths.

## Rebase note

Rebased onto current `main`, which includes **recoverable triggers (#1962)**. That feature parks a `Trigger` in a pending map and re-activates it on worker restart; the namespace rides through by construction (activation inserts the full trigger into the live registry). Pinned with `a_recovered_pending_trigger_keeps_its_namespace` — a recovered trigger still resolves in the namespace it was registered in.

## Out of scope / follow-ups

- **Docs.** `architecture/PROTOCOL.md` / `SDK.md` (referenced by repo rules) do not exist in this repo. Protocol/SDK reference docs live in `docs/next/reference/engine-protocol.mdx` (+ `.skill.md`, CI-validated via `iii-skill-render`) and `docs/*/api-reference/sdk-*.mdx`. Namespace additions to those, plus a Divio how-to for duplicate function names across namespaces, land in a **separate docs PR**.
- **Test isolation bug (pre-existing, unrelated).** The config-reload e2e tests watch `<cwd>/engine/config`, which the suite regenerates mid-run; under a large/near-full `target/` this causes timing-dependent failures in the working checkout (passes cleanly in a fresh worktree). Worth a follow-up to isolate the reload tests' config dir. Related to the 0.21.x `./config` default.

## ⚠️ Breaking change — RabbitMQ durable subscriber queue names

Durable queue identity is now namespace-qualified and **collision-free** (injective encoding: `@`/`\` are escaped per segment, joined by a single unescaped `@`). For the **builtin** broker this is transparent — legacy `{topic}::{fid}` subscriber queues are drained into their namespace on upgrade, and a legacy name with no `@` encodes byte-identically to before.

For the **RabbitMQ** adapter there is **no auto-migration**: durable queues from a pre-namespace deploy keep their old names and will no longer be addressed after upgrade. Operators must drain/redeclare affected queues (an operator-gated migration tool is a tracked follow-up). This is accepted and intentional — surfaced here so it is not a silent runtime change.

## Review fixes

Every external review round was addressed in full; each finding was reproduced before fixing and ships a mutation-proven test (red when the fix line is neutered). Round 2 (Plan C):

- `4980f7830` — browser SDK: namespace support + keep the fatal state after a rejection (state no longer overwritten failed→disconnected; pending invocations rejected). Final-state assertion, not transient.
- `67d9cb211` — expose `namespace` on `MiddlewareFunctionInput` across all three SDKs so typed middleware can re-target; revert = **compile error** (Rust E0609 / Node TS2339 / Python AttributeError).
- `938bf2f27` — scope `workers::info` trigger types by resolved connection id and report the worker namespace.
- `0607da239` — unify the two ownership maps into one atomic `(owner, kind)` entry, closing the non-atomic claim/release window.
- `57e9afc49` — make durable queue identity injective (see breaking-change note above).

Round 3 (three findings, two of them regressions the round-2 changes introduced):

- `87a7ee1f` — **restore** `namespace` on the Python `TriggerRequest`. `67d9cb211` dropped it while editing the neighbouring type; pydantic ignores the unknown kwarg, so a typed `TriggerRequest(namespace=...)` silently routed to `default` (the dict path still worked, which is why the existing dict test missed it). Test now uses a typed instance.
- `157385227` — expose `namespace` on the **browser** `MiddlewareFunctionInput`; the round-2 fix covered node/rust/python but not `iii-browser`. Guarded by a `tsc --noEmit` type-check.
- `40d60511` — `workers::info` still spliced a connectionless (`worker_id=None`) internal trigger type into a WebSocket worker that shared its name in a non-default namespace. A resolved WS worker now claims only types pinned to its own connection; name-based attribution stays for the in-process runtime worker. New e2e goes red under the previous guard while the round-2 test stays green.

Round 4 (Go SDK parity + two engine/SDK edge cases):

- `8f8122c87` — complete Go SDK namespace support: `WithNamespace` (+ `III_NAMESPACE` env fallback), `namespace` on the worker-register payload, `TriggerRequest.Namespace` threaded onto the invoke wire, and the `registrationrejected` message type + terminal handling (`FatalError`, stop reconnect, fail pending) so a Go worker no longer loops rejection/reconnect under the default `MaxRetries: -1`.
- `b9cf5650b` — a late `engine::workers::register` (arriving after the grace timer fixed the connection to `default`) declaring a different namespace no longer splits state. The lease, metadata, and drained functions all follow the already-fixed namespace; a paused-time test drives grace-expiry then a real announce.
- `4a1a89c16` — the Python fatal state (`failed`) is no longer regressed to `disconnected` by the socket close the rejection itself triggers.

## Draft status

Opened as draft. The engine + SDK code is complete and mutation-reviewed; independent review of the largest engine task (6.5) and the SDKs passed. One review-completeness debt: Task 4's five modified tests were verified by orchestrator-run mutations but not each re-run individually.

## Summary by CodeRabbit

* **New Features**
  * Added namespace-aware worker, function, trigger, invocation, middleware, and queue routing.
  * Added namespace configuration through SDK options and `III_NAMESPACE`.
  * Added namespace details to discovery and introspection results.
  * Added registration conflict reporting with structured errors and fatal connection handling.
* **Bug Fixes**
  * Prevented cross-namespace routing, ownership, middleware, and durable queue collisions.
  * Preserved namespaces during trigger fan-out, retries, reconnects, and queued execution.
* **Documentation**
  * Added namespace usage, configuration, protocol, queue migration, and upgrade guidance.


docs/next/creating-workers/queues.mdx
```diff
@@ -238,6 +238,15 @@ The `durable:subscriber` trigger's `config` controls how each subscriber consume
 A failed delivery retries with exponential backoff (1 second, then 2 seconds) for up to 3 attempts,
 then the message dead-letters.
 
+Each subscriber's durable queue is scoped by the subscribing worker's namespace, so two subscribers
+of the same topic and function id in different namespaces are two queues that each receive every
+published event, rather than two competing consumers of one queue.
```
anthonyiscoding, line 243: I'm not certain I understand this, two queues in separate namespaces will receive each other's published events? That sounds incorrect

docs/next/creating-workers/worker-manifest.mdx
```diff
@@ -147,6 +147,14 @@ env:
   MY_API_KEY: replace-me
 ```
 
+`III_NAMESPACE` is the supported way to put a managed worker in a namespace: the SDK reads it when no
```
anthonyiscoding: It is very unlikely that a user would want to configure a namespace for the shipped worker itself. This is primarily a compose file concern and should be made clear to users.

docs/next/creating-workers/worker-manifest.mdx
```diff
@@ -147,6 +147,14 @@ env:
...
+explicit `namespace` option is passed. There is no per-worker `namespace:` key in `config.yaml`.
```
anthonyiscoding: This would be compose files, not config.yaml

docs/next/creating-workers/workers.mdx
```diff
@@ -126,6 +129,16 @@ process can be deployed anywhere reachable on the network.
   </Tab>
 </Tabs>
 
+`namespace` scopes everything this worker registers, so an identically-named worker or function id
+can coexist in another namespace. The SDK resolves it from the explicit option, then the
+`III_NAMESPACE` environment variable, and falls back to the engine's `default` namespace when neither
```
anthonyiscoding:
```suggestion
`III_NAMESPACE` environment variable, and falls back to the `default` namespace when neither
```
The engine has no notion of a namespace and should be "dumb" in this regard.

docs/next/reference/engine-protocol.mdx
```diff
@@ -94,14 +100,21 @@ functions (`HttpInvocationRef`); leave it `null` for in-process handlers.
...
 }
 ```
 
 `config` is the per-trigger-type configuration; the shape is defined by whatever worker advertised
 that `trigger_type` (e.g. `http` for `http` triggers). The engine responds with a
 `TriggerRegistrationResult` carrying an optional `error: ErrorBody`.
 
+`namespace` is optional and names the namespace `function_id` resolves in when the trigger fires. It
+is independent of the registering connection's own namespace: an absent `namespace` means `default`,
+not "the namespace this worker registered under". The typed SDK helpers that pair a function with its
```
anthonyiscoding: According to these docs this is a separate namespace from the one defined in compose/worker init/III_NAMESPACE? Why would it be separate?

docs/next/reference/engine-protocol.mdx
```diff
@@ -182,6 +200,37 @@ Failure:
...
+  "worker_name": "state",
+  "owner_worker_id": "3f9c1a2e-…"
+}
+```
+
+Pushed when a registration collides with a worker that is already live in `namespace`.
+`owner_worker_id` identifies the worker that holds the contested identity. `code` decides the
+severity:
+
+| `code`                        | `worker_name` carries | Connection | Severity                                                                     |
```
anthonyiscoding: worker_name has two different types of information in it potentially? This is an anti-pattern if true.

docs/next/reference/engine-protocol.mdx
```diff
@@ -182,6 +200,37 @@ Failure:
...
+| ----------------------------- | --------------------- | ---------- | ---------------------------------------------------------------------------- |
+| `WORKER_NAMESPACE_CONFLICT`   | the worker name       | closed by the engine | Fatal. The SDK stops the worker and does not reconnect.            |
+| `FUNCTION_NAMESPACE_CONFLICT` | the function id       | stays open | Non-fatal. Only that one registration is refused; the worker keeps serving its other functions. |
```
anthonyiscoding: This can lead to unexpected behavior where a worker succeeds but some of its functions don't register. Then you can have someone expecting that worker to do something on an invocation like:

```js
iii.trigger("sendEmail", namespace: "1") // conflict, fails, another worker already registers "sendEmail"
iii.registerTrigger("emailSent", namespace: "1") // the worker that is live with sendEmail doesn't emit an emailSent event as expected
```

docs/next/reference/engine-protocol.mdx
```diff
@@ -200,27 +249,72 @@ without an `invocation_id` and never expects a reply. For `Enqueue` the engine h
 to the queue worker, which persists it and re-invokes the target function on a subscriber according
 to the queue's retry policy.
 
+## Namespaces
+
+A namespace is a routing dimension carried alongside a function id, never folded into it: `state::get`
```
anthonyiscoding:
```suggestion
A namespace is a routing dimension carried alongside a function id, never folded into it the function name. For example `state::get`
```

docs/next/reference/engine-protocol.mdx
```diff
@@ -200,27 +249,72 @@ without an `invocation_id` and never expects a reply. For `Enqueue` the engine h
...
+is `state::get` in every namespace. Registries are keyed by `(namespace, function_id)` and
+`(namespace, worker_name)`, so the same id or worker name may appear once per namespace.
+
+A worker declares its namespace on the `engine::workers::register` call
+([`RegisterWorkerInput.namespace`](#engine-discovery-functions)), not on a protocol frame. A
```
anthonyiscoding: We should never document what a function doesn't do.


```suggestion
A worker declares its namespace on the `engine::workers::register` call
([`RegisterWorkerInput.namespace`](#engine-discovery-functions)). A
```

docs/next/reference/engine-protocol.mdx
```diff
@@ -200,27 +249,72 @@ without an `invocation_id` and never expects a reply. For `Enqueue` the engine h
...
+| `"orders"`                 | `orders` only         |
+
+A miss returns `function_not_found` naming the namespaces where the id does exist.
+
+Introspection resolution is deliberately looser, so that an engine whose every worker is namespaced
+can still answer questions about itself. For `engine::functions::info` and `engine::workers::info`
+with no explicit `namespace`: a `default` entry wins; otherwise an id or name unique to one
+non-default namespace resolves; an id or name present in several non-default namespaces at once is
+reported as an ambiguity naming the candidates, never resolved by guessing. Passing an explicit
+`namespace` restores strict resolution.
```
anthonyiscoding, line 332: This seems unhelpful. You'd want an agent or person querying without a namespace to be able to discover what is available in all namespaces not just `default`. Or have some other way of querying all namespaces. Otherwise you'll lead to a preference for the `default` namespace when it's inappropriate.

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
+---
+title: "Namespaces"
+description:
+  "Why iii routes on a namespace dimension carried alongside the function id, instead of folding it
+  into the name."
```
anthonyiscoding: Users don't need to be told this. This is another example of documenting what the software doesn't do.

```suggestion
  "iii routes functions based on their namespace allowing for multiple tenants, agents, projects, and teams to exist in the same system without conflicts."
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+owner: "devrel"
+type: "explanation"
+---
+
+## What a namespace is
+
+A namespace is a routing dimension the engine carries alongside a function id and a worker name. It
+is not part of either. `state::get` is `state::get` whether it lives in `default`, in `orders`, or in
+`analytics`; what changes is the second coordinate the engine files it under.
```
anthonyiscoding:
```suggestion
`analytics`.
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+  "Why iii routes on a namespace dimension carried alongside the function id, instead of folding it
+  into the name."
+owner: "devrel"
+type: "explanation"
+---
+
+## What a namespace is
+
+A namespace is a routing dimension the engine carries alongside a function id and a worker name. It
+is not part of either. `state::get` is `state::get` whether it lives in `default`, in `orders`, or in
```
anthonyiscoding:
```suggestion
is not part of either. `state::get` is `state::get` whether it is in `default`, in `orders`, or in
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+`analytics`; what changes is the second coordinate the engine files it under.
+
+The registries are keyed by pairs rather than by strings: `(namespace, function_id)` for functions,
+`(namespace, worker_name)` for worker identity. A namespace is declared once per connection, and
```
anthonyiscoding:
```suggestion
`(namespace, worker_name)` for worker identity. A namespace is declared once per worker connection, and
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+every registration that connection makes inherits it.
+
+## The problem it solves
+
+Before namespaces, a running engine had one flat registry. A function id was globally unique, and so
+was a worker name. That is a comfortable model right up to the moment two things legitimately want
+the same name.
```
anthonyiscoding: Similar to the "don't document what we don't do rule" outside of migration docs and changelogs it's generally not recommended to document what we no longer do.
```suggestion
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+
+The concrete case is two projects on one engine, each declaring a `state` worker. Both are correct;
+neither is willing to rename. Without a second coordinate the engine has to pick one, and every
```
anthonyiscoding:
```suggestion
neither is able to rename. Without a second coordinate the engine has to pick one, and every
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+`(namespace, worker_name)` for worker identity. A namespace is declared once per connection, and
+every registration that connection makes inherits it.
+
+## The problem it solves
+
+Before namespaces, a running engine had one flat registry. A function id was globally unique, and so
+was a worker name. That is a comfortable model right up to the moment two things legitimately want
+the same name.
+
+The concrete case is two projects on one engine, each declaring a `state` worker. Both are correct;
```
anthonyiscoding: I'd suggest rewriting this section around what it enables which is the ability to ship workers that can be used by more than one tenant on the same system.

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+out. A worker's published contract would change depending on where someone chose to run it.
+
+Keeping the id verbatim means the namespace stays a property of the deployment rather than of the
+contract. A worker author never writes a namespace into their function ids; an operator decides,
+at deploy time, which namespace the worker registers into. The two concerns stay separable, which is
+the whole point of having a second coordinate rather than a longer first one.
+
+## Why routing is strict
+
+Resolution never searches. A call with no namespace resolves in `default` and nowhere else; a call
```
anthonyiscoding:
```suggestion
A call with no namespace resolves to `default`; a call
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+without saying so reaches `default`, not its own neighbour.
+
+The tempting alternative is a best-fit search: try the caller's namespace, fall back to `default`,
+fall back to whatever exists. It reads as helpful and behaves as a hazard. Under best-fit, adding a
+worker to a namespace can silently re-target calls that were resolving somewhere else, and a typo in
+a function id can resolve to a real function in another tenant instead of failing. The failure is
+invisible, and it is a wrong answer rather than a missing one.
+
+Strictness makes the failure loud instead. A miss returns `function_not_found` and names the
+namespaces where the id does exist, so the caller learns both that the call was wrong and how to fix
```
anthonyiscoding:
```suggestion
any namespaces where the id does exist, so the caller learns both that the call was wrong and how to fix
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+not influence where its calls land: a worker registered in `analytics` that calls `state::get`
+without saying so reaches `default`, not its own neighbour.
+
+The tempting alternative is a best-fit search: try the caller's namespace, fall back to `default`,
+fall back to whatever exists. It reads as helpful and behaves as a hazard. Under best-fit, adding a
+worker to a namespace can silently re-target calls that were resolving somewhere else, and a typo in
+a function id can resolve to a real function in another tenant instead of failing. The failure is
+invisible, and it is a wrong answer rather than a missing one.
+
+Strictness makes the failure loud instead. A miss returns `function_not_found` and names the
```
anthonyiscoding:
```suggestion
Strictness makes the failure loud instead. A miss returns `function_not_found` and reports the
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+it, without the engine ever guessing on their behalf.
+
+Introspection is the one place where a looser rule earns its keep. Asking "what is this engine
+running?" with no namespace in hand is a reasonable question, and an engine whose every worker is
+namespaced would otherwise be unable to describe itself. So `engine::functions::info` and
+`engine::workers::info` prefer `default`, then resolve a name that is unique to one non-default
+namespace, and report an ambiguity when a name lives in several at once. They stop short of guessing
+for exactly the same reason invocation does: naming the candidates is useful, picking one silently is
+not. Passing an explicit namespace restores strict resolution.
+
```
anthonyiscoding: I don't really understand this paragraph. Can you clarify it?

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+fall back to whatever exists. It reads as helpful and behaves as a hazard. Under best-fit, adding a
+worker to a namespace can silently re-target calls that were resolving somewhere else, and a typo in
+a function id can resolve to a real function in another tenant instead of failing. The failure is
+invisible, and it is a wrong answer rather than a missing one.
+
+Strictness makes the failure loud instead. A miss returns `function_not_found` and names the
+namespaces where the id does exist, so the caller learns both that the call was wrong and how to fix
+it, without the engine ever guessing on their behalf.
+
+Introspection is the one place where a looser rule earns its keep. Asking "what is this engine
```
anthonyiscoding:
```suggestion
Asking "what is this engine
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+running?" with no namespace in hand is a reasonable question, and an engine whose every worker is
+namespaced would otherwise be unable to describe itself. So `engine::functions::info` and
+`engine::workers::info` prefer `default`, then resolve a name that is unique to one non-default
+namespace, and report an ambiguity when a name lives in several at once. They stop short of guessing
+for exactly the same reason invocation does: naming the candidates is useful, picking one silently is
+not. Passing an explicit namespace restores strict resolution.
+
+## Why a collision is a rejection
```
anthonyiscoding:
```suggestion
## Why a namespace collision is a rejection
```

docs/next/using-iii/configuration.mdx
```diff
@@ -125,6 +125,26 @@ without rewriting the stored value. A field that consists of a single placeholde
 schema's scalar type after expansion: `port: ${HTTP_PORT:3111}` validates as the integer `3111`, not
 a string. Pass `raw: true` to `configuration::get` to read the stored template form.
 
+## Registration namespace grace
```
anthonyiscoding: There is no `creating-workers/namespaces.mdx` and there really should be one that covers how to use namespaces and also shows the recommended way of selecting a namespace: grab it from the III_NAMESPACE environment variable that's declared at runtime.

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+
+When a worker claims a name another live worker already holds in the same namespace, the engine
+refuses the registration rather than accepting it.
+
+The alternative, last-write-wins, is how a flat registry degrades in practice: a second copy of a
+worker starts, quietly takes ownership of the name, and calls begin landing somewhere their author
+did not intend. Nothing errors. The symptom surfaces much later, as a behaviour change with no
+obvious cause.
+
```
anthonyiscoding:
```suggestion
```

docs/next/reference/engine-protocol.mdx
```diff
@@ -200,27 +249,72 @@ without an `invocation_id` and never expects a reply. For `Enqueue` the engine h
...
+
+### Reserved ids
+
+`engine::*` is reserved for engine and builtin infrastructure, which lives in `default`. A worker
```
anthonyiscoding: Is this going to impact other workers that expose engine::* functions? Namely the queue worker.

https://workers.iii.dev/workers/queue?tab=api#fn-engine::queue::dlq_messages several engine::* functions in it.

Also what happens when these queue engine functions are namespaced?

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+A contested **worker name** is an identity conflict: the worker cannot function as itself, so the
+engine closes the connection and the SDK stops rather than reconnecting into a rejection loop. A
+contested **function id** is narrower: only that one registration is refused, the connection stays
+open, and the worker keeps serving everything else it registered. Treating both as fatal would take
+a healthy worker offline over one duplicated function.
+
+There is a case that looks like a collision and is not: a worker restarting faster than the engine
+cleans up its previous connection, contending with itself. The engine treats a connection that has
+begun tearing down as no longer live, so a restart reclaims its own name immediately. Without that
+carve-out the default `hostname:pid` naming would make fast restarts fail against their own ghost.
```
anthonyiscoding:
```suggestion
<Note>
There is a case that looks like a collision and is not: a worker restarting faster than the engine
cleans up its previous connection, contending with itself. The engine treats a connection that has
begun tearing down as no longer live, so a restart reclaims its own name immediately. Without that
carve-out the default `hostname:pid` naming would make fast restarts fail against their own ghost.
</Note>
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+
+## Why it arrives per connection
+
+A namespace is declared on the `engine::workers::register` call rather than on each registration
```
anthonyiscoding:
```suggestion
A namespace is declared on the `engine::workers::register` call
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+frame, because it describes the worker, not the individual function. One worker is one identity in
```
anthonyiscoding:
```suggestion
because it describes the worker, not the individual function. One worker is one identity in
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+one namespace; letting individual registrations disagree would create a worker that is partly in two
+places.
```
anthonyiscoding:
```suggestion
or more places.
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+
+That choice has a cost, and the cost is ordering. SDKs have always flushed their `RegisterFunction`
+frames as soon as the socket opens, before announcing worker metadata. If the engine filed those
+registrations as they arrived it would file them before knowing the namespace. So it buffers a
+connection's registrations until the namespace is known, then drains them in arrival order.
```
anthonyiscoding:
```suggestion
SDKs flush their `RegisterFunction`
frames as soon as the socket opens, before announcing worker metadata. If the engine filed those
registrations as they arrived it would file them before knowing the namespace. So it buffers a
connection's registrations until the namespace is known, then drains them in arrival order.
```

I'm not sure I understand this, wouldn't workers send their metadata first? I assume the worker metadata would be what includes the `namespace`?

This paragraph also covers both SDK and engine behavior, should split into separate paragraphs.

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+
+The buffer is bounded by a grace period rather than waiting forever, because a worker might never
+announce itself at all. When the grace expires the connection is fixed to `default` and drained. This
+is what makes the whole feature backward compatible: a worker built against an older SDK sends no
+namespace, waits out the grace, lands in `default`, and behaves exactly as it always did. Every
+namespace field on the wire is optional for the same reason.
```
anthonyiscoding: I don't really understand this either it's non-technical and non-specific language. I'm not certain why a buffer has a grace period and it's not the same thing. We should also:

1. Specify what the default grace period is
2. Document how to change it in using-iii or creating-workers
3. Not document what we no longer do/what maintains backwards compatibility

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+
+Durable queue identity is the visible case. A subscriber queue is now scoped by namespace, so two
+subscribers of the same topic and function id in different namespaces are two queues receiving the
+same event rather than two competing consumers of one queue. That is the correct behaviour, and it
+means the queue names changed. The builtin broker migrates its own queues; an external broker holds
+queues iii does not own, so those need operator action.
+
+<Note>
+  For the migration steps, see [Upgrading from 0.22.x](../upgrading/from-0-22-x).
+</Note>
```
anthonyiscoding: Have you used a tool like caveman and/or ponytail? They can potentially help with this.

https://github.com/juliusbrussee/caveman
https://github.com/DietrichGebert/ponytail

```suggestion
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+
+## Isolate workers with namespaces
+
+Use this when two projects on one engine both want the same worker name or the same function id, and
+neither can rename. Declaring a namespace scopes each worker's registrations so they coexist.
```
anthonyiscoding: Move this to the top of the doc, drop if it's already covered up there.

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+</Tabs>
+
+<Warning>
+  The browser SDK has no environment to read, so `namespace` must be passed explicitly there.
+</Warning>
+
+Every function, service, and trigger type this worker registers now lands in `orders`. A second
+worker can register `state` in `analytics` on the same engine without either one displacing the
+other.
+
```
anthonyiscoding: This section belongs in creating-workers/namespaces. The resolution order is also incorrect because the SDK does not automatically pick up the namespace from the environment, it needs to be passed in through code. This is something I originally suggested and then we decided against to avoid any magic.

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+the worker under `default`.
+
+<Tabs>
+  <Tab title="Node / TypeScript">
+    ```typescript
+    import { registerWorker } from "iii-sdk";
+
+    const worker = registerWorker(process.env.III_URL!, {
+      workerName: "state",
+      namespace: "orders",
```
anthonyiscoding:
```suggestion
      namespace: process.env.III_NAMESPACE,
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+## Isolate workers with namespaces
+
+Use this when two projects on one engine both want the same worker name or the same function id, and
+neither can rename. Declaring a namespace scopes each worker's registrations so they coexist.
+
+### Declare the worker's namespace
+
+Pass `namespace` when registering the worker. The SDK resolves it in this order: the explicit option,
+then the `III_NAMESPACE` environment variable, then nothing at all, in which case the engine files
+the worker under `default`.
```
anthonyiscoding:
```suggestion
Pass `namespace` when registering the worker. The recommended approach is to always reference
the III_NAMESPACE environment variable so that it can be dynamically assigned at runtime.

The SDK's behavior is to assign the `default` workspace when `namespace` is not specified. If 
namespace is an invalid value (empty string, null, none, undefined, non-string) then the SDK will
error.
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+
+  </Tab>
+  <Tab title="Python">
+    ```python
+    import os
+    from iii import register_worker, InitOptions
+
+    worker = register_worker(
+        os.environ["III_URL"],
+        InitOptions(worker_name="state", namespace="orders"),
```
anthonyiscoding:
```suggestion
        InitOptions(worker_name="state", namespace=os.environ["III_NAMESPACE"]),
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+    use iii_sdk::{InitOptions, register_worker};
+
+    let url = std::env::var("III_URL").expect("III_URL must be set");
+    let worker = register_worker(
+        &url,
+        InitOptions {
+            namespace: Some("orders".into()),
+            ..Default::default()
+        },
+    );
```
anthonyiscoding: Double check this code, I'm not a Rust dev, just copying how url worked.
```suggestion
    let url = std::env::var("III_URL").expect("III_URL must be set");
    let namespace = std::env::var("III_NAMESPACE").expect("III_NAMESPACE must be set");
    let worker = register_worker(
        &url,
        InitOptions {
            namespace: &namespace,
            ..Default::default()
        },
    );
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+    ```
+
+  </Tab>
+  <Tab title="Browser">
+    ```typescript
+    import { registerWorker } from "iii-browser-sdk";
+
+    const worker = registerWorker("ws://localhost:49135", {
+      workerName: "checkout-tab",
+      namespace: "orders",
```
anthonyiscoding:
```suggestion
      namespace: "users-browser", // Browser has no environment variable
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+    });
+    ```
+
+  </Tab>
+</Tabs>
+
+<Warning>
+  The browser SDK has no environment to read, so `namespace` must be passed explicitly there.
```
anthonyiscoding:
```suggestion
  The browser SDK has no environment to read, so `namespace` must be passed explicitly or obtained
  at runtime via other methods. For untrusted clients like browsers it's common to hardcode their namespace.
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+</Warning>
+
+Every function, service, and trigger type this worker registers now lands in `orders`. A second
+worker can register `state` in `analytics` on the same engine without either one displacing the
+other.
+
+### Namespace a worker iii starts for you
```
anthonyiscoding:
```suggestion
### Shipping with a custom namespace
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+
+Managed workers pick the value up from `III_NAMESPACE`, so set it in the worker's `iii.worker.yaml`
+manifest:
```
anthonyiscoding:
```suggestion
If you have a preferred namespace for your shipped worker you can set it in its `iii.worker.yaml`:
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+
+```yaml iii.worker.yaml
+env:
+  III_NAMESPACE: orders
+```
+
+A worker whose manifest you don't control inherits `III_NAMESPACE` from the environment the engine
+runs in, which namespaces every managed worker in that project at once. There is no per-worker
+`namespace:` key in `config.yaml`.
```
anthonyiscoding: This here was also documenting what the software doesn't do.

```suggestion
<Note>
This value will replace `default` when namespace is otherwise unspecified. If a consumer does specify `III_NAMESPACE` in their environment then that env takes precedence.
</Note>
```

docs/next/understanding-iii/namespaces.mdx
```diff
@@ -0,0 +1,348 @@
...
+    {
+        eprintln!("{code}: {worker_name} in {namespace} owned by {owner_worker_id}");
+    }
+    ```
+  </Tab>
+</Tabs>
+
+Fix a `WORKER_NAMESPACE_CONFLICT` by giving one of the two workers a different namespace or a
+different `workerName`. A worker restarting against its own not-yet-cleaned connection does not
+produce this error; the engine lets a restart reclaim its own name.
```
anthonyiscoding: This section belongs in using-iii/namespace

docs/next/upgrading/from-0-22-x.mdx
```diff
@@ -0,0 +1,115 @@
...
+and behave exactly as before, so most projects upgrade with no code change. The one hard break is
+durable queue naming on the RabbitMQ adapter. Apply the steps below that touch surfaces your project
+uses.
+
+## Step 1: Redeclare RabbitMQ durable subscriber queues
+
+Durable subscriber queue names are now namespace-qualified, so two subscribers of the same topic and
+function id in different namespaces get two queues instead of competing for one. The RabbitMQ adapter
+has no automatic migration: queues declared by a pre-0.23 deploy keep their old names and stop being
+addressed after the upgrade.
```
anthonyiscoding: I'm unclear what `queues declared by a pre-0.23 deploy keep their old names and stop being
addressed after the upgrade.` means. Is this saying that old named queues stop being used after upgrade? Can make this clearer if so.

docs/next/upgrading/from-0-22-x.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+| Version | Subscriber queue                        | Dead-letter queue                     |
+| ------- | --------------------------------------- | ------------------------------------- |
+| 0.22.x  | `iii.{topic}.{function_id}.queue`       | `iii.{topic}.{function_id}.dlq`       |
+| 0.23.x  | `iii.{topic}.{function_id}@{ns}.queue`  | `iii.{topic}.{function_id}@{ns}.dlq`  |
+
+`{ns}` is the subscriber's namespace, `default` when it declared none. Any `@` or `\` inside a
+topic, function id, or namespace is backslash-escaped before the join, so the single unescaped `@` is
+always the namespace boundary.
```
anthonyiscoding:
```suggestion
topic, function id, or namespace is backslash-escaped before the join.
```

docs/next/upgrading/from-0-22-x.mdx
```diff
@@ -0,0 +1,115 @@
...
+For each affected subscriber:
+
+1. Stop publishing to the topic.
+2. Let the existing consumers drain the old queue to empty.
+3. Upgrade and restart. The adapter declares the new namespace-qualified queue on the next subscribe.
+4. Delete the drained queue and its dead-letter queue from the broker.
+
+## Step 2: Verify the builtin broker migrated
+
+The builtin broker migrates itself. On the first boot after the upgrade it drains each legacy
```
anthonyiscoding: What is meant by builtin broker? The engine's queue broker?
Also we should have no notion of builtin in the docs.
```suggestion
The engine's internal broker migrates itself. On the first boot after the upgrade it drains each legacy
```

docs/next/upgrading/from-0-22-x.mdx
```diff
@@ -0,0 +1,115 @@
...
+subscriber queue into its namespaced identity and logs one line per queue:
+
+```text
+Drained legacy subscriber queue into its namespace
+```
+
+No action is needed. A queue name that already contains the separator is skipped and logged as such,
+because it cannot be told apart from an already-migrated name.
+
+## Step 3: Tune the registration grace, if needed
```
anthonyiscoding:
```suggestion
## Step 3: Tune the registration grace period, if needed
```

docs/next/upgrading/from-0-22-x.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+The default grace is 5000 ms. Raise it if workers on slow links register but land in `default`
+unexpectedly:
+
+```yaml config.yaml
+workers:
+  - name: iii-worker-manager
+    config:
+      registration_namespace_grace_ms: 10000
+```
```
anthonyiscoding, line 89: This needs to be documented in using-iii as well

docs/next/upgrading/from-0-22-x.mdx
```diff
@@ -0,0 +1,115 @@
...
+No action is needed. A queue name that already contains the separator is skipped and logged as such,
+because it cannot be told apart from an already-migrated name.
+
+## Step 3: Tune the registration grace, if needed
+
+A connection's namespace arrives on its `engine::workers::register` call, which SDKs send after their
+first registration frames. The engine buffers a connection's registrations until the namespace is
+known, or until a grace period expires and the connection is fixed to `default`.
+
+The default grace is 5000 ms. Raise it if workers on slow links register but land in `default`
```
anthonyiscoding:
```suggestion
The default grace period is 5000 ms. Raise it if workers on slow links register but land in `default`
```

docs/next/upgrading/from-0-22-x.mdx
```diff
@@ -0,0 +1,115 @@
...
+unexpectedly:
+
+```yaml config.yaml
+workers:
+  - name: iii-worker-manager
+    config:
+      registration_namespace_grace_ms: 10000
```
anthonyiscoding: I'm surprised max grace period is a per-worker config and not an engine-wide config. We should reconsider this. Also this means workers must already be known to the engine and have a configuration before their grace period can be adjusted which is awkward to do in a system with live registration.

docs/next/upgrading/from-0-22-x.mdx
```diff
@@ -0,0 +1,115 @@
...
+```
+
+`III_NAMESPACE_GRACE_MS` overrides the configured value at runtime and takes precedence.
```
anthonyiscoding: Where is this set? engine's environment or the worker's?

docs/next/upgrading/from-0-22-x.mdx
```diff
@@ -0,0 +1,115 @@
...
+
+## Step 4: Check for `engine::*` function ids outside `default`
+
```
anthonyiscoding, line 96:
```suggestion
<Note>
Your own workers are unlikely to have `engine::` prefixed function names but we've documented this as it would be a breaking change if they do exist.
</Note>
```

docs/next/upgrading/from-0-22-x.mdx
```diff
@@ -0,0 +1,115 @@
...
+`engine::*` is reserved for engine and builtin infrastructure, which lives in `default`. A worker
```
anthonyiscoding:
```suggestion
`engine::*` is reserved for engine and builtin infrastructure, which is in `default`. A worker
```

docs/next/upgrading/from-0-22-x.mdx
```diff
@@ -0,0 +1,115 @@
...
+that registers an `engine::*` function id while running in another namespace now has that one
+registration refused with a `FUNCTION_NAMESPACE_CONFLICT`; the connection stays open and its other
+functions keep serving. Rename any such id to your own `service::name` prefix.
+
+This only affects workers that both use a namespace and register under the `engine::` prefix, so a
+project that has not adopted namespaces is unaffected.
```
anthonyiscoding: Does every project not inherently use namespaces going forward? Even those that don't specify will get the default namespace.
```suggestion
```

docs/next/upgrading/from-0-22-x.mdx
```diff
@@ -0,0 +1,115 @@
...
+## Step 5: Scope RBAC expose rules when adopting a namespace
+
+`expose_functions` rules on the `iii-worker-manager` RBAC listener are namespace-scoped. A rule that
+names no namespace applies to `default` only, so it will not expose a function reached in another
+namespace. When you move a worker into a namespace, scope its rules to match:
+
+```yaml
+expose_functions:
+  - match: "orders::*"
+    namespace: orders
```
anthonyiscoding, line 121: Do namespaces accept glob matches in RBAC workers? I believe we discussed allowing it, if so we should show a second example here of that.

docs/next/upgrading/from-0-22-x.mdx
```diff
@@ -0,0 +1,115 @@
...
+```
+
+`forbidden_functions` and `allowed_functions` still match on the bare function id and apply in every
+namespace; they come from the session's auth result, which carries no namespace.
+
```
anthonyiscoding, line 130: This needs to be documented in creating-workers/worker-manager

anthonyiscoding (changes requested):
Overall good but definitely some technical accuracy, SDK/engine behavior discrepancies, wording/phrasing, and organization issues to address.

Note: There may be one or two comments that are redundant and already addressed in the docs as there were quite a lot of docs to review. Use your best judgement in these cases and we can always revise later if needed.

Also I'd suggest trying out caveman and ponytail to make your Agent a bit more technical when it's writing. They're not expressly for this but can help.

https://github.com/juliusbrussee/caveman
https://github.com/DietrichGebert/ponytail

You can also try telling the agent to "Use ASD-STE100 simplified technical english" which seems to be working well for me.

anthonyiscoding (changes requested):
Overall good but definitely some technical accuracy, SDK/engine behavior discrepancies, wording/phrasing, and organization issues to address.

Note: There may be one or two comments that are redundant and already addressed in the docs as there were quite a lot of docs to review. Use your best judgement in these cases and we can always revise later if needed.

Also I'd suggest trying out caveman and ponytail to make your Agent a bit more technical when it's writing. They're not expressly for this but can help.

https://github.com/juliusbrussee/caveman
https://github.com/DietrichGebert/ponytail

You can also try telling the agent to "Use ASD-STE100 simplified technical english" which seems to be working well for me.

docs/next/creating-workers/queues.mdx
guibeira: Yeah, imagine the scenario like this. Topics are global, and anyone can subscribe, even from different namespaces.
```bash

                 publish(topic="emails")
                                │
                                ▼
                    ┌─────────────────────┐
                    │ global topic        │
                    │ "emails"            │
                    └──────────┬──────────┘
                               │ fan-out
                   ┌───────────┴───────────┐
                   ▼                       ▼
      ┌────────────────────────┐  ┌────────────────────────┐
      │ durable queue          │  │ durable queue          │
      │ emails::send@orders    │  │ emails::send@analytics │
      └────────────┬───────────┘  └────────────┬───────────┘
                   │                           │
                   ▼                           ▼
      orders / emails::send       analytics / emails::send
```

docs/next/creating-workers/worker-manifest.mdx
guibeira: Make sense, let me update this.

docs/next/reference/engine-protocol.mdx
guibeira: Good point. This is not a separate namespace system. It is the target namespace for the function that the trigger calls.

Usually, the target namespace is the same as the worker namespace. A trigger can also call a function in a different namespace. For this reason, `RegisterTrigger` includes the target `namespace`.

The typed helper sets this field to the worker namespace. The low-level `registerTrigger` API does not set it. If the field is not present, the target namespace is `default`.

I will update the documentation and add examples for both cases.

docs/next/reference/engine-protocol.mdx
guibeira: Fixed in commit [`02af50fc`](https://github.com/iii-hq/iii/commit/02af50fc809d5e36983abce4746410b3b09da2f9). `RegistrationRejected` now uses `worker_name` only for `WORKER_NAMESPACE_CONFLICT` and the separate `function_id` field for `FUNCTION_NAMESPACE_CONFLICT`. The engine, SDKs, console bridge, protocol documentation, and regression tests were updated accordingly.

docs/next/understanding-iii/namespaces.mdx
guibeira: This is the default, if you don't pass it, it will be used. This way is the explicit way.

docs/next/understanding-iii/namespaces.mdx
guibeira: Addressed at commit [`f25f543b`](https://github.com/iii-hq/iii/commit/f25f543b).

docs/next/reference/engine-protocol.mdx
guibeira: The all-namespace discovery path is `engine::functions::list` and `engine::workers::list`. Both return entries from every namespace, and each row includes its `namespace`. The behavior described here applies only to `info`, which inspects one function or worker and must resolve a single identity. Without an explicit namespace, it preserves backward-compatible lookup: `default` wins, a unique non-default match resolves, and multiple non-default matches return an ambiguity. An agent can use `list` to discover the full catalog and then pass the selected namespace to `info`.

docs/next/understanding-iii/namespaces.mdx
guibeira: This is the default behavior, I would prefer to keep being explicit in this case.

docs/next/understanding-iii/namespaces.mdx
guibeira: This is the default behavior if you don't send namespace, for this reason, I would prefer to be explicit.

docs/next/upgrading/from-0-22-x.mdx
guibeira: Addressed in commit [`7a698ef3`](https://github.com/iii-hq/iii/commit/7a698ef3). The guide now states that a 0.23 subscriber declares and consumes a new namespace-qualified queue, does not rename, consume, or migrate the old queue, and that the old durable binding can continue to route copies until the queue is deleted.

docs/next/understanding-iii/namespaces.mdx
guibeira: Ideally yes, but we can receive the register function after registering a worker. When this happens, we can sign that function to the correct namespace as soon as the worker finishes the registration.

docs/next/understanding-iii/namespaces.mdx
guibeira: Addressed in commit [`b1b178b8`](https://github.com/iii-hq/iii/commit/b1b178b8). The docs now describe the `5000 ms` namespace resolution timeout, list the message types held while the namespace is pending, specify both timeout outcomes, remove the backwards-compatibility narrative, and link to the using-iii configuration section for `registration_namespace_grace_ms` and `III_NAMESPACE_GRACE_MS`.

docs/next/upgrading/from-0-22-x.mdx
guibeira: Good question. `III_NAMESPACE_GRACE_MS` is read from the engine process environment, not from a worker environment. It applies to namespace resolution for all new worker connections and overrides `registration_namespace_grace_ms` from `config.yaml`. Clarified in commit [`4bd886f4`](https://github.com/iii-hq/iii/commit/4bd886f4) in both the upgrade guide and the configuration documentation.

docs/next/upgrading/from-0-22-x.mdx
guibeira: You are right. Every worker has an effective namespace. When none is specified, the worker uses the default namespace. The intended distinction here is between the default namespace and a non-default namespace, not between projects that have or have not adopted namespaces. I updated the wording to distinguish non-default namespaces and removed the final paragraph in commit [3a17f333](https://github.com/iii-hq/iii/commit/3a17f333fdf0fc0450c8c88a0a25ab030bdb1953).

docs/next/upgrading/from-0-22-x.mdx
guibeira: Good point. The grace period is now an engine-wide setting at the root of config.yaml. The engine loads it before it creates any workers, so live workers do not need to be known or configured in advance. Resolved in commit [a0e3af6e](https://github.com/iii-hq/iii/commit/a0e3af6ee8339c2c9cc1289f86514ff2fca399e4).

docs/next/reference/engine-protocol.mdx
guibeira: Good point. The docs now state that worker registration and function registration are separate operations. They also explain that an early `RegisterFunction` message is held until the namespace is known; it is not registered in `default` and moved later.

The updated section makes the partial-registration behavior explicit: the current owner keeps the function, later invocations go to that owner, and the rejected worker stays connected with its other functions. It also warns that a connected worker does not confirm that all function registrations succeeded.

Addressed in commit [`72851e5`](https://github.com/iii-hq/iii/commit/72851e51).

sdk/packages/go/iii/client.go
```diff
@@ -953,6 +1190,19 @@ func (c *Client) Close() error {
...
+		views := make([]*Client, 0, len(c.namespaceViews))
+		for _, v := range c.namespaceViews {
+			views = append(views, v)
+		}
+		c.namespaceViews = nil
+		c.viewsMu.Unlock()
+		for _, v := range views {
+			_ = v.Close()
+		}
+
```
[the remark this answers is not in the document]
guibeira: The `UseNamespace` and `namespaceViews` implementation was removed from the Go SDK, so this panic path no longer exists. The current `Close` implementation has no namespace-view map to clear. Closing this outdated thread.

docs/next/creating-workers/worker-manifest.mdx
guibeira: Addressed in commit [`02af50fc`](https://github.com/iii-hq/iii/commit/02af50fc). The manifest docs now state that managed and reusable workers should receive `III_NAMESPACE` from the compose deployment. The explicit SDK option remains available for programmatic registration.

docs/next/creating-workers/worker-manifest.mdx
guibeira: Addressed in commit [`02af50fc`](https://github.com/iii-hq/iii/commit/02af50fc). The docs now identify the compose deployment as the place to set `III_NAMESPACE` and state that `config.yaml` has no per-worker `namespace` key.

docs/next/creating-workers/workers.mdx
guibeira: Addressed in commit [`02af50fc`](https://github.com/iii-hq/iii/commit/02af50fc). The current text describes the effective SDK resolution order and uses `default` without attributing it to an engine configuration.

docs/next/reference/engine-protocol.mdx
guibeira: Addressed in commit [`02af50fc`](https://github.com/iii-hq/iii/commit/02af50fc). The docs now explain that this is the target namespace of the function called by the trigger, not a second namespace system. They also distinguish typed helpers from the low-level API.

docs/next/understanding-iii/namespaces.mdx
guibeira: Clarified in commit [`72851e5`](https://github.com/iii-hq/iii/commit/72851e51). A client can send registration messages before worker metadata. The engine holds those messages; it does not put them in `default` and move them later. After `engine::workers::register`, it sets the connection namespace and processes the held messages in order.

docs/next/upgrading/from-0-22-x.mdx
guibeira: Addressed in commit [`3da9b5b4`](https://github.com/iii-hq/iii/commit/3da9b5b4). The upgrade guide now uses `engine-managed queue storage` and explains the startup migration sequence and its limits without using `builtin broker`.

docs/next/upgrading/from-0-22-x.mdx
guibeira: The current wording uses `engine infrastructure` and does not expose the internal `builtin` term. This also matches the earlier feedback on the queue migration documentation, so I kept the more specific current wording.

docs/next/understanding-iii/namespaces.mdx
guibeira: Checked against the current Rust SDK. `InitOptions.namespace` is `Option<String>`, so the current `Some("orders".into())` example is valid and intentionally shows an explicit override. The SDK reads `III_NAMESPACE` when the option is absent. The suggested `namespace: &namespace` value does not match the current type.

docs/next/upgrading/from-0-22-x.mdx
guibeira: Addressed in commit [`b1b178b8`](https://github.com/iii-hq/iii/commit/b1b178b8). The `using-iii/configuration` guide now documents the global `registration_namespace_grace_ms` setting, its `5000 ms` default, and the engine-process `III_NAMESPACE_GRACE_MS` override.

docs/next/upgrading/from-0-22-x.mdx
```diff
@@ -0,0 +1,116 @@
...
+The builtin broker migrates itself. On the first boot after the upgrade it drains each legacy
+subscriber queue into its namespaced identity and logs one line per queue:
+
+```text
+Drained legacy subscriber queue into its namespace
+```
+
+No action is needed. A queue name that already contains the separator is skipped and logged as such,
+because it cannot be told apart from an already-migrated name.
```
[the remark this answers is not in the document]
guibeira: Addressed in commit [`1e0425b`](https://github.com/iii-hq/iii/commit/1e0425beb3e4531c1fd54df319a0fa47786a6e87). The migration checklist now requires recovery of every legacy queue skipped because its name contains `@`. The existing warning gives the recovery sequence: stop publishing, restart 0.22.x with the same storage, drain the queue, and retry the upgrade.

docs/next/understanding-iii/namespaces.mdx
guibeira: Addressed in commit [`1e0425b`](https://github.com/iii-hq/iii/commit/1e0425beb3e4531c1fd54df319a0fa47786a6e87). The introduction now uses shorter wording: the function has the same id in each namespace, and the namespace remains a separate routing value.

docs/next/understanding-iii/namespaces.mdx
guibeira: Addressed in commit [`1e0425b`](https://github.com/iii-hq/iii/commit/1e0425beb3e4531c1fd54df319a0fa47786a6e87). The problem section is now framed around the enabled use case: one reusable worker package can be deployed for multiple tenants on one engine while each tenant keeps the same worker name and function ids.

docs/next/understanding-iii/namespaces.mdx
guibeira: Addressed in commit [`1e0425b`](https://github.com/iii-hq/iii/commit/1e0425beb3e4531c1fd54df319a0fa47786a6e87). The collision section now states the ownership rules directly and separates the two outcomes: a worker-name conflict closes the new connection, while a function-id conflict rejects only that function.

docs/next/using-iii/configuration.mdx
guibeira: Addressed in commit [`1e0425b`](https://github.com/iii-hq/iii/commit/1e0425beb3e4531c1fd54df319a0fa47786a6e87). Following the later review feedback on location, the practical guide is now `using-iii/namespaces.mdx`. It documents the recommended compose-time `III_NAMESPACE` setting and the explicit SDK option.

docs/next/reference/engine-protocol.mdx
guibeira: Addressed in commit [`1e0425b`](https://github.com/iii-hq/iii/commit/1e0425beb3e4531c1fd54df319a0fa47786a6e87). The docs now state that the queue worker supplied with the engine registers its `engine::queue::*` functions in `default`, so the reserved-id check does not reject them. The rejection applies to a custom worker connection that registers an `engine::*` id in a non-default namespace.

docs/next/understanding-iii/namespaces.mdx
guibeira: Addressed in commit [`1e0425b`](https://github.com/iii-hq/iii/commit/1e0425beb3e4531c1fd54df319a0fa47786a6e87). I removed the implementation-focused “What it costs” section from the conceptual guide. Operational migration details remain in the upgrade guide.

docs/next/understanding-iii/namespaces.mdx
guibeira: Addressed in commit [`1e0425b`](https://github.com/iii-hq/iii/commit/1e0425beb3e4531c1fd54df319a0fa47786a6e87). Following the later location feedback, the procedural content moved to `using-iii/namespaces.mdx`. The guide documents the current SDK resolution order: explicit option, then `III_NAMESPACE`, then `default`.

docs/next/understanding-iii/namespaces.mdx
guibeira: Addressed in commit [`1e0425b`](https://github.com/iii-hq/iii/commit/1e0425beb3e4531c1fd54df319a0fa47786a6e87). Deployment configuration, SDK examples, cross-namespace calls, trigger targets, discovery, and conflict handling now live in the new `using-iii/namespaces.mdx` guide. The understanding page now contains concepts only.

docs/next/upgrading/from-0-22-x.mdx
guibeira: Addressed in commit [`1e0425b`](https://github.com/iii-hq/iii/commit/1e0425beb3e4531c1fd54df319a0fa47786a6e87). Namespace values do not support glob matching in the current implementation. Function-id `match` values support `*`, while `namespace` uses an exact match. The worker-manager and upgrade guides now state this and require one rule per namespace.

docs/next/upgrading/from-0-22-x.mdx
guibeira: Addressed in commit [`1e0425b`](https://github.com/iii-hq/iii/commit/1e0425beb3e4531c1fd54df319a0fa47786a6e87). The worker-manager guide now has a “Namespace-scoped function rules” section with function-id and metadata examples. It also explains the `default` scope and exact namespace matching.

docs/next/reference/engine-protocol.mdx
```diff
@@ -120,9 +121,31 @@ A trigger can call a function in a different namespace. For this reason, `Regist
...
+Both SDK paths set this field to the worker's namespace when the caller omits it: the typed helper
+`registerTriggerType(...).registerTrigger` and the low-level `registerTrigger`. A trigger names a
+function, and a worker's functions land in the worker's namespace, so a trigger that defaulted
+anywhere else would fire and resolve nothing.
+
+The wire default is unchanged: an absent `namespace` still means `default` to the engine. What
+changed is that the SDKs stopped omitting it, except for a worker that declared no namespace of its
+own -- there is nothing to inherit, so the field stays absent and the engine reads it as `default`.
+Set `namespace` explicitly to target another namespace, `default` included: that is how a worker
+inside a namespace reaches an engine builtin.
```
anthonyiscoding: Delete

anthonyiscoding (changes requested):

anthonyiscoding (approved):

PR #1987, opened 2026-07-17
fix(queue): rebuild the transport named by the stored configuration at boot

### LINKED ISSUE #1986: queue worker: restart silently falls back to the builtin adapter — persisted config is adopted, but its transport is never built
## Summary

When `iii-queue` is configured with a non-builtin adapter (e.g. `rabbitmq`) and the engine's `config.yaml` is writable, an in-place process restart silently boots the queue worker on the **builtin in-memory adapter** instead. The first boot persists the queue config to the configuration store and strips the `config:` block from `config.yaml` (seed-strip). On the next restart the worker is constructed with `config: None` → builtin adapter; the boot path then re-adopts the persisted configuration **value** (which still says `rabbitmq`) but never rebuilds the **transport**, and it overwrites the diff signal that would have triggered the runtime hot-swap. Result: 0 broker consumers, durable queues freeze with their backlog intact, and any work enqueued after the restart goes into the in-memory builtin queue and is lost on the next shutdown — while the process keeps reporting healthy (cron triggers keep firing, subscriptions log `[REGISTERED]`).

## Environment

- Observed in a production-like staging environment on engine **0.21.4** (Docker, RabbitMQ adapter, engine started via `iii --config <file>.yaml`).
- Root cause traced and confirmed present on **`main` @ `d13a407`** — all `file:line` references below are against that commit.
- Precondition: `config.yaml` is writable inside the container (the Docker template deliberately bind-mounts it `:ro`, which prevents the seed-strip — see the comment at `engine/src/workers/config_rewrite.rs:140-153`; any deployment that copies the file into the image or mounts it read-write is exposed).

## Reproduction

1. Declare `iii-queue` with a non-builtin adapter in `config.yaml` (file writable by the engine):

   ```yaml
   workers:
     - name: iii-queue
       config:
         adapter:
           name: rabbitmq
           config:
             amqp_url: ${AMQP_URL}
             queue_mode: standard
         queue_configs:
           # per-topic max_retries / backoff ...
   ```

2. Start the engine (fresh container / clean working directory). Verify:
   - the broker has consumers on the `iii.*` queues (`rabbitmqctl list_consumers`);
   - after boot, the `config:` block under `- name: iii-queue` in `config.yaml` has been replaced by the seed-strip comment, and the value now lives in `./config/iii-queue.yaml`.

3. Restart the process **in place** — `docker restart` of the same container, or kill + relaunch in the same working directory (so the stripped `config.yaml` and `./config/iii-queue.yaml` survive).

4. Check the broker again.

### Observed

- `rabbitmqctl list_consumers` shows **0** consumers on any `iii.*` queue, permanently.
- Boot log shows the builtin adapter path: `iii::builtins::queue No persisted queue state found, rebuilding from job records` + `Queue state rebuilt from storage`.
- All `Started function queue consumer` lines (42 in our case) appear within <50 ms of boot — no AMQP setup happened.
- Subscriptions still log `[REGISTERED] Subscription <topic> → <ns::fn>` a few seconds later, but they are registered against the builtin adapter.
- Durable queues freeze (e.g. a work queue stuck at ~8,600 ready messages, 0 consumers) while internal cron triggers keep firing — the process looks alive and healthy.
- We traced 17 consecutive in-place restarts over ~2 h: every single boot landed on builtin. A full container re-create (fresh writable layer → original `config.yaml` restored) immediately recovered ~94 consumers and stayed stable.
- The YAML never changed between the working and broken boots (git diff empty) — the mutation lives in the container's writable layer.

### Expected

A restart resumes on the adapter named by the configuration source of truth — the persisted entry, which still says `rabbitmq`. At minimum, if the engine boots on a different adapter than the stored configuration selects, it should fail loudly or log an error, not silently degrade to an in-memory queue.

## Root cause analysis

The queue worker's boot path adopts the persisted configuration **value** but never its **transport**, and destroys the diff that would have triggered the hot-swap. Chain, against `main` @ `d13a407`:

1. **Adapter selection happens only from the yaml block passed to `create`.** The boot pipeline hands each worker entry's `config:` block verbatim to the constructor (`WorkerEntry::create_worker`, `engine/src/workers/config.rs:397-411`). `create_with_adapters` resolves the adapter exclusively from that block; if the block is `None`, it falls back to `DEFAULT_ADAPTER_NAME = "builtin"` (`engine/src/workers/traits.rs:206-251`; `engine/src/workers/queue/queue.rs:1261`).

2. **First boot seeds the store, then strips the yaml.** With the block present, the rabbitmq adapter is built. `start_background_tasks` seeds the persisted `iii-queue` configuration entry (`engine/src/workers/queue/queue.rs:1142-1154`; seed-only-if-absent in `engine/src/workers/queue/configuration.rs:55-62`), persisted by the `fs` adapter to `./config/iii-queue.yaml` (`engine/src/workers/configuration/adapters/fs.rs:43-48`). Then `serve()` rewrites `config.yaml`, replacing the now-redundant `config:` block with a breadcrumb comment (`engine/src/workers/config.rs:768-836`, `apply_strip` call at `:831`; strip mechanics in `engine/src/workers/config_rewrite.rs:54-167`). This is the documented intent: the persisted entry is the runtime source of truth, the yaml block is seed-only.

3. **Restart: `create` sees `config: None` → builtin.** The stripped `config.yaml` and the persisted entry both survive an in-place restart. Step 1 now resolves builtin, and the builtin adapter logs `No persisted queue state found, rebuilding from job records` at create time (`engine/src/workers/queue/adapters/builtin/adapter.rs:183-190` → `engine/src/builtins/queue.rs:230-236`). ✓ matches the observed log.

4. **The persisted entry is re-adopted — value only.** `start_background_tasks` fetches the stored entry, which correctly says `adapter: rabbitmq`, validates it, and adopts it via `self.set_config(config)` (`engine/src/workers/queue/queue.rs:1156-1179`). But `set_config` swaps **only the config**, never the adapter (`queue.rs:696-699`). Consumers are then spawned from `adapter_snapshot()` — still builtin (`queue.rs:1186-1191`). That is why all "Started function queue consumer" lines appear within milliseconds: no AMQP connection is made. This adoption is also completely silent (no log line), unlike e.g. `iii-state`, which logs "Using persisted … configuration entry" — which is part of why the bug is hard to spot.

5. **The runtime hot-swap can never catch it.** `apply_config` does have a full transport hot-swap branch: it compares `effective_adapter(&old)` against the new value and rebuilds the adapter on change (`queue.rs:975-1015`). But `old = config_snapshot()` was already overwritten in step 4 with the persisted (rabbitmq) config — so on any subsequent configuration event, old == new, the hot-swap branch is skipped, and the adapter stays builtin for the life of the process.

6. **Full re-create recovers** because the fresh writable layer restores the original `config.yaml` with the adapter block, so step 1 resolves rabbitmq again. ✓ matches the observed recovery.

The defect is the asymmetry between the documented intent ("the stored entry is the runtime source of truth; the yaml block is seed-only" — which is what justifies the strip) and the boot implementation, where **transport selection is derived exclusively from the yaml block passed to `create`**.

### Suggested fix direction

All the needed pieces already exist: `effective_adapter` (`queue.rs:740-746`), `resolve_adapter` (`queue.rs:724-735`), and the atomic `set_live` (`queue.rs:713-719`). In `start_background_tasks`, when the validated stored config selects a different effective adapter than the one built at `create`, rebuild the transport and swap both atomically instead of calling `set_config`:

```rust
Ok(config) => match config.validate() {
    Ok(()) => {
        if Self::effective_adapter(&self.config_snapshot()) != Self::effective_adapter(&config) {
            match Self::resolve_adapter(&self.engine, &config).await {
                Ok(adapter) => self.set_live(config, adapter),
                Err(err) => tracing::warn!(error = %err, "iii-queue: failed to build the adapter named by the stored configuration; continuing with the statically-resolved adapter"),
            }
        } else {
            self.set_config(config);
        }
    }
    Err(err) => /* unchanged */,
},
```

The swap happens before consumers are spawned (`queue.rs:1186-1191`), so they start directly on the right transport; on build failure the current degrade-to-static behavior is preserved, consistent with the `apply_config` gate. The first-boot path (same effective adapter) is unchanged.

Note: the same "adopt the value but not the transport" pattern may exist in other adapter-backed `ConfigurableWorker`s whose yaml blocks are also seed-stripped (stream, pubsub, state) — not audited here, but probably worth a look in the same pass.

## Impact

- **Silent loss of durability on every in-place restart**: durable queues stop being consumed (backlog freezes intact), and all work enqueued after the restart lives in the builtin in-memory queue — lost on the next shutdown.
- **Restart loops with external health tooling**: monitoring that checks broker consumers sees 0, restarts the process, and lands in the exact same state. We hit precisely this — an external watchdog restarted the container every ~320 s for two hours.
- The process reports healthy the entire time (cron triggers fire, subscriptions register), so nothing internal flags the degradation.
- Recovery requires a full container re-create; a restart can never fix it.

## Related

- #1958 (RabbitMQ adapter: channel-level AMQP error silently kills every consumer) — different mechanism (channel death at runtime vs adapter selection at boot), but the same blast pattern: all consumers silently gone while the process reports healthy. Fixing this issue would also remove one of the ways an engine ends up consumer-less with no error surfaced.

racso-dev:
## Summary

When a process restarts in an environment where the engine has already persisted the queue worker's configuration (the seeded `config:` block is stripped from `config.yaml` after first boot, and the stored configuration entry becomes the runtime source of truth), the queue silently comes up on the builtin in-memory adapter even though the stored configuration names an external one (e.g. `rabbitmq`). The mechanism: `QueueWorker::create` resolves the adapter from the (now empty) static config — defaulting to `builtin` — and `start_background_tasks` adopted the validated stored configuration via `set_config` only, so the config *value* was updated but the live transport was never rebuilt; overwriting the snapshot also destroyed the old-vs-new adapter diff that `apply_config` relies on to trigger a transport hot-swap, so every consumer spawned on the builtin adapter and the process stayed pinned there for its whole lifetime while durable messages accumulated unconsumed in the external broker.

## Fix

In `start_background_tasks` (`engine/src/workers/queue/queue.rs`), when the validated stored configuration selects a different effective adapter than the one resolved at create time, resolve/build that adapter and swap config + transport atomically with `set_live` **before** consumers spawn. If the adapter build fails, degrade all-or-nothing to the statically-resolved adapter and seed config — the same gate `apply_config` applies for runtime updates. When the stored selection matches the create-time adapter (including `None` vs the explicit default), the existing `set_config` path is kept unchanged, so there is no spurious transport rebuild on the common path.

This is the minimal change: it reuses the existing `resolve_adapter` / `set_live` machinery already used by `apply_config` for runtime hot-swaps, only extending the boot-time adoption path to apply the same transport semantics it already applies to the config value. ~30 lines, no behavior change for first boots or for boots where the static and stored adapters agree.

## Testing

- `cargo test -p iii --all-features --test queue_configuration_e2e`
  - **without the fix** (patch stashed): 13/14 — the new test `restart_without_yaml_adapter_block_rebuilds_stored_transport` fails precisely on the assertion that the stored adapter's transport was rebuilt (red confirmed)
  - **with the fix**: 14/14 passed (green confirmed)
- New regression tests in `engine/tests/queue_configuration_e2e.rs`:
  - `restart_without_yaml_adapter_block_rebuilds_stored_transport` — simulates a restart where the yaml no longer carries the adapter block but the stored configuration names a non-default adapter; asserts the transport is rebuilt before consumers start
  - `restart_with_matching_adapter_keeps_boot_transport` — asserts no spurious swap when the stored selection matches the boot-time adapter (`None` vs explicit default)
- `cargo test -p iii --all-features --lib workers::queue` → 163 passed
- `cargo test -p iii --all-features --test queue_e2e_happy_path --test queue_e2e_failure_retry --test queue_e2e_fanout --test queue_e2e_concurrency_resilience --test queue_integration --test iii_queue_optin_registration` → all passed (1+3+3+7+2+12), 0 failed
- `RABBITMQ_URL=amqp://localhost:5672 cargo test -p iii --test rabbitmq_queue_integration` against a disposable `rabbitmq:3-management-alpine` broker → 20 passed
- `cargo build -p iii --all-features` → OK; `cargo fmt --all -- --check` → clean
- `cargo clippy -p iii --all-features --all-targets -- -D warnings` fails on pre-existing lints on `main` (engine clippy is currently commented out in CI); with those pre-existing lint classes allowed, no diagnostics touch the files modified here

## Licensing :
- [X] I license my contributions to this repository under Apache 2.0, and I have all necessary rights over the code I am contributing.

Fixes #1986

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved queue worker restarts to correctly restore the configured adapter and transport selection.
  * Ensured queued work reaches consumers after a restart when stored transport settings differ from defaults.
  * Preserved the existing transport when stored and active adapter selections already match.
  * Continued applying stored per-queue settings during restart.

### DIFF
--- a/engine/src/workers/queue/queue.rs
+++ b/engine/src/workers/queue/queue.rs
@@ -1169,7 +1169,36 @@ impl Worker for QueueWorker {
             // restart. On failure keep the seed (already validated in
             // `initialize`), matching the runtime apply gate.
             Ok(config) => match config.validate() {
-                Ok(()) => self.set_config(config),
+                // The stored entry is the runtime source of truth and the
+                // config.yaml block is seed-only (it is stripped after the
+                // first boot), so on a later boot `create` may have resolved
+                // the default adapter while the stored value names another
+                // one. The stored selection must drive the transport too:
+                // rebuild it here, before consumers spawn below. Adopting only
+                // the config value would also erase the old-vs-new adapter
+                // diff that `apply_config` relies on, silently pinning the
+                // queue to the wrong transport for the process lifetime.
+                Ok(()) => {
+                    if Self::effective_adapter(&self.config_snapshot())
+                        != Self::effective_adapter(&config)
+                    {
+                        match Self::resolve_adapter(&self.engine, &config).await {
+                            // No consumers exist yet, so unlike the hot-swap in
+                            // `apply_config` there is nothing to rebind.
+                            Ok(adapter) => self.set_live(config, adapter),
+                            // Keep the seed config alongside the seed adapter
+                            // (all-or-nothing, matching the `apply_config`
+                            // gate) so the two can never disagree.
+                            Err(err) => tracing::warn!(
+                                error = %err,
+                                "iii-queue: failed to build the adapter named by the stored \
+                                 configuration; continuing with static config"
+                            ),
+                        }
+                    } else {
+                        self.set_config(config);
+                    }
+                }
                 Err(err) => tracing::warn!(
                     error = %err,
                     "iii-queue: stored configuration is invalid; continuing with static config"
--- a/engine/tests/queue_configuration_e2e.rs
+++ b/engine/tests/queue_configuration_e2e.rs
@@ -5,10 +5,10 @@
 // See LICENSE and PATENTS files for details.
 
 //! End-to-end test for the `iii-queue` ↔ `configuration` worker integration:
-//! seed-on-first-boot, no-clobber across worker restarts, hot-add of a queue
-//! (live consumer), hot-change of per-queue settings, full adapter/transport
-//! hot-swap, all-or-nothing gate on an invalid value, and `${VAR:default}`
-//! expansion.
+//! seed-on-first-boot, no-clobber across worker restarts, adoption of the
+//! stored transport on restart, hot-add of a queue (live consumer), hot-change
+//! of per-queue settings, full adapter/transport hot-swap, all-or-nothing gate
+//! on an invalid value, and `${VAR:default}` expansion.
 //!
 //! Modeled on `engine/tests/http_configuration_e2e.rs` — composes the two
 //! workers against a real `FsAdapter` on a `tempfile::tempdir()`. The queue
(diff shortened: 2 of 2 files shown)


guibeira:
Hey @racso-dev thanks for your contribution. Before we move on, do you mind following the `License agreement required` instructions?

racso-dev:
Hey @guibeira done ✅ lmk if you need anything else!

guibeira (approved):

guibeira:
Thanks for your contribution @racso-dev !

PR #1988, opened 2026-07-17
fix(engine): park undeliverable trigger binds instead of dropping them (MOT-4081)

andersonleal:
## Context

Recoverable triggers (#1962) park a `register_trigger` whose trigger type is not yet registered and activate it when the type appears. MOT-4073 removed all worker-side pending-bind retry machinery on that strength — but a bind that reaches the engine while the type **is** registered and whose activation then fails was **dropped, not parked**, staying silently unbound until the binding worker restarted (e.g. guidance hooks missing from prompts, workflow wake/stamp hooks lost).

## Fix

Reuses the existing park/drain machinery (`pending_triggers` + type-registration drain); no new machinery.

- **Found-type path** (`TriggerRegistry::register_trigger`): a bind whose delivery to the provider fails is parked and returns `Deferred`; the drain re-activates it when the type (re)registers. Delivery failure is classified by a new `RegistratorUnavailable` marker, attached where `WorkerConnection`'s register send fails (provider channel closed — worker dying/disconnecting). Provider **rejections** (e.g. invalid config from the in-process cron/http/queue/… registrators) still fail synchronously: failure acks and `trigger_registration_failed` are unchanged, and a rejected re-registration keeps the previous live registration of the same id tracked for teardown.
- **Async registrator error** (`router_msg` handling `TriggerRegistrationResult`): the bind is re-parked instead of removed (fire-and-forget binds have no other recovery channel); the error is still forwarded to the originator.
- **Unregister vs late rejection race**: `unregister_trigger`'s live path now also clears `pending_triggers`, so an explicit unregister wins over a concurrent late-rejection re-park in every interleaving.

## Behavior changes

- A bind whose provider connection is closing now defers silently (matching unknown-type behavior) instead of erroring, and recovers when the provider re-registers the type.
- An async provider rejection parks the bind for replay on the next type registration; the error is still forwarded to the originator.

## Tests

- `register_trigger_transport_failure_parks_and_next_type_registration_activates` — park + heal through a real closed-channel `WorkerConnection`, exercising the marker classification end to end
- `undeliverable_reregistration_of_live_binding_ends_parked_not_split` — live-XOR-parked invariant on an undeliverable re-registration
- `rejected_reregistration_keeps_existing_live_binding_tracked` — a rejection never untracks the previous live activation
- `test_register_trigger_propagates_registrator_error` — rejections still fail and are never parked
- `unregister_live_trigger_clears_stale_pending_intent` — explicit unregister clears a raced re-park
- `test_trigger_registration_result_error_forwards_parks_and_replays` — async rejection forwards the error, parks, and replays to the reconnected provider over real connection channels

`cargo test -p iii --lib`: 1964 passed. Changed-line coverage verified with `cargo llvm-cov` (all functional lines hit; the only unhit line is the empty fall-through of the concurrent-reap guard, a cross-thread race window with no await points).

Fixes MOT-4081

## Summary by CodeRabbit

* **Bug Fixes**
  * Failed trigger registrations are now safely parked and replayed after the trigger type/provider is available again.
  * Undeliverable provider-connection issues are treated as deferred/parkable instead of being lost permanently.
  * Definitive provider rejections still surface as errors and are not retried.
  * Trigger unregistrations are now race-safe, preventing late failures from resurrecting bindings.
* **Tests**
  * Updated and expanded coverage for park-and-replay behavior and concurrent unregister/replacement vs. late rejections.

### DIFF
--- a/engine/src/engine/mod.rs
+++ b/engine/src/engine/mod.rs
@@ -626,7 +626,20 @@ impl Engine {
                     return Ok(());
                 }
 
-                self.trigger_registry.triggers.remove(id);
+                // Re-park instead of dropping; park only when this call
+                // removed it — a concurrent disconnect GC may have already
+                // reaped it, and parking then would resurrect a dead binding.
+                if let Some((_, failed)) = self.trigger_registry.triggers.remove(id) {
+                    tracing::warn!(
+                        trigger_id = %id,
+                        trigger_type = %stored_trigger_type,
+                        function_id = %stored_function_id,
+                        "Trigger registration rejected by provider; parked as pending until the trigger type re-registers"
+                    );
+                    self.trigger_registry
+                        .pending_triggers
+                        .insert(failed.id.clone(), failed);
+                }
 
                 let Some(originator_id) = originator_id else {
                     tracing::debug!(
@@ -3503,7 +3516,7 @@ mod tests {
     }
 
     #[tokio::test]
-    async fn test_trigger_registration_result_forwards_error_to_originator() {
+    async fn test_trigger_registration_result_error_forwards_parks_and_replays() {
         ensure_default_meter();
         let engine = Engine::new();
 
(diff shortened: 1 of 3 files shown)


anthonyiscoding:
Can we change this from "park" to "optimistic registration" and "store"? Park has no meaning outside of agile/scrum.

guibeira (dismissed):

guibeira (approved):

PR #2031, opened 2026-07-31
feat(website): make the hero headline "an agentic development environment"

anthonyiscoding:
Replaces `build unreasonably <simple> software` with a claim about what iii actually is, carried by two carousels instead of one.

```
an agentic     hero-subj, 8s, ink
development    hero-desc, 3s, accent
environment    static, mute
```

`hero-subj` rotates between **an agentic** and **a human**. `hero-desc` rotates through development, production, testing, staging, debugging, orchestration, integration, and deployment.

## Behavior

- The page holds the full phrase **an agentic development environment** for eight seconds on arrival, so it reads as a statement before anything moves. Both carousels then start together at that mark.
- The subject turns slowly, on a period that is not a multiple of the noun's, so the two keep landing on fresh pairings rather than marching in lockstep.
- Each carousel sits alone on its own `.line`, so a long word like `orchestration` cannot rewrap the line above it. Only the changing word moves.

## Implementation

The old inline rotation loop became a `rotateWords(el, words, period)` helper that both carousels call. The fade-and-lift transition on `.desc` is unchanged. The only new CSS is `.desc.plain`, which keeps the second rotating word in ink so the line retains a single accent.

A third carousel (`for building / architecting / experimenting / vibing …`) is included commented out in both files. Three rotating lines at once was too much motion, but the copy is worth keeping to hand, and `rotateWords` no-ops on a missing element so uncommenting the markup alone would also work.

Branched off `main` rather than stacked onto #2030; the two changes are independent.

## Summary by CodeRabbit

* **Enhancements**
  * Updated the hero headline to highlight an agentic development environment.
  * Added rotating hero text featuring environment-related terms and alternating agentic/human messaging.
  * Improved headline styling for clearer, more consistent text presentation.

### DIFF
--- a/website/src/components/landing/scripts/02-page-behaviors.html
+++ b/website/src/components/landing/scripts/02-page-behaviors.html
@@ -71,23 +71,36 @@
     });
   });
 
-  // ============= Hero rotating descriptor =============
-  const rotatingDescriptors = [
-    'simple','scalable','durable','extensible','fast',
-    'composable','observable','reliable','interoperable'
-  ];
-  const heroDesc = document.getElementById('hero-desc');
-  if (heroDesc){
-    let rotIdx = 0;
-    setInterval(() => {
-      heroDesc.classList.add('is-anim');
+  // ============= Hero rotating descriptors =============
+  // The headline reads "an agentic development environment" on arrival and
+  // holds it long enough to be read as a claim before anything starts moving.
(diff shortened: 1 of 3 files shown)


anthonyiscoding:
Superseded: the hero headline change now lives on #2030 alongside the explainer sections.

PR #2034, opened 2026-08-03
feat(compose): worker-compose daemon — declare a project, start it as a graph, supervise it

guibeira:
Targets `feat/namespare` (#1984): compose is the first real consumer of namespace routing, and does not build without it.

## What this is

`iii compose` runs a declared set of workers as one project. A `worker-compose.yaml` names the containers, their dependencies and how each one starts. The daemon validates the declaration, resolves configuration, starts the graph in order, and supervises what it created.

There is one command, and it does not background itself:

```bash
iii compose --ns dev
```

That connects to an engine and serves the `compose::*` functions in the foreground. Every operation on a project is a normal trigger, addressed to one daemon:

```bash
iii trigger compose::validate --namespace dev file=./worker-compose.yaml
iii trigger compose::up       --namespace dev file=./worker-compose.yaml
iii trigger compose::status   --namespace dev file=./worker-compose.yaml
iii trigger compose::down     --namespace dev file=./worker-compose.yaml
iii trigger compose::list     --namespace dev
iii trigger compose::stop     --namespace dev
```

The foreground is the shape a process supervisor already wants. systemd, launchd or a shell redirect gives restart-on-failure and log rotation, so compose does not have to grow either.

A greenfield crate (`crates/iii-compose`). It shares no runtime code with `crates/iii-worker`, whose lifecycle system it is meant to replace rather than extend.

## Why it needs namespaces

Two namespaces are in play, and they answer different questions.

**The daemon's namespace** is its address. The worker name is always `compose`, so the engine leases `(namespace, compose)` to one connection. `--ns` is what tells two daemons on one engine apart, and it is what an operator types into `--namespace` to reach exactly one of them. A second daemon claiming a namespace that is taken is refused at registration with `DAEMON_ALREADY_SERVING`. Without `--ns` the daemon generates one and prints it, because a shared well-known default is the collision the namespace exists to prevent.

**The project's namespace** is where its containers register. It comes from the `namespace:` field in the compose file, and a project that declares none lands in `default`. Nothing about the file's path enters it. A value outside `[a-z0-9_-]` is refused with `INVALID_NAMESPACE` instead of being rewritten to fit, so what the file declares is what the operator types.

Readiness is `(namespace, container)` as seen in `engine::workers::list`. It is the pair, never the name alone, or another project's worker would satisfy it.

## The parts

**Declaration.** Strict schema (unknown fields are errors, duplicate container keys rejected), dependency graph with cycle paths in declaration order, Kahn ordering. `compose::validate` needs nothing but the file: it does not take the project on, and `package://` containers are reported under `deferred_packages` rather than resolved. So a CI job that only validates leaves the daemon owning nothing.

**Identity.** A project is its compose file and nothing else names one. The same file reached twice is the same project however the path was spelled, so there is no second identity to keep in sync. The container key is the worker name, and it wins over the manifest's `name`: the key is what reaches the child as `III_WORKER_NAME`.

**Spawn contract.** The child's environment is built, not inherited: a host baseline, then `env_file` in declared order, then `environment`, then five reserved variables the daemon owns (`III_URL`, `III_NAMESPACE`, `III_WORKER_NAME`, `III_CONFIG`, `III_CONFIG_NAME`). A stale `III_URL` in the operator's shell can never point a child at the wrong engine, and a project starts the same way whoever launched it. Declaring a reserved variable is caught at validation, in an `env_file` as well as in `environment`.

**Configuration.** A container names the entry it owns with `config_name`, and `config_override` is merged on top. Precedence is what the package ships, then what the configuration worker holds, then the override. The merged result is written to an owner-only file and passed as `III_CONFIG`, so the worker reads configuration the way it always does. A container that declares `config_name` does not start when the fetch fails.

**Supervision.** Each child leads its own process group, so teardown reaches grandchildren. A pid is never signalled without a birth-identity match, because pids are recycled. After readiness a poll loop notices an exit nobody asked for, fires `post_run`, and cascades to the transitive dependents in reverse-topological order. Nothing is restarted: v1 has no restart policy. On restart, survivors are adopted back, because a process left out of the children map is one `down` walks past while reporting success.

**Registry.** `package://` resolves against the registry, verifies the sha256 before anything reaches the final directory, and caches by `(name, version, target)`.

**State.** One place per machine, `~/.iii/compose` or `$III_COMPOSE_STATE_DIR`, keyed by daemon namespace and a derived project directory. It holds the child records, the resolved configuration files, and what each container printed while starting. Capture stops at readiness: before it, the container's own output is the only account of itself; after it, the worker's logging is the record and compose keeps no second copy.

**Diagnosis over silence.** A worker that ignores `III_NAMESPACE`, one that registers under its own name, and one whose functions were filed in another namespace are all healthy processes that readiness can never satisfy. Each is named the moment it can be known, instead of twenty seconds of nothing followed by `STARTUP_TIMEOUT`. Codes cross the wire, stable and documented.

## Verification

- **134 unit and integration tests** in the crate, clippy clean.
- **11 e2e tests** (`engine/tests/compose_daemon_e2e.rs`) against a real in-process engine: `compose::*` answers only in the daemon's namespace, a request aimed at another daemon is refused, one daemon holds several projects, one file is one project however it is spelled, a child that never registers times out and rolls back, a duplicate namespace is refused while two distinct ones both serve, and a configuration that cannot be read stops the container.
- **The `compose-test-matrix` CI job** runs the crate suite on Linux, macOS and Windows, because supervision is platform code.
- **A scenario suite** run outside this repo against real binaries: polyglot projects, registry installs, two projects side by side, daemon killed and restarted, engine restarted, container crashed after readiness. Several defects in this PR were found by running those and none by the unit tests: a targeted `down` in the wrong order, orphaned grandchildren, a stale worker satisfying readiness, hooks shredding the spinner, and teardown returning on a dead leader while its worker kept serving.

## Known gaps

- **Windows runs the portable tests only.** Job Objects replace process groups and `CTRL_BREAK` replaces `SIGTERM`, but the unix suites are cfg-gated, so the Windows job proves compilation of the job-object path and little of its behaviour.
- **macOS never adopts.** Without `libproc` there is no birth fingerprint, so every live pid is unverifiable. That is safe, but a restart there reports children for manual cleanup.
- **No e2e with a container that reaches ready.** That needs a fixture binary speaking the SDK. The scenario suite covers it, outside CI.
- **Two copies of one project collide.** The namespace comes from the file, so running the same project twice needs two files, or one edited between runs.
- **No restart policy, and no way to declare a failure's blast radius.** A failure during `up` ends the operation; a failure after readiness takes the dependents. Neither is configurable in v1.
- **Only `path://` and `package://` resolve.** There is no container runtime behind `image://`, which is out of scope for v1.

## Summary by CodeRabbit

* **New Features**
  * Added the `iii compose` command for validating and managing multi-worker projects.
  * Supports dependency-aware startup, shutdown, restart, status, listing, validation, configuration, environment files, lifecycle hooks, and package-based workers.
  * Added persistent recovery, process supervision, worker bundles, and cross-platform support.
* **Documentation**
  * Added CLI reference, usage guidance, schema details, lifecycle behavior, validation rules, and limitations.
* **Tests**
  * Added comprehensive unit, integration, end-to-end, recovery, lifecycle, and cross-platform coverage.


crates/iii-compose/src/lib.rs
```diff
@@ -0,0 +1,381 @@
...
+    client.shutdown();
+    answered
+}
+
+/// Follows a detached daemon's log until interrupted.
+///
+/// Attaching is a view, not a reparenting: the daemon keeps its own process
+/// group and keeps running when this returns. What comes back is what it says
+/// from here on, plus the tail of what it already said.
+async fn attach() -> Result<()> {
```
anthonyiscoding: ### `--attach` is not attaching, and `-d` creates a log-management problem

`--attach` reads `~/.iii/compose/daemon.log`, seeks back 8 KiB, and tails it (`crates/iii-compose/src/lib.rs:317`). There is no connection to the daemon process: no stdio takeover, no signal forwarding, no reparenting. Ctrl-C in an "attached" terminal returns from the tail and leaves the daemon running. The flag name promises a session and delivers `tail -f`, which is the kind of gap an operator only discovers when they need it to be true.

`-d` is the reason that log file exists. It re-execs the daemon with stdout and stderr pointed at a single append-only file (`lib.rs:182`), and from that moment we own a log-management problem we have not solved: no rotation, no size cap, no retention, one file shared by every project on the machine. It grows without bound, and the append-is-never-truncate choice (correct on its own terms) makes that worse rather than better. Solving it properly means rotation, and rotation means either a dependency or our own copy of logrotate's edge cases.

Proposal: drop both flags for now. `iii compose` serves in the foreground, which is the shape every process supervisor already wants, and daemonizing is documented rather than implemented:

```bash
iii compose >> /var/log/iii-compose.log 2>&1 &
```

```ini
[Unit]
Description=iii compose
After=network.target

[Service]
Type=simple
ExecStart=/usr/local/bin/iii compose
Restart=always

[Install]
WantedBy=multi-user.target
```

That hands rotation to logrotate or the journal, gives us restart-on-failure for free, and removes the re-exec, the detach guard (`III_COMPOSE_DETACHED`), the pid-matched readiness poll in `daemon_is_serving`, and the tail loop. `compose::stop` and the signal handlers already cover the stop path either way.

Worth noting for whoever writes the unit file: `Type=notify` needs the process to call `sd_notify`, which compose does not do, so it would be killed at `TimeoutStartSec`. `Type=simple` is the correct choice until we ship readiness notification.

crates/iii-compose/src/remote.rs
```diff
@@ -0,0 +1,180 @@
...
+use crate::{daemon::Daemon, error::ComposeError};
+
+/// Payload accepted by every `compose::*` function. All fields optional: the
+/// bare call is the common one.
+#[derive(Debug, Clone, Default, Deserialize, schemars::JsonSchema)]
+#[serde(default)]
+pub struct ComposeRequest {
+    /// Which project. Chosen by the caller on the first `up` and used to reach
+    /// it ever after: one daemon holds many, and this is what tells them apart.
+    pub id: Option<String>,
```
anthonyiscoding: ### Call this field `namespace:` rather than `name:`

`name:` in the compose file has exactly one effect: it is the namespace the project's containers register in. The value is what an operator types into `iii trigger --namespace` and into a `worker.trigger` call, so the field name is worth spending a word on.

`name` reads as a display label, the project's human title, which is a reasonable thing for a file to have and is not what this is. `sanitize` rewriting the value reinforces that reading. `namespace:` states what it sets, matches the flag the value ends up in, and matches the vocabulary the rest of the engine uses for the same coordinate.

`id` is unaffected. It addresses the project through `compose::*` and is a separate concern from where the containers register; the rename is only about the field that feeds `III_NAMESPACE`.

crates/iii-compose/src/namespace.rs
```diff
@@ -0,0 +1,146 @@
...
+    candidates
+        .into_iter()
+        .flatten()
+        .map(str::trim)
+        .find(|value| !value.is_empty())
+}
+
+/// Namespaces travel into queue names, metric labels and log lines, so the
+/// derived form stays inside `[a-z0-9_-]`.
+fn sanitize(value: &str) -> String {
```
anthonyiscoding, line 58: ### Validate `name` against the character set instead of rewriting it

`sanitize` accepts anything and quietly rewrites it: `My Shop!` becomes `my-shop`, a run of illegal characters collapses to one `-`, leading and trailing dashes are trimmed, and a name made entirely of illegal characters becomes the invented literal `project`.

Recommend rejecting instead. Keep the same character set, `[a-z0-9_-]`, and fail when `namespace:` (or `namespace=`) does not match it.

The name is the one string an operator has to be able to predict and type: it is the value typed into `iii trigger --namespace` and into `worker.trigger` calls. A value that is rewritten on the way in is one they cannot type back. Worse, the rewrite is lossy, so `My Shop!` and `my/shop` both resolve to `my-shop` and become the same project without either declaration saying so. The `project` fallback has the same problem in a sharper form, because it names a project after nothing the file contains.

Rejecting is also cheaper than it looks: this runs at load time, before any container starts, and the error can quote the offending name and the permitted set. A misspelled name then fails the way an unknown compose key already does, which is the strictness the rest of `config.rs` commits to.

crates/iii-compose/src/config.rs
```diff
@@ -0,0 +1,577 @@
...
+
+    for dependency in &raw.depends_on {
+        if dependency == key {
+            return Err(ComposeError::SelfDependency {
+                container: key.to_string(),
+            });
+        }
+    }
+
+    if raw.config_name.is_some() && raw.config_uri.is_some() {
```
anthonyiscoding, line 200: ### Keep `config_uri` and drop `config_name`

`config_name` and `config_uri` are two spellings of the same value. Both collapse to one string in `Container::config_name`, and the only difference is that the uri form states where the value comes from while the name form assumes it.

Recommend keeping `config_uri` alone. The scheme is the flexibility: `worker://configuration/get/<name>` is the only accepted form today, and `file://`, a direct adapter, or another worker can be added later under the same key. `config_name` cannot grow that way, so every future source would arrive as one more top-level field, and each new pair would need its own conflict rule.

Dropping it also removes machinery that exists only because there are two keys. `CONFLICTING_CONFIG_SOURCE` has nothing left to guard, the three-arm match over `(config_name, config_uri)` becomes a single `parse_config_uri` call, and `UNSUPPORTED_CONFIG_URI` becomes the one error that answers "where does this container's configuration come from".

Worth doing now rather than after v1 ships: while `config_name` exists it is the shorter of the two, so it is what people will write, and removing it later is a breaking change to every compose file.

crates/iii-compose/src/manifest.rs
```diff
@@ -0,0 +1,204 @@
...
+        return Err(ComposeError::MissingWorkerDirectory {
+            container: key.to_string(),
+            path: dir.clone(),
+        });
+    }
+
+    let manifest = read_manifest(dir)?;
+    // A manifest that renames the container would split identity between the
+    // compose graph and the engine registration.
+    if let Some(name) = manifest
```
anthonyiscoding, line 87: ### `worker-compose.yaml` should win over `iii.worker.yaml` everywhere, including the name (ie. namespace)

Precedence between the two files is inconsistent today. For the start command the compose file wins: `run` overrides the manifest's `scripts.start`, and the comment right below gives the reason, the compose file is the operator's file and the manifest is the worker author's default. For the name the same disagreement is a hard error instead.

Make the rule uniform. The container key overrides the manifest's `name`, the same way `run` overrides `scripts.start`, and `MANIFEST_NAME_MISMATCH` goes away. The mechanism is already in place: the key is passed to the child as `III_WORKER_NAME`, so a worker that honours the reserved contract registers under the compose key with no further work.

State it as the general rule rather than a per-field decision, because compose will read more of the manifest over time and each new field would otherwise need its own answer: **where the two files describe the same thing, `worker-compose.yaml` wins and the manifest is the default.** An operator deploying a worker they did not write must be able to override the author's declaration without editing a vendored directory.

Keep the readiness check as the backstop. A worker that hardcodes its name in code ignores `III_WORKER_NAME` and still lands under the wrong name, and `WORKER_NAME_MISMATCH` already catches that and reports the name it took. That path covers the case the manifest cannot predict, which is the case worth an error.

crates/iii-compose/src/spawn.rs
```diff
@@ -0,0 +1,283 @@
...
+        .filter_map(|name| {
+            std::env::var(name)
+                .ok()
+                .map(|value| (name.to_string(), value))
+        })
+        .collect();
+
+    env.extend(ctx.user_env.clone());
+
+    env.insert("III_URL".to_string(), ctx.engine_url.to_string());
```
anthonyiscoding, line 118: ### Invert this precedence: a declared environment variable supersedes static configuration

The reserved four are forced on top of the container's own environment here, and declaring one in `environment` or an `env_file` is refused outright with `RESERVED_ENV_OVERRIDE`. That is the wrong way round for three of the four. An environment variable is the more specific and the more immediate statement, so it should win over a value derived from static configuration.

Recommend shrinking `RESERVED_ENV` to `III_URL` alone. `III_NAMESPACE`, `III_WORKER_NAME` and `III_CONFIG` become defaults: compose computes them as it does now, then the container's `env_file` and `environment` are applied on top.

- `III_NAMESPACE` is the clearest case. It is derived from `namespace:`, a project-wide declaration, and no container can say it registers somewhere else. A worker that legitimately belongs in another namespace, a shared one addressed by two projects for instance, cannot express that at all today, and the workaround is a second compose file.
- `III_WORKER_NAME` is the container key, and an operator deploying a worker they did not write has no other way to change the name it registers under.
- `III_CONFIG` pointing at a file the operator wrote instead of the merged one is a reasonable thing to want. Compose can skip writing its own when the variable is set.

`III_URL` is the one that stays owned by the daemon. Readiness runs over the daemon's own connection, so a container pointed at another engine is invisible to this daemon no matter how long it waits, and the failure would surface as a `startup_timeout` over a healthy worker. Two engines means two daemons, and `RESERVED_ENV_OVERRIDE` keeps saying so for this key.

One thing has to follow the other three, or an override turns into a silent timeout: readiness watches `(project_namespace, container)`. When a container overrides `III_NAMESPACE` or `III_WORKER_NAME`, the readiness check, the `ChildRecord`, and `compose::status` all have to use the overridden pair. Otherwise compose waits in the namespace it derived, the container registers in the one it was told, and the result is `WORKER_IGNORED_NAMESPACE` or a full `startup_timeout`.

crates/iii-compose/src/manifest.rs
```diff
@@ -0,0 +1,204 @@
...
+        let worker_dir = match &container.worker {
+            WorkerSource::Package { .. } => {
+                deferred_packages.push(key.clone());
+                continue;
+            }
+            WorkerSource::Path { dir, .. } => dir,
+        };
+        // Env files are read at spawn time, but a missing one must fail here:
+        // finding out at `up` means half the graph is already running.
+        for env_file in &container.env_file {
```
anthonyiscoding, line 159: ### `validate` checks that env files exist and never reads them

This loop confirms each `env_file` is present and stops there, so every check that depends on the file's contents happens at `up` instead. `environment` is validated at parse time, which means the same mistake is reported at two different moments depending on which of the two an operator used.

The `RESERVED_ENV_OVERRIDE` check is the case that exists today: a reserved key inside an `env_file` is only found when `resolve_user_env` runs during `start_one`, and by then earlier containers in the graph are already running and get rolled back. A CI job that runs `compose::validate` reports the project as fine.

Recommend parsing the env files here as well and reporting whatever the parse can decide offline. `parse_env_file` needs no engine and no network, so this costs one read per file at validation time and moves the error to the same place the `environment` version of it is already reported.

The remark applies past this one check: any rule that holds for `environment` should be evaluated for `env_file` at the same stage, otherwise `validate` keeps a blind spot that grows every time a rule is added.

crates/iii-compose/src/dag.rs
```diff
@@ -0,0 +1,165 @@
...
+                queue.push_back(dependency.clone());
+            }
+        }
+    }
+    closure
+}
+
+/// Local containers that must stop before `key` can stop: its transitive
+/// dependents, nearest first.
+pub fn transitive_dependents(file: &ComposeFile, key: &str) -> Vec<String> {
```
anthonyiscoding, line 148: ### Discovery order is not a reverse-topological order, so a deep cascade stops in the wrong sequence

This returns dependents in breadth-first discovery order, and `down` uses that list as the teardown order. For a cascade deeper than one level the result violates the rule teardown exists to enforce.

Project: `db` <- `api` <- `web`, plus `reports` depending on `db`.

```
api exits -> stop order ["web", "api"]                   correct
db  exits -> stop order ["api", "reports", "web", "db"]  api stops before web
```

`web` depends on `api`, so in the second case `web` is left calling a stopped dependency for the rest of the teardown. That is the state the dependents-first rule prevents, and it applies to both `compose::down container=db` and the supervisor path, which routes through the same `down`.

The regression test at `lifecycle.rs:630` passes because it targets `api`. That is a depth-1 cascade, where discovery order and reverse-topological order agree. Any target with a dependent chain of two or more below it diverges.

The fix keeps the existing topological sort instead of a second ordering rule: take `start_order()`, reverse it, and filter to the dependent set plus the target. Worth extending the test to a depth-2 target, since the current one cannot fail.

crates/iii-compose/src/state.rs
```diff
@@ -0,0 +1,255 @@
...
+}
+
+impl StateStore {
+    /// `~/.iii/compose/<daemon-id>`, or `$III_COMPOSE_STATE_DIR/<daemon-id>`
+    /// when the operator relocates it (a read-only home, a tmpfs, a test).
+    pub fn for_daemon(daemon_id: &str) -> Result<Self> {
+        if let Some(root) = std::env::var_os("III_COMPOSE_STATE_DIR") {
+            return Ok(Self::at(PathBuf::from(root).join(daemon_id)));
+        }
+        let home = dirs::home_dir().ok_or(ComposeError::StateDirUnavailable)?;
```
anthonyiscoding: ### Keep a project's state and logs beside its `worker-compose.yaml`, not under a global `~/.iii`

State, resolved config, and container logs all land under `~/.iii/compose/<id>/` today, keyed by an id the caller invented. The compose file is the project, so the project's directory is where its state belongs: something like `<compose dir>/.iii/`, next to the file that declares it.

Several of the awkward parts of the current design are consequences of the global location:

- The compose-file binding exists to stop one id from adopting another project's children, and `STATE_BINDING_MISMATCH` is the error for it. State that lives beside the file cannot point anywhere else, so the binding is structural instead of checked.
- Package name collisions and cross-project confusion in `~/.iii/compose/packages/` are the same problem in a different directory.
- Logs are the sharper case. A container's output is written under the project's state directory and never served, so a global path means an operator has to know the id to find output for a project they are standing inside. Beside the compose file it is `ls .iii/logs/`.
- Removing a project currently leaves its state behind unless someone knows to clear `~/.iii/compose/<id>`. In-tree state is removed with the checkout.
- Two checkouts of the same repository are two projects with two state directories, with no id juggling and no `III_COMPOSE_STATE_DIR` override to arrange it.

`III_COMPOSE_STATE_DIR` already exists for the tests, which is a signal that the global path is inconvenient before any user has seen it.

Worth deciding now, because it interacts with the two other notes on this review: dropping `-d` removes the last reason for a machine-global log file.

The docs section that describes the current layout is `next/using-iii/compose#where-compose-keeps-state`, and it needs to change with this.

crates/iii-compose/src/lifecycle.rs
```diff
@@ -0,0 +1,661 @@
...
+                records.insert(key.clone(), record);
+                started.push(key.clone());
+                results.push(ContainerResult {
+                    container: key.clone(),
+                    state: ChildStatus::Ready,
+                    changed: true,
+                    error: None,
+                });
+            }
+            Err(error) => {
```
anthonyiscoding, line 156: ### One container failing should be a declarable policy, and the two failure paths currently answer differently

What one failure does depends on when it happens, and the difference is easy to miss.

**During `up`.** The first container that fails ends the operation. Everything this operation started is rolled back in reverse order, and every container after it in the start order is never attempted. So on a first `up` of a five-container project, a failure in the last container leaves the whole project down, including containers that have no relationship to the one that failed.

**After a container is ready.** The supervisor takes the dead container's transitive dependents down and leaves everything else running. A container that nothing depends on takes nothing with it.

Both are defensible on their own, and together they mean the same declaration produces two different blast radii depending on timing. A `mailer` that nothing depends on ends the whole start when it fails at `up`, and is contained when it fails a minute later.

This looks like something the file should state rather than something compose decides. Roughly:

```yaml
containers:
  mailer:
    worker: path://./workers/mailer
    required: false        # its failure does not fail the operation
  api:
    worker: path://./workers/api
    restart: on-failure    # what the supervisor does when a ready container exits
```

Two separate questions worth separating: whether a container's failure fails the operation, and what happens when a ready container exits. The second is where a restart policy belongs, and there is none today, so any crash is permanent until someone runs `up` again.

No strong opinion on the defaults. `required: true` and no restart match today's behaviour and are the conservative starting point. The part worth fixing regardless of the knob is the asymmetry: whatever the policy is, it should read the same at start time and at run time.

anthonyiscoding (changes requested):
Looking good, definitely have some runtime/behavior and configuration specifics to figure out first though.

Also I added some additional docs to this branch that I felt were needed for using-iii/compose and understanding-iii/compose. They'll need further updating if behavior or other specifics change as I wrote them to currently reflect how the branch is at this moment.

crates/iii-compose/src/dag.rs
guibeira: Confirmed and fixed in `e82c35e7`.

I reproduced it before changing anything, with your exact example — `db <- api <- web` plus `reports <- db`:

```
stop order for db: ["api", "reports", "web", "db"]
```

`api` stops while `web` is still calling it.

I took the fix to `transitive_dependents` itself rather than to the call site, so both consumers are covered by one change: the targeted `down`, and the supervisor cascade in `project.rs:213`, which logs the same list. It now reverses `start_order()` and filters to the dependent set, which is your suggestion — the topological sort was already there, and this removes the second ordering rule rather than adding to it.

Worth noting the comment at `lifecycle.rs:218` asserted the property that was false:

```rust
// `transitive_dependents` already returns them nearest-first, so the
// target goes last and the list needs no reversing.
```

That is now true rather than deleted.

On extending the test: you were right that the existing one could not fail. It targets `api`, a depth-1 cascade, where discovery order and reverse-topological order agree. The new case targets `database` with `web` two levels above it. It asserts relative positions instead of the whole list, because the order between `api` and `reports` is not part of the contract — only that `web` precedes `api` and the target goes last.

Verified against the smoke suite as well: the six scenarios that exercise teardown (`03`, `05`, `11`, `12`, `16`, `17`) pass, 87 checks. Worth flagging one limit: those scenarios are all depth-1, so the smoke suite would not have caught this on its own. The new unit test is what covers it.

crates/iii-compose/src/manifest.rs
guibeira: Accepted, fixed in `71b5ba58`.

Two things came out of it.

**Secrets.** `resolve_user_env` reads at spawn time on purpose ("env files hold secrets, and holding them in memory for the daemon's whole life buys nothing"). So the check reads, inspects the key names, and drops the contents inside the function — the values do not outlive the call. Reasoning is on the function, since the way this regresses is someone caching the parse later as an optimisation.

**A bigger instance of the same bug.** The loop sat *after* the branch that defers `package://` containers, so every registry worker was exempt from all of it — including the missing-file check whose comment two lines down promises it fails before anything starts. It now runs ahead of that branch; nothing in it needs the worker on disk.

The general rule is on `check_env_files` rather than only in this thread, which is how the next rule would grow the blind spot again.

Three tests: reserved key in an env file, the same for a `package://` container plus its missing-file case, and an ordinary env file that must still validate. Ran them against the pre-fix code first — the two regression tests fail there, so they are not just agreeing with the new code.

One I left alone: `parse_env_file` skips malformed lines silently, so a line missing its `=` disappears and the worker fails later with no trace. Cheap to report now that validate reads the file, but it would fail compose files that work today — separate change if you agree it's a gap.

crates/iii-compose/src/namespace.rs
guibeira: Accepted, fixed in `849669de`.

Ran the values through a live daemon first. It was lossier than the two collisions you named — four spellings, not two:

```
name: "My Shop!"   ->  namespace: 'my-shop'
name: "my/shop"    ->  namespace: 'my-shop'
name: "my shop"    ->  namespace: 'my-shop'
name: "MY-SHOP"    ->  namespace: 'my-shop'
name: "!!!"        ->  namespace: 'project'
```

**The part that made this easy to accept.** The project already rejects — on the other path, with a different rule. `cli.rs` checked `--ns` for emptiness, path separators and `.`/`..`, and never sanitized what passed. So `--ns 'My Shop!'` registered a daemon under the literal namespace:

```
name='compose'  namespace='My Shop!'
```

One string, two namespaces, depending on which of the two the operator used to say it — and the checked path was the one accepting spaces. Both now go through the same `namespace::check`, so this aligns the file with the rule the flag was already reaching for, and fixes the flag.

Uppercase gets its own reason rather than the generic charset one, since lowercasing is the rewrite an operator is likeliest to assume happened: *"namespaces are lowercase, and are not lowercased for you"*.

Breaking, as you'd expect: a `name:` outside `[a-z0-9_-]` now fails to load. None of the 23 names in the smoke suite were affected.

Also removed `daemon_worker_name` and `UNNAMED_DAEMON` — left over from when a daemon had a name apart from its namespace, and called by nothing but their own tests.

Smoke coverage in `compose-smoke-tests@fd674b0`: two fixtures, plus a check that `--ns 'My Shop!'` is refused *and* that no daemon starts.

crates/iii-compose/src/manifest.rs
guibeira: Accepted, fixed in `9c4380a2`.

The rule is now stated on `resolve_start` rather than decided per field, as you suggested.

**One detail that made the decision easy.** The comment defending the check was already false:

```rust
// A manifest that renames the container would split identity between the
// compose graph and the engine registration.
```

Compose passes the container key as `III_WORKER_NAME` (`spawn.rs:120`), so a worker honouring the reserved contract registers under the key whatever its manifest declares. There was no split identity to prevent. The error refused a configuration that already worked.

`Manifest.name` read nothing else, so it went with the check.

`WORKER_NAME_MISMATCH` stays as the backstop, as you proposed. It covers the case the manifest cannot predict, and it reports the name the worker actually took.

Breaking: `MANIFEST_NAME_MISMATCH` is removed. It was also documented in `using-iii/compose.mdx`, so that section and its rendered artifact are updated.

Smoke coverage in `compose-smoke-tests@ed04551`: `web` in `02-daemon` now starts from a manifest that declares `name: web-frontend`. One container carries both halves of the rule — the manifest supplies the start command because the compose file declares no `run`, and the key wins over the name. The check fails if `web-frontend` appears in the worker list.

I ran the rewritten unit test against the pre-fix code first. It fails there.

crates/iii-compose/src/remote.rs
guibeira: Accepted, done in `1e4076ff`. Hard break, no alias — nothing has shipped, so there is no file anywhere to migrate.

Two things since you wrote this made the case stronger than it was.

**`id` is gone** (`c606647f`, about seven hours after the review). You noted `id` was unaffected; there is no longer an `id` to be unaffected. `namespace:` is now the only thing the file declares about the project, and it was not a name.

**`sanitize` is gone** (`849669de`, your other note). So the field can no longer be read as a label that becomes a namespace after cleaning. The value arrives verbatim. It *is* the namespace.

**One thing that fell out of the rename.** The validation report carried both `project` and `namespace`. `project` was `name.unwrap_or(namespace)`, so with the rewriting removed the two always held the same string:

```
namespace: 'orders'  ->  project='orders'   namespace='orders'
namespace: absent    ->  project='default'  namespace='default'
```

The rewrite was the only thing that ever made them differ. I dropped `project` — a second field carrying one value under another word is one that drifts. That is a second change to `compose::validate` you did not ask for, so say the word if you would rather keep it.

The strict schema refuses the old spelling and points at the right key:

```
unknown field `name`, expected one of `namespace`, `startup_timeout`, `stop_timeout`, `containers`
```

There is a test pinning that. Not as a migration aid, but so `name:` does not come back later as a convenience alias with two keys on one coordinate.

Docs, the `compose-file.md` tech spec and 44 files in the smoke suite follow (`compose-smoke-tests@76cc531`). `iii.worker.yaml` keeps its `name:` — that one belongs to the worker author — and so does the configuration worker's storage, where `name:` is the entry's display name beside `id:` and `value:`.

crates/iii-compose/src/config.rs
guibeira: Agreed on the problem, inverted on the answer: `config_uri` is gone and `config_name` stays (`a2d2eb94`).

You were right that two spellings of one coordinate is not affordable, and everything you wanted from removing one lands: `CONFLICTING_CONFIG_SOURCE` and `UNSUPPORTED_CONFIG_URI` are gone, `parse_config_uri` is gone, and the three-arm match is now `raw.config_name.clone()`.

**Why the name rather than the uri.** How a configuration is read and stored is the configuration worker's business — it has an adapter for exactly that, `fs` today, and it is itself configurable. A scheme in the compose file would be compose describing transport it does not own, and a `file://` form would contradict the adapter that actually decides where entries live. The tech spec already committed to this:

> Names resolve through the **configuration worker**, whose adapter decides storage — local fs today, secrets manager or a separate iii instance in cloud. File paths are not the contract; names are, so the same compose file works when storage moves.

So the extensibility the scheme bought was extensibility into another worker's concern.

**And the field stopped being a source after you wrote this.** `III_CONFIG_NAME` landed in `5d92443b`, the day after the review. The value now drives three things:

```
config_name: state-shop-a
      ├─► fetch_config("state-shop-a")        read as the base
      ├─► publish_config("state-shop-a", v)   write the merged result back
      └─► III_CONFIG_NAME=state-shop-a        tell the child which entry is its own
```

It names the entry the container owns, not where to get something. A `get` verb describes the first arrow and misleads about the other two.

If a base ever needs to come from somewhere the configuration worker does not reach, that seems like a separate field rather than a scheme on this one — at that point it genuinely *is* a source, and can be named like one.

Docs, both tech-spec pages and the smoke suite follow (`compose-smoke-tests@253f165`). Scenario 18 loses its `aliased` container, which existed only to show the two spellings reaching one entry. There is a test pinning that `config_uri` is refused by the strict schema with `config_name` named in the error, so it cannot come back quietly as an alias.

crates/iii-compose/src/spawn.rs
guibeira: Declining the inversion, but you were right that the rule was defended as a block and not per key. Each of the five now earns its place separately (`f7083550`), so a sixth is a decision rather than a habit.

**`III_URL`** — agreed, and for your reason. That is the one where the failure is undiagnosable rather than merely wrong.

**`III_WORKER_NAME`** — already solved by `9c4380a2`. The container key is the registered name; `MANIFEST_NAME_MISMATCH` was what stopped an operator changing it, and that is gone.

**`III_CONFIG`** — this one splits a delivery in half. Compose merges the configuration, writes it to the file *and* publishes the same value to the entry named by `config_name`. Pointing the child at a hand-written file does not stop the publish, so the worker reads one value while the configuration worker holds another, with nothing to say which is in force. `III_CONFIG_NAME` is the same hazard from the other side; it landed in `5d92443b`, the day after you wrote this, which is why the list is five and not four.

**`III_NAMESPACE`** — the one case that survives, and I am declining it as a design statement rather than on cost. A namespace is declared per file and a project *is* its file, so a container registering somewhere else is describing a different project; a second compose file says that, and keeps the property that reading one file tells you where everything in it lands. That is now written down in `understanding-iii/compose` rather than being implicit.

I take the cost seriously too — you named the plumbing exactly right, and readiness, the `ChildRecord` and `compose::status` all needing the overridden pair is the kind of thing that half-lands and turns into the silent timeout you describe. But that is a reason to do it carefully, not a reason not to. If a concrete case turns up for a shared namespace, worth reopening.

**Two gaps your comment surfaced.** `III_CONFIG_NAME` was missing from the reference table, and from the test that walks the reserved keys — a hand-written list of four that had not followed the constant. The key was enforced, just untested. The test now iterates `RESERVED_ENV`, so the next one added cannot slip through the same way.

crates/iii-compose/src/lib.rs
guibeira: Accepted in full. Both flags are gone in `38f6cc70`; compose serves in the foreground only.

Your reading of `--attach` was exactly right, and it was the easier half to argue: a flag named for a session that opens a file and seeks back 8 KiB is going to be discovered at the worst moment. Renaming it was the obvious smaller move, but it would have kept the log file alive, and the log file is the real problem.

That is the half that decided it. There is no rotation, no cap, no retention, and the append-is-never-truncate choice is correct on its own terms while making growth worse. Solving it means shipping rotation, and shipping rotation means either a dependency or our own copy of logrotate's edge cases — for a file that exists only because compose insisted on backgrounding itself.

**One detail in your description had aged.** The log stopped being machine-global in `c606647f`: it is `~/.iii/compose/<ns>/daemon.log`, one per daemon rather than one per machine. Which changes nothing about rotation, so the conclusion holds.

Gone with the flags: the re-exec, `III_COMPOSE_DETACHED`, the pid-matched readiness poll in `daemon_is_serving`, the tail loop, `daemon_log_path`, and the `DETACH_FAILED`, `NO_DAEMON_TO_ATTACH` and `CONFLICTING_FLAGS` errors that only those paths raised. 504 lines.

Both your alternatives are in `using-iii/compose` now — the shell redirect and the unit file, including your `Type=simple` note, which was already there and is now the documented path rather than the fallback.

**Two things I kept from the smoke suite.** `15-detach` and `19-attach` were 50 checks between them, but two properties in `15` had nothing to do with backgrounding and nothing else covered them: a second daemon naming a namespace already held exits with `DAEMON_ALREADY_SERVING` instead of shadowing the first, and `compose::stop` ends one daemon over the engine without touching its neighbour. It is now `15-daemon-lease`, with both daemons started the way a supervisor starts them, plus a check that the process the shell launched is the one serving — no re-exec in between (`compose-smoke-tests@4e05087`).

crates/iii-compose/src/state.rs
guibeira: Keeping the layout, but you moved three things (`4483440e`).

**Four of the five consequences you listed were consequences of the `id`.** They were accurate when you wrote this and stopped being so in `c606647f`, where the slug became a hash of the compose file's canonical path:

- the binding you wanted to make structural already is — a state directory cannot be pointed at another project;
- `STATE_BINDING_MISMATCH` had no raisers left, and its message still said *"Use a different id="*. Removed;
- two checkouts of one repository are already two projects, with no id juggling;
- `packages/` is keyed by name, version and target, and sharing it is deliberate — the smoke suite asserts a second project reuses the first's download.

**Discovery is the one that survives**, and you were right about it: finding a container's log means asking `compose::status` for `state_dir` rather than listing a directory you are standing in. The reference now shows that with the command rather than mentioning the field in passing.

**Three reasons the layout is global, which were nowhere before.** A checkout is not always writable, and a CI runner mounting the repository read-only would be unable to start a project at all. A `.iii/` directory in the user's tree is theirs to keep out of version control forever, and every generated file that lands there is one an ordinary `git add` sweeps up — that is a recurring cost paid by everyone who runs compose, in exchange for a shorter path. And `packages/` stays machine-global regardless, so in-tree state splits the layout in two rather than replacing it.

`$III_COMPOSE_STATE_DIR` is the escape hatch for anyone whose home directory is the wrong place. You read its existence as a signal the global path is inconvenient; I read it as the reason the question stays optional.

**One thing I am not fixing.** Removing a project still leaves its state behind — nothing prunes `~/.iii/compose/<ns>/<slug>` when the compose file goes away. That is real and unaddressed by any of the above. It looks like a `compose::forget file=...` rather than a layout change, but I have not added one.

And you were right that this interacts with the other note: `-d` is gone, so `daemon.log` is gone with it. What is left under the daemon's namespace is project state only.

crates/iii-compose/src/lifecycle.rs
guibeira: The diagnosis is right and I am not implementing either policy in this PR. Both are filed: #2051 for `required`, #2052 for `restart`.

The asymmetry is not a judgement call, it is two code paths answering differently for one declaration, and the docs said neither. They do now: the reference states both radii in a table, and `understanding-iii/compose` argues why each rule is defensible where it stands and what is missing — that the file has no way to say which it wants. It also says the thing that was nowhere at all, that there is no restart policy, so a crash is permanent until the next `up`.

**Why they are two issues rather than one.** `required` is a branch in `lifecycle.rs`. A restart policy is a supervisor, and needs backoff, a cap, a decision about what happens to dependents during the restart window, and a state `compose::status` does not have. Shipping it without backoff and a cap turns a crash loop into a busy loop, so it seemed worse to land a partial version than to name it.

**One thing worth adding to your `required: false` sketch.** It changes what the return means. Today `up` returns `status: ok` only when every container it planned is up, and the rollback rule is that an operation undoes itself — those two are what make `up` safe to retry. A non-required container that fails leaves an `ok` with something down, so the result has to say which, and `05-failure-rollback` asserts the current rule and would need to grow the case. That is in the issue.

Agreed on the defaults you suggested, and on the framing: whatever the knob turns out to be, the answer should read the same at start time and at run time.

crates/iii-compose/src/config.rs
anthonyiscoding: I don't agree with the logic here. The URI was what we discussed previously so that end users wouldn't have to rely on a single way to obtain configuration.

I also cannot find this quote in the tech specs or anywhere else. Did the AI hallucinate it?

> Names resolve through the configuration worker, whose adapter decides storage — local fs today, secrets manager or a separate iii instance in cloud. File paths are not the contract; names are, so the same compose file works when storage moves.

crates/iii-compose/src/spawn.rs
anthonyiscoding: Locally declared envs should always supersede hardcoded values which is how it works everywhere else. Is this the behavior that is happening?

crates/iii-compose/src/config.rs
guibeira: We discussed it with @sergiofilhowz , the reason config_uri using `file://` or `worker://configuration/get/<name>` is a change at configuration worker. Imagine we have a configuration-worker reading only the file, using `worker://configuration/get/<name>` mean we need to change the worker to handle this, and what `worker://` means? What is the source of this configuration? `worker://configuration/get/<name>` is the same as configuration_name, the way it going to be loaded depends on how we implemented configuration worker, today it's using file, tomorrow maybe a AWS KMS, so it's decoulpled.

PR #2041, opened 2026-08-05
fix(engine): fail over a function id instead of deleting it on disconnect

andersonleal:
## What

When two workers register the same function id, the engine now fails the id over to a still-connected worker on disconnect instead of deleting it.

`claim_function` keeps the outgoing handler in a new `superseded_functions` stack (it runs before `register_function` overwrites the entry, so the handler is still readable). `release_function_if_owner` pops the newest standby whose worker is still connected and moves only the owner index — the service registry keeps its entry because the function never goes away. Standbys for workers that have themselves disconnected are skipped and dropped, so failover can't resurrect a dead connection. The id is removed only when no live holder is left.

## Why

Two workers registering the same id is legal — the second overwrites the first in `FunctionsRegistry` and takes ownership in `function_owners`. But `release_function_if_owner` deleted the id outright when that second worker disconnected, even though the first was still connected and still serving it. Every later call returned `function_not_found` until the surviving worker was restarted.

Observed live: a second database worker started without `III_WORKER_NAME` registered `database::*` alongside the configured one; stopping it took the whole database surface down with it.

## Notes

- No behavior change for the single-registration case: with nothing superseded, the id is removed exactly as before.
- Two tests cover it: failover to the still-connected worker (and removal once the last holder goes), and skipping a standby whose worker already disconnected.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved function failover when a live worker replaces an existing function registration.
  * Automatically restores the newest available registration when the active worker disconnects.
  * Prevents disconnected or stale standby registrations from being restored.
  * Ensures functions are removed only when no valid fallback remains.

### DIFF
--- a/engine/src/engine/mod.rs
+++ b/engine/src/engine/mod.rs
@@ -230,6 +230,19 @@ pub struct Engine {
     /// still-live worker (the fast-restart race). In-process workers do not
     /// populate this map, so the absence of an entry means "no WS owner."
     pub(crate) function_owners: Arc<DashMap<String, Uuid>>,
+    /// Registrations for a function id that a later worker SUPERSEDED, oldest
+    /// first. `function_owners` holds one owner per id and `FunctionsRegistry`
+    /// one handler, so a second worker registering the same id overwrites the
+    /// first. Without this, that second worker disconnecting deletes the id
+    /// outright and the first worker — still connected, still believing it is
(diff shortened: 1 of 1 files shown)


andersonleal:
fixed in [1984](https://github.com/iii-hq/iii/pull/1984)

PR #2082, opened 2026-08-24
fix(compose): enforce --up flag contract

guibeira:
## What

- Reject programmatic `ComposeCli` values that set `file` without `up`.
- Update the README and startup script to use `iii compose --up`.
- Add the Compose startup breaking change and migration command to the 0.23.0 changelog.

## Why

This follow-up to #2081 makes the public planning API enforce the same contract as Clap and completes the command migration in repository guidance.

## Notes

`iii compose up` is rejected. Use `iii compose --up`; `--file` is valid only with `--up`.

## Testing

- `cargo test -p iii-compose` (250 passed)
- `cargo test -p iii --bin iii` (146 passed)
- `cargo clippy -p iii-compose --all-targets -- -D warnings`
- `bash -n scripts/start-iii.sh`
- docs skill artifacts verified against sources

## Summary by CodeRabbit

* **Breaking Changes**
  * Updated Compose CLI syntax: use `iii compose --up` instead of `iii compose up`.
  * The `--file` option now requires `--up`.
  * Running `iii compose` without `--up` only starts the daemon.

* **Documentation**
  * Updated quick-start instructions, scripts, and the upcoming 0.23.0 changelog to reflect the new command syntax.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ Before iii:
 After iii:
 
 - Declare workers in `worker-compose.yaml`
-- Run them with `iii compose up`
+- Run them with `iii compose --up`
 - Done. It is in the system, traceable, and callable.
 
 Platform teams publish workers. Application teams register functions and declare triggers. Agents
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -92,7 +92,7 @@ Before iii:
 After iii:
 
 - Declare workers in `worker-compose.yaml`
-- Run them with `iii compose up`
+- Run them with `iii compose --up`
```
guibeira, line 95: Fixed in fa6148c. The README now uses `iii compose --namespace dev --up`. A repository-wide scan also found and updated three positional `up` examples in the getting-started and engine-config skills.

PR #2100, opened 2026-08-28
docs(compose): correct and restructure the Compose reference

anthonyiscoding:
## What

- Corrected `docs/next/using-iii/compose.mdx` against `crates/iii-compose` on main: registry worker spelling, `container` payload scope for `down`, daemon stdout/stderr, `post_run` await behavior, config publish-back, `DAEMON_ALREADY_SERVING` vs `DAEMON_NAMESPACE_TAKEN`, `--file` requiring `--up`, and the error-code table (added ~20 missing codes).
- Added a dedicated `## Namespaces` section (daemon namespace, project namespace, precedence, charset, collisions) ahead of the `worker-compose.yaml` schema; removed the scattered namespace mentions it replaced.
- Updated all `iii compose` / `iii compose build` / `iii compose logs` option tables to match the clap-generated help text verbatim; regenerated `docs/next/cli-reference/index.mdx` from source and confirmed no drift.
- Replaced "container" with "worker" throughout the page's prose; kept `container` only where it is the literal payload field, CLI argument, YAML key, or error code.
- Fixed a broken `${VAR:default}` interpolation example (single colon isn't recognized; corrected to `${VAR:-default}`) and a typo.
- Re-rendered the page's skill artifact; `iii-skill-check` clean (structure, vale, verify-rendered).
- Removed `next/upgrading/from-0-21-x` from the `next` sidebar in `docs.json` (file kept on disk).

## Notes

Temporary `{/* REVIEW-N */}` markers used during review have been removed. A few findings remain open (not fixed in this PR): `container` payload scope description for `down`, `compose::stop` being daemon-wide, `engine.workers` not being `${VAR}`-interpolated, the engine URL/`iii-worker-manager` port-match rule, and `--engine` vs `engine:` ownership priority being scoped to a daemon's initial file only.

## Summary by CodeRabbit

* **Documentation**
  * Clarified that path workers using a base image run in a local VM, with invalid image references producing errors instead of falling back to host execution.
  * Documented startup command precedence and the error shown when no startup command is configured.
  * Explained VM storage behavior for bundles and local-image workers.
  * Updated engine-managed worker terminology and corrected the Compose example’s service dependencies.
  * Removed the outdated version 0.21.x upgrade guide.


docs/next/using-iii/compose.mdx
```diff
@@ -141,35 +63,31 @@ iii compose logs queue --stream stderr   # only stderr
...
+
+For example on most terminals (bash, zsh) you can run:
 
 ```bash
 iii compose --namespace dev --engine ws://127.0.0.1:49134 >> ~/iii-compose.log 2>&1 &
 ```
 
-On a server, a unit file gives restart-on-failure and hands the output to the journal. Set a
-namespace when this daemon must be isolated from another daemon on the same engine.
+Or on a server with systemd support you can use a unit file. This is a basic example:
```
guibeira, line 90: I'm not sure we need this init example

docs/next/using-iii/compose.mdx
```diff
@@ -187,203 +105,150 @@ WantedBy=multi-user.target
...
-so an edit made by hand takes effect without restarting the daemon. It accepts `container` like the
-two halves it is made of. Nothing cleverer yet: no rolling restart, and no keeping a container that
-did not change.
+### Adding workers
+
+`compose::add` declares one or more workers in the compose file and reconciles the project once. On
+the CLI, repeat `worker=` for each worker. Each value takes a registry package name (`state`), a
+package name with a version (`state@0.21.4`), or a directory (`./workers/api`).
+
+A JSON payload for this function can use a single list: `{ "workers": ["database", "web"] }`.
```
guibeira, line 181: Would be great to have a cli command example, like:
`iii trigger compose::add worker=<worker_name>`

docs/next/using-iii/compose.mdx
anthonyiscoding: I would prefer to keep, users will ask.

PR #2127, opened 2026-09-02
feat(compose): required: false contains a failed start

### LINKED ISSUE #2051: compose: let a container declare that its failure does not fail the operation
Split out of the review of #2034 ([comment](https://github.com/iii-hq/iii/pull/2034#discussion_r3714350907)).

## The problem

What one container's failure takes with it depends on when it happens.

| When | What comes down |
| --- | --- |
| During `up` | The operation ends. Everything it started is rolled back, and containers later in the start order are never attempted. |
| After it is ready | Its transitive dependents. Everything else keeps running. |

So a `mailer` that nothing depends on ends the whole start if it fails at `up`, and is contained if
it fails a minute later. Same declaration, same container, two blast radii separated only by timing.

Each rule is defensible on its own. What is missing is a way for the compose file to say which it
wants.

## Shape

```yaml
containers:
  mailer:
    worker: path://./workers/mailer
    required: false        # its failure does not fail the operation
```

`required: true` is the default and matches today's behaviour.

## What has to be decided with it

`up` currently returns `status: ok` only when every container it planned is up, and the rollback
rule is that an operation undoes itself. With `required: false` an `up` can return `ok` with a
container down, so the return has to say which. `05-failure-rollback` in the smoke suite asserts the
current rule and would need to grow the non-required case.

Documented as a v1 limitation in `understanding-iii/compose`.

tacoda:
A container's blast radius depended on the clock: a failure during `up` ended the operation and rolled back everything it had started, while the same failure a minute later took only that container's dependents. Same declaration, two blast radii separated by timing, and no way for the file to say which it wanted.

`required: false` says it. The failure is reported against that container, nothing is rolled back, and the operation carries on. Dependents carry on too: `start_after` is a start order, not a claim that the dependent cannot run without the dependency, so a container that waited on a non-required one starts as if it had come up. A dependent that genuinely needs it fails on its own.

`required: true` is the default, so a file written before this field keeps the behaviour it was written against.

That moves what `status: ok` means. It used to say every planned container is up; it now says every required one is, so the mutation return names the rest in `not_required_failures` rather than leaving a caller to compare the plan against a later status call.

Start time only. A ready container that exits still cascades into its dependents whatever it declared: that is the restart-policy half of the same question, and it does not exist yet. The docs say so.

Closes #2051

## What

A new container field, and the operation rule that follows from it:

```yaml
containers:
  mailer:
    worker: path://./workers/mailer
    required: false
```

- config.rs: required: bool, defaulting to true. Strict as everything else in the file, so a non-boolean is INVALID_COMPOSE_FILE.
- lifecycle.rs: a failed non-required container no longer ends the wave. It is recorded, the plan's progress advances for it, and the following waves start.
- daemon.rs: not_required_failures on the mutation return. Derived rather than plumbed, because a container result carrying an error under status: ok can only be a non-required one: a required failure is what makes the status failed. error still carries the first reason.
- report.rs: a line before the summary, so a partial project does not read as a clean start.

✗ mailer STARTUP_TIMEOUT container 'mailer' was not ready after 2s
✓ api ready (215ms)
1 container failed and is not required: the project is up without it
up: 1 of 2 changed in 2.2s

Tests: config_validation covers the default, the per-container declaration, and the rejection. compose_daemon_e2e runs the case the field exists for: mailer never registers and times
out, api declares start_after: [mailer] and still reaches ready, up returns oks: ["mailer"], and a status call confirms api was not rolled back.

Docs: the field row and the compose::up exception in using-iii/compose; in undhe blast radius depends on the clock" stops calling this a v1 limitation andgains a subsection for what the field settles and what it leaves open.

## Why

Each rule was defensible on its own. An up that reported success over a half-started project would be worse than one that refuses, and a supervisor that tore down a whole project because one leaf died would be worse than one that contains it. What was missise file to say which it wants, so the choice was compose's rather than theoperator's.

Split out of the review of #2034 (comment), and documented as a v1 limitation in understanding-iii/compose until now.

## Notes

- Not a breaking change. required: true is the default and is today's behaviour exactly. not_required_failures is omitted unless something failed under an ok.
- compose::restart is unchanged. A named container that fails still fails the ared: the operator named one container and asked for it.
- The .skill.md twins were edited by hand, because iii-skill-render was not available locally. Worth regenerating before merge.
- The 05-failure-rollback smoke case is not in this repository, so the non-reqded to it.
- One pre-existing test failure on macOS. add_starts_a_managed_project_declared_with_null_containers asserts on a /private tmpdir symlink path and fails identically on a clean checkout of main. Everything else in compose_daemon_e2e and all of cargo test -p iii-compose is green.

## Summary by CodeRabbit

* **New Features**
  * Added optional containers with `required: false`, allowing their startup failures without failing or rolling back the overall operation.
  * Dependent containers can still start when a non-required container fails.
  * Results now report non-required container failures while retaining a successful status.
  * The `required` setting defaults to `true`.

* **Documentation**
  * Updated compose configuration and lifecycle documentation with the new behavior and response details.

* **Tests**
  * Added validation and end-to-end coverage for optional container failures.

## Summary by CodeRabbit

* **New Features**
  * Added optional containers that can fail without failing the overall startup operation.
  * Startup responses identify non-required containers that failed.
  * Dependents can still start when a non-required container fails.
  * The `required` setting defaults to `true`.

* **Bug Fixes**
  * Prevented unnecessary rollback when non-required containers fail.

* **Documentation**
  * Documented the `required` setting and partial-start behavior.

### DIFF
--- a/crates/iii-compose/src/config.rs
+++ b/crates/iii-compose/src/config.rs
@@ -123,6 +123,19 @@ pub struct Container {
     pub env_file: Vec<PathBuf>,
     /// Readiness budget for this container: its own override, else the file's.
     pub startup_timeout: Duration,
+    /// Whether a failed start fails the operation that started it.
+    ///
+    /// `true` is the default and the rule everything else is written against:
+    /// an `up` that cannot start a declared container refuses and undoes
+    /// itself. `false` says the project would rather run without this one, so
+    /// the failure is reported against the container and the operation carries
+    /// on.
+    ///
+    /// Dependents carry on too. `start_after` is a start order, not a claim
+    /// that the dependent cannot run without the dependency, so a container
+    /// that waited on a non-required one starts as if it had come up. A
+    /// dependent that genuinely needs it says so by failing on its own.
+    pub required: bool,
 }
 
 #[derive(Debug, Clone, PartialEq)]
@@ -428,6 +441,7 @@ fn validate_container(
             .map(|path| resolve_relative(base_dir, path))
             .collect(),
         startup_timeout,
+        required: raw.required,
     })
 }
 
@@ -730,6 +744,15 @@ struct RawContainer {
     env_file: Vec<PathBuf>,
     #[serde(default)]
     startup_timeout: Option<String>,
+    /// Absent means `true`: the default is the strict rule, so a file written
+    /// before this field existed keeps the behaviour it was written against.
+    #[serde(default = "required_by_default")]
+    #[schemars(default = "required_by_default")]
+    required: bool,
+}
+
+fn required_by_default() -> bool {
+    true
 }
 
 #[derive(Debug, Deserialize, JsonSchema)]
--- a/crates/iii-compose/src/daemon.rs
+++ b/crates/iii-compose/src/daemon.rs
@@ -1026,6 +1026,14 @@ pub struct MutationOutcome {
     /// Concise failure for the first worker that could not reach its target state.
     #[serde(skip_serializing_if = "Option::is_none")]
     error: Option<MutationError>,
+    /// Workers that failed while the operation still succeeded, which is only
+    /// possible for a container declaring `required: false`.
+    ///
+    /// `status: ok` used to mean every planned container is up. It now means
+    /// every *required* one is, so the return has to name the rest rather than
+    /// leave a caller to compare the plan against a later status call.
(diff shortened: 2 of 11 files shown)


guibeira (approved):
Hey @tacoda thanks for the contribution!

PR #2153, opened 2026-09-09
fix(python-sdk): allow bounded invocation messages above 1 MiB

McHersheys:
## Problem

The Python SDK calls `websockets.connect` without `max_size`, inheriting the library's 1 MiB receive limit. Valid larger invocation inputs or results therefore close the worker transport before the handler can receive them. The caller can see only `invocation_stopped`, which obscures the actual limit.

## Fix

Set an explicit finite 16 MiB receive envelope on the SDK connection. This is an internal connection-default correction, not a new public SDK option or protocol/Engine change. Oversized messages remain rejected with WebSocket close code 1009. Add real-loopback WebSocket regressions for larger round trips and the finite bound.

## Evidence

- Unmodified current-main source: the new >1 MiB round-trip regression fails with close code 1009.
- Exact changed source: both regression tests pass, including 8,000,000-byte round-trip and rejection above 16 MiB.
- Separately, actual Engine 0.23.0 plus Python SDK 0.21.6 reproduced the dropped invocation before its handler. The same one-line connection fix passed full worker/Engine request-and-result round trips at ~1.06 MB and 8 MB.
- The current 0.23.0 SDK source was separately overlaid into that existing dependency environment and passed those native checks. This is source-level compatibility evidence, not a freshly installed whole 0.23.0 dependency-closure or production acceptance claim.
- Product checks ran in isolated, bounded Linux sandboxes. No Engine changes or broad monorepo build were made.

This PR intentionally leaves retries, invocation semantics, and public configuration unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Increased the maximum supported incoming WebSocket message size to 16 MB.
  * Messages exceeding this limit are now closed safely with the appropriate WebSocket error code.
  * Receive operations continue to enforce a finite timeout.

### DIFF
--- a/sdk/packages/python/iii/src/iii/iii.py
+++ b/sdk/packages/python/iii/src/iii/iii.py
@@ -422,6 +422,7 @@ class III:
             self._ws = await websockets.connect(
                 self._address,
                 additional_headers=self._options.headers,
+                max_size=16 * 1024 * 1024,  # Bounded engine-message envelope, not library 1 MiB default.
             )
             log.info(f"Connected to {self._address}")
             await self._on_connected()
--- a/sdk/packages/python/iii/tests/test_websocket_receive_bound.py
+++ b/sdk/packages/python/iii/tests/test_websocket_receive_bound.py
@@ -1,0 +1,64 @@
+"""Real WebSocket receive-bound regressions for the Python SDK connection.
+
+These tests exercise _do_connect without registration or reconnection machinery.
+Full worker/Engine request-and-response qualification is separate evidence.
+"""
+
+import asyncio
+import unittest
+from types import SimpleNamespace
+from unittest.mock import AsyncMock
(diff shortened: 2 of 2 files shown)


sdk/packages/python/iii/src/iii/iii.py
```diff
@@ -422,6 +422,7 @@ async def _do_connect(self) -> None:
             self._ws = await websockets.connect(
                 self._address,
                 additional_headers=self._options.headers,
+                max_size=16 * 1024 * 1024,  # Bounded engine-message envelope, not library 1 MiB default.
```
guibeira, line 425: Can we move this into a constant? Also, it would be great to allow users to change the max_size using an ENV. It could be `III_SDK_MAX_SIZE_WEBSOCKET`

PR #2159, opened 2026-09-09
(MOT-4732) fix(engine): forward a null function result as null, not as no result

andersonleal:
Closes [MOT-4732](https://linear.app/motia/issue/MOT-4732)

## Problem

`state::get` on a missing key answers `null`. `InvocationResult.result` is `Option<Value>`, and serde folds a JSON `null` into `None`, so the engine forwarded the reply with the field omitted and the Node SDK resolved `undefined`. The Linkly agent's natural `if (existing === null)` never matched and one `link::create` spun on `state::get` at ~575 calls/s: the span storms behind MOT-4728 and run-3 finding F-R3-1.

## Fix

- **engine** (`protocol.rs`): `result` deserializes through `deserialize_present` together with `#[serde(default)]`, so a present `null` is `Some(Value::Null)` and only an absent field is `None`. Serialization is unchanged: `Some(Null)` goes out as `"result": null`, `None` stays omitted, so a void reply still resolves `undefined` in Node.
- **python sdk**: `_to_dict` keeps an explicit `result=None` on the wire instead of dropping it through `exclude_none`. Error replies never set `result` and are untouched.
- **node sdk**: no code change; regression tests pin both directions (frame → promise, handler → frame).
- Checked the other SDKs: Rust `trigger` already does `unwrap_or(Value::Null)`; Go marshals a `nil` return as `null` inside `json.RawMessage`.

## Tests

- `protocol::tests::invocation_result_null_stays_null_and_absent_stays_absent`
- `engine::tests::a_null_result_from_the_executor_reaches_the_caller_as_null`: a worker replies with a raw `"result": null` frame; the caller's frame carries `"result":null`.
- Node `tests/invocation-result-null.test.ts` (3 tests), Python `tests/test_invocation_result_wire.py` (3 tests).

## Live check

Release build of this branch, iii-sdk 0.23.0 worker registering three functions and triggering them:

```
probe::null:  typeof=object    ===null=true   json=null
probe::void:  typeof=undefined ===null=false  json=undefined
probe::value: typeof=object    ===null=false  json={"ok":true}
```

Before (main, from the ticket): `typeof: undefined | r === null: false`.

https://claude.ai/code/session_01PvtXK7JgHHNrG192afbRAk

## Summary by CodeRabbit

- **Bug Fixes**
  - Explicit `null` invocation results are now preserved and delivered to callers as `null`.
  - Missing results remain distinguishable from explicit `null` values.
  - State lookups for missing items now consistently return `null`.
  - Python responses preserve `None` as `null`, while error responses remain unchanged.

- **Documentation**
  - Added a changelog entry describing improved reliability for null-valued invocation results.

### DIFF
--- a/sdk/packages/python/iii/src/iii/iii.py
+++ b/sdk/packages/python/iii/src/iii/iii.py
@@ -533,6 +533,14 @@ class III:
             data: dict[str, Any] = msg.model_dump(by_alias=True, exclude_none=True)
             if "type" in data and hasattr(data["type"], "value"):
                 data["type"] = data["type"].value
+            # A handler that returned None answered `null`, not "nothing":
+            # keep it on the wire so a TypeScript caller sees null instead
+            # of undefined (MOT-4732). Error replies never set `result`.
+            if (
+                isinstance(msg, InvocationResultMessage)
+                and "result" in msg.model_fields_set
+            ):
+                data.setdefault("result", None)
             return data
         return {"data": msg}
 
--- a/sdk/packages/python/iii/src/iii/iii.py
+++ b/sdk/packages/python/iii/src/iii/iii.py
@@ -533,9 +533,6 @@ class III:
             data: dict[str, Any] = msg.model_dump(by_alias=True, exclude_none=True)
             if "type" in data and hasattr(data["type"], "value"):
                 data["type"] = data["type"].value
-            # A handler that returned None answered `null`, not "nothing":
-            # keep it on the wire so a TypeScript caller sees null instead
-            # of undefined (MOT-4732). Error replies never set `result`.
             if (
                 isinstance(msg, InvocationResultMessage)
                 and "result" in msg.model_fields_set
--- a/docs/changelog/index.mdx
+++ b/docs/changelog/index.mdx
@@ -43,6 +43,9 @@ type: "reference"
 
   ## Reliability
 
+  - A function that returns `null` now reaches its caller as `null`. The engine used to drop the
+    field on the way through, so the Node SDK resolved `undefined` and a strict `=== null` check on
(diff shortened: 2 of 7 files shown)


sdk/packages/python/iii/src/iii/iii.py
```diff
@@ -533,6 +533,14 @@ def _to_dict(self, msg: Any) -> dict[str, Any]:
             data: dict[str, Any] = msg.model_dump(by_alias=True, exclude_none=True)
             if "type" in data and hasattr(data["type"], "value"):
                 data["type"] = data["type"].value
+            # A handler that returned None answered `null`, not "nothing":
+            # keep it on the wire so a TypeScript caller sees null instead
+            # of undefined (MOT-4732). Error replies never set `result`.
```
guibeira: You can remove those comments

guibeira (dismissed):
Looks good. Can you confirm if we need to update the Go and Rust sdk tests as well?

sdk/packages/python/iii/src/iii/iii.py
andersonleal: Done in a644aadae.

Removed there, and applied the same pass to the rest of the PR, since the null-vs-absent rule was stated four times:

- the engine test comment is gone, the test name is the claim
- `protocol.rs` field doc trimmed from 6 lines to 3, matching the neighbouring `metadata` / `namespace` field docs

Two I kept deliberately:

- a short doc on `deserialize_present`, because the `default` + `deserialize_with` pairing is the one part that is not self-evident
- one line on the node assertion, so the `toBeUndefined()` -> `toBeNull()` flip is not read as a weakened test and reverted later

PR #2167, opened 2026-09-10
fix(cli): honor III_URL in iii trigger

anthonyiscoding:
Closes MOT-4746.

## Problem

`iii trigger` took its endpoint as a host and a port and ignored `III_URL`, the variable every managed context already sets: compose exports it to each worker it spawns, and `iii compose --engine` reads it. On a project that does not own port 49134, every call silently reached whatever engine did.

The failure is quiet, because a call that lands on the wrong engine still gets an answer:

```
~/iii/projects/testing/compose/test30-restarting-status ❯ iii trigger -n restarting-status compose::restart worker=slowboot
Error: function_not_found ... not found in namespace restarting-status.
It is registered in namespace(s): onboardingwork.
```

## Change

**A `--engine` flag, matching `iii compose`.** `iii trigger --engine ws://host:port` names the engine in one value, scheme included. The endpoint order is:

| Precedence | Source |
| --- | --- |
| 1 | `--address` / `--port` (each over its own half of the URL) |
| 2 | `--engine <URL>` |
| 3 | `III_URL` |
| 4 | `ws://localhost:49134` |

**`--address` and `--port` are deprecated.** They keep working and keep overriding their own half of whichever URL won, so `--port 4300` still retargets the port and inherits the host. Using either prints a notice on **stderr**, never stdout, so output a script parses is unchanged:

```
warning: --port is deprecated; use `--engine ws://host:port` instead
```

**`wss://` is refused, from any source.** No iii engine terminates TLS: the worker socket is a plain TCP listener with an HTTP/1 upgrade and no TLS acceptor, and both call sites rebuild the address as `ws://{host}:{port}`. Accepting a `wss://` value would send a caller who asked for TLS over the wire in the clear (CWE-319), so it fails instead:

```
Error: Encrypted websocket connections are not currently supported.
```

**The accepted shapes are exactly `ws://host` and `ws://host:port`.** The endpoint is rebuilt as `ws://{host}:{port}`, so anything past the authority used to be dropped without a word: `ws://host:4300/ws` lost its path, and `ws://user:secret@host:4300` connected with no credentials while the caller believed it had sent them. Each part is now refused and named:

```
Error: --engine "ws://127.0.0.1:4300/ws" must not have a path; expected ws://host or ws://host:port
Error: --engine "ws://127.0.0.1:4300?x=1" must not have a query string; expected ws://host or ws://host:port
Error: --engine "ws://127.0.0.1:4300#f" must not have a fragment; expected ws://host or ws://host:port
Error: --engine "ws:127.0.0.1:4300" must start with ws://; expected ws://host or ws://host:port
Error: --engine must not carry credentials; expected ws://host or ws://host:port
```

The credentials error is the one message that does not echo the value it rejects: the value holds the password, and an error line ends up in terminals, CI logs and bug reports. A test asserts that neither the username nor the password appears in it.

**An unusable URL fails the call, whichever source carried it.** An earlier version of this PR fell back to `localhost:49134` for a malformed `III_URL`. That reintroduces the bug the PR exists to fix: the call reaches whichever engine owns that port while the operator believes the variable pointed it elsewhere. The error names its source, because a stale variable and a mistyped flag need different fixes:

```
Error: III_URL "not a url" must start with ws://; expected ws://host or ws://host:port
Error: III_URL must not carry credentials; expected ws://host or ws://host:port
```

A blank `III_URL` is the exception: it is read as unset rather than unusable, so an exported but empty variable still resolves to the local default. A blank `--engine` is not, because the caller typed the flag. A URL with no port keeps 49134, because the `url` crate's scheme defaults (80 and 443) are not this engine's, and a bare trailing slash is accepted since `ws://host:4300` and `ws://host:4300/` are the same address.

## Verification

`cargo test -p iii --bins cli_trigger`: 45 passed. Cases cover each precedence step, both flags overriding halves, `wss` in any spelling, every rejected URL shape, malformed input from each source, a blank value from each source, credentials kept out of the error text, IPv6 brackets, and a URL with no port.

Live, against an engine started by compose on `ws://127.0.0.1:4300`:

| Command | Result |
| --- | --- |
| `--engine ws://127.0.0.1:4300` | reached the engine |
| `--engine wss://127.0.0.1:4300` | `Error: Encrypted websocket connections are not currently supported.` |
| `III_URL=wss://engine.test:443` | same error |
| `--port 4300` | deprecation notice on stderr, then the result |
| `--engine http://127.0.0.1:4300` | `Error: --engine "http://127.0.0.1:4300" must start with ws://…` |
| `--engine ws://127.0.0.1:4300/ws` | `Error: --engine "ws://127.0.0.1:4300/ws" must not have a path…` |
| `--engine ws://user:secret@127.0.0.1:4300` | `Error: --engine must not carry credentials…` (value not echoed) |
| `--engine ""` | `Error: --engine "" must start with ws://…` |

`cargo fmt --check -p iii` and `cargo clippy -p iii --bins` are clean. `docs/next/cli-reference/index.mdx` and its `.skill.md` were regenerated with `scripts/generate-cli-docs.sh`.

## Note on TLS

This makes `iii trigger` honest about `wss://`, it does not add TLS. Nothing in the project serves it. Worth filing separately: `crates/iii-compose/src/managed_engine.rs:304` accepts `wss://` in `engine.url`, then binds a plaintext listener from its host and port while the daemon connects with the `wss://` URL, so that combination cannot work today either.

### DIFF
--- a/engine/src/cli_trigger/mod.rs
+++ b/engine/src/cli_trigger/mod.rs
@@ -127,9 +127,11 @@ fn engine_url_endpoint(raw: &str, source: &str) -> anyhow::Result<(String, u16)>
 /// won, so a flag can retarget one component and inherit the other.
 ///
 /// A URL that cannot be used fails the call, whichever source carried it: a
-/// `wss://` one with its own message, anything else as malformed. An empty or
-/// blank value is not a URL at all and is skipped, so an exported but unset
-/// `III_URL` still resolves to the default.
+/// `wss://` one with its own message, anything else as malformed. A blank
+/// `III_URL` is the exception, read as unset rather than as a URL, so an
+/// exported but empty variable still resolves to the default. A blank
+/// `--engine` is not: the caller typed the flag, so it fails like any other
+/// value that names no engine.
 ///
 /// Takes the environment value as an argument rather than reading it, so the
 /// precedence is testable without mutating process state.
@@ -139,14 +141,21 @@ fn resolve_endpoint(
     engine_flag: Option<&str>,
     engine_url: Option<&str>,
 ) -> anyhow::Result<(String, u16)> {
-    let named = [(engine_flag, "--engine"), (engine_url, "III_URL")]
-        .into_iter()
-        .find_map(|(value, source)| {
-            value
-                .map(str::trim)
-                .filter(|url| !url.is_empty())
-                .map(|url| (url, source))
-        });
+    let named = match engine_flag {
+        // Typed by the caller, so even an empty value is a statement: it names
+        // no engine, and quietly reading another source would send the call
+        // somewhere they did not ask for. Blank values from the other sources
+        // mean the opposite, that nothing was set.
+        Some(flag) => Some((flag, "--engine")),
+        None => [(engine_url, "III_URL")]
+            .into_iter()
+            .find_map(|(value, source)| {
+                value
+                    .map(str::trim)
+                    .filter(|url| !url.is_empty())
+                    .map(|url| (url, source))
+            }),
+    };
     let (source_host, source_port) = match named {
         Some((url, source)) => {
             let (host, port) = engine_url_endpoint(url, source)?;
@@ -390,7 +399,7 @@ mod tests {
     fn an_unusable_engine_flag_errors_instead_of_falling_back() {
         // The caller typed this one, so silence would send the call to
         // localhost while the operator believes it went somewhere else.
-        for url in ["not a url", "http://127.0.0.1:49734", "ws://"] {
+        for url in ["", "   ", "not a url", "http://127.0.0.1:49734", "ws://"] {
             let err = resolve_endpoint(None, None, Some(url), None)
                 .unwrap_err()
                 .to_string();
@@ -399,6 +408,14 @@ mod tests {
                 "unexpected error for --engine {url:?}: {err}"
             );
         }
+    }
+
+    #[test]
+    fn a_blank_engine_flag_does_not_fall_through_to_the_environment() {
+        let err = resolve_endpoint(None, None, Some("  "), Some("ws://127.0.0.1:49734"))
+            .unwrap_err()
+            .to_string();
+        assert!(err.starts_with("--engine"), "unexpected error: {err}");
     }
 
     #[tokio::test]
--- a/engine/src/cli_trigger/mod.rs
+++ b/engine/src/cli_trigger/mod.rs
@@ -41,13 +41,14 @@ pub struct TriggerArgs {
     #[arg(long)]
     pub json: Option<String>,
 
-    /// Engine host address.
-    #[arg(long, default_value = "localhost")]
-    pub address: String,
+    /// Engine host address. Taken from `III_URL` when omitted, else
+    /// `localhost`.
+    #[arg(long)]
+    pub address: Option<String>,
 
-    /// Engine WebSocket port.
-    #[arg(long, default_value_t = DEFAULT_PORT)]
-    pub port: u16,
+    /// Engine WebSocket port. Taken from `III_URL` when omitted, else 49134.
+    #[arg(long)]
+    pub port: Option<u16>,
 
     /// Max time to wait for the invocation result (milliseconds).
     #[arg(long, default_value_t = 30_000)]
@@ -65,12 +66,77 @@ pub struct TriggerArgs {
     pub help: bool,
 }
 
+/// Host used when neither a flag nor `III_URL` names one.
+const DEFAULT_ADDRESS: &str = "localhost";
+
+/// Splits a `ws://` or `wss://` engine URL into host and port.
+///
+/// Returns `None` for anything that is not one of those, so a stale or
+/// malformed `III_URL` falls back to the defaults instead of failing a command
+/// that never asked about it. A URL without a port keeps `DEFAULT_PORT`: the
+/// scheme defaults the `url` crate knows (80 and 443) are not this engine's.
+fn engine_url_endpoint(raw: &str) -> Option<(String, u16)> {
+    let url = url::Url::parse(raw.trim()).ok()?;
+    if !matches!(url.scheme(), "ws" | "wss") {
+        return None;
+    }
+    let host = match url.host()? {
+        url::Host::Ipv6(address) => format!("[{address}]"),
+        host => host.to_string(),
+    };
+    Some((host, url.port().unwrap_or(DEFAULT_PORT)))
+}
+
+/// Resolves the engine endpoint from the flags and `III_URL`.
+///
+/// `III_URL` is how every managed context already names its engine: compose
+/// exports it to each worker it spawns, and `iii compose --engine` reads it.
+/// This CLI used to ignore it, so on a project that does not own port 49134
+/// every call silently reached whatever engine did.
+///
+/// `--address` and `--port` still win, each over its own half of the URL, so a
+/// flag can retarget one component and inherit the other.
+///
+/// Takes the environment value as an argument rather than reading it, so the
+/// precedence is testable without mutating process state.
+fn resolve_endpoint(
+    address: Option<&str>,
+    port: Option<u16>,
+    engine_url: Option<&str>,
+) -> (String, u16) {
+    let from_env = engine_url
+        .map(str::trim)
+        .filter(|url| !url.is_empty())
+        .and_then(engine_url_endpoint);
+    let (env_host, env_port) = match from_env {
+        Some((host, port)) => (Some(host), Some(port)),
+        None => (None, None),
+    };
+
+    (
+        address
+            .map(str::to_string)
+            .or(env_host)
+            .unwrap_or_else(|| DEFAULT_ADDRESS.to_string()),
+        port.or(env_port).unwrap_or(DEFAULT_PORT),
+    )
+}
+
+impl TriggerArgs {
+    /// The engine this invocation talks to.
+    fn endpoint(&self) -> (String, u16) {
+        let engine_url = std::env::var("III_URL").ok();
+        resolve_endpoint(self.address.as_deref(), self.port, engine_url.as_deref())
+    }
+}
+
 pub async fn run_trigger(args: &TriggerArgs) -> Result<(), TriggerCliError> {
+    let (address, port) = args.endpoint();
     if args.help {
         help::print(
             args.function_path.as_deref(),
-            &args.address,
-            args.port,
+            &address,
+            port,
             args.timeout_ms,
             args.namespace.as_deref(),
         )
@@ -92,8 +158,8 @@ pub async fn run_trigger(args: &TriggerArgs) -> Result<(), TriggerCliError> {
     exec::invoke(
         function_path,
         payload,
-        &args.address,
-        args.port,
+        &address,
+        port,
         args.timeout_ms,
         args.namespace.as_deref(),
     )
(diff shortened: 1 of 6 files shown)


engine/src/cli_trigger/mod.rs
```diff
@@ -65,12 +66,77 @@ pub struct TriggerArgs {
...
+fn engine_url_endpoint(raw: &str) -> Option<(String, u16)> {
+    let url = url::Url::parse(raw.trim()).ok()?;
+    if !matches!(url.scheme(), "ws" | "wss") {
+        return None;
+    }
+    let host = match url.host()? {
+        url::Host::Ipv6(address) => format!("[{address}]"),
+        host => host.to_string(),
+    };
+    Some((host, url.port().unwrap_or(DEFAULT_PORT)))
```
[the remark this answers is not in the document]
anthonyiscoding: Confirmed and fixed in 3e8dc12. The scheme was dropped by `engine_url_endpoint`, and both `exec::invoke` and `help::print` rebuilt the address as `ws://{host}:{port}`, so a `wss://` value in `III_URL` was sent in the clear.

Rather than carry the scheme through, `wss://` is now refused from every source:

```
Error: Encrypted websocket connections are not currently supported.
```

Carrying it would have promised TLS the project cannot serve. The engine's worker socket is a plain `TcpListener` with an HTTP/1 upgrade and no TLS acceptor (`engine/src/workers/worker/mod.rs`), so a `wss://` connection could only ever reach a TLS proxy that no part of this repo provides. An explicit refusal is honest; a silent downgrade was not, and a working TLS handshake against a plaintext listener would not have been either.

Regression tests added: `a_wss_environment_url_is_refused_rather_than_downgraded`, `a_wss_engine_flag_is_refused`, and `a_wss_url_is_refused_even_when_both_flags_supply_the_endpoint`.

Two related changes came with it: `iii trigger --engine ws://host:port` now matches `iii compose --engine`, with `--address` and `--port` deprecated (still working, warning on stderr); and an unusable URL from any source fails the call instead of falling back to `localhost:49134`, which would have reached whichever engine owned that port.

Separately, and outside this PR: `crates/iii-compose/src/managed_engine.rs:304` accepts `wss://` in `engine.url`, then binds a plaintext listener from its host and port while the daemon connects using the `wss://` URL. That combination cannot work today.

engine/src/cli_trigger/mod.rs
```diff
@@ -65,12 +73,136 @@ pub struct TriggerArgs {
...
+///
+/// Takes the environment value as an argument rather than reading it, so the
+/// precedence is testable without mutating process state.
+fn resolve_endpoint(
+    address: Option<&str>,
+    port: Option<u16>,
+    engine_flag: Option<&str>,
+    engine_url: Option<&str>,
+) -> anyhow::Result<(String, u16)> {
+    let from_source = match engine_flag.map(str::trim).filter(|url| !url.is_empty()) {
```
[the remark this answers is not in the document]
anthonyiscoding: Correct finding, but it was not fixed by 3e8dc12. I verified the built binary from that commit:

```
$ III_URL=ws://127.0.0.1:4999 iii trigger compose::list --engine "" --timeout-ms 3000
Error: Timed out waiting for the engine (no response within the timeout). ...
```

The empty flag was still filtered out and `III_URL` still won, which is what the timeout on the dead port shows. Fixed in 216a88e, which takes the flag before any trimming:

```
$ III_URL=ws://127.0.0.1:4999 iii trigger compose::list --engine ""
Error: --engine "" must be a ws:// URL with a host, e.g. ws://localhost:49134
```

A blank `III_URL` keeps the opposite reading, unset rather than unusable, so an exported but empty variable still resolves to the local default. The two cases differ because the caller typed the flag.

Coverage: `""` and `"   "` are now in `an_unusable_engine_flag_errors_instead_of_falling_back`, plus `a_blank_engine_flag_does_not_fall_through_to_the_environment` for the precedence itself.

engine/src/cli_trigger/mod.rs
```diff
@@ -65,12 +73,167 @@ pub struct TriggerArgs {
...
+/// `wss://` is refused with its own message. No iii engine terminates TLS, so
+/// accepting one would send a caller who asked for TLS over the wire in the
+/// clear.
+///
+/// A URL without a port keeps `DEFAULT_PORT`: the scheme defaults the `url`
+/// crate knows (80 and 443) are not this engine's.
+fn engine_url_endpoint(raw: &str, source: &str) -> anyhow::Result<(String, u16)> {
+    let raw = raw.trim();
+    let malformed = |detail: &str| {
+        anyhow::anyhow!("{source} {raw:?} {detail}; expected ws://host or ws://host:port")
```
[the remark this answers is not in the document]
anthonyiscoding: Confirmed and fixed in ed7c438d9.

Reproduced first: `iii trigger compose::list --engine "http://user:secret@127.0.0.1:4400"` printed the password back. Every rejection path now quotes a redacted form, including the wrong-scheme path that reports before the URL is ever parsed:

```
Error: --engine "http://***@127.0.0.1:4400" must start with ws://; expected ws://host or ws://host:port
```

Host and port survive so the message still tells you what was wrong. Covered by `credentials_never_reach_the_error_text` (five URL shapes, asserts neither `user` nor `secret` appears) and `the_redacted_value_keeps_the_part_that_helps`.

engine/src/cli_trigger/mod.rs
```diff
@@ -65,12 +73,167 @@ pub struct TriggerArgs {
...
+    };
+
+    // Checked before parsing, so `wss:` in any form is named for what it is.
+    if raw.len() >= 4 && raw[..4].eq_ignore_ascii_case("wss:") {
```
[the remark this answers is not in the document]
anthonyiscoding: Confirmed and fixed in ed7c438d9.

Reproduced first: `iii trigger compose::list --engine "wssé"` panicked at `cli_trigger/mod.rs:110:29`, as did `"ws:é"` and `"wss√"`. Both prefix checks now use `raw.get(..4)` / `raw.get(..5)` with `is_some_and`, so a short or non-ASCII value falls through to the normal rejection:

```
Error: --engine "wssé" must start with ws://; expected ws://host or ws://host:port
```

Covered by `a_non_ascii_value_is_rejected_rather_than_panicking` over `["wssé", "ws:é", "wsé", "wss√", "é", "ws"]`.

engine/src/cli_trigger/mod.rs
```diff
@@ -65,12 +76,229 @@ pub struct TriggerArgs {
...
+    }
+    if url.fragment().is_some() {
+        return Err(malformed("must not have a fragment"));
+    }
+
+    let host = match host {
+        url::Host::Ipv6(address) => format!("[{address}]"),
+        host => host.to_string(),
+    };
+    Ok((host, url.port().unwrap_or(DEFAULT_PORT)))
```
guibeira, line 181: **[P2] Preserve an explicitly configured port 80**

`url::Url` normalizes away a scheme's default port, so `Url::parse("ws://engine.test:80")?.port()` returns `None` even though the caller explicitly supplied `:80`. Consequently, `engine_url_endpoint("ws://engine.test:80", "--engine")` currently returns `("engine.test", 49134)`, not port 80. I reproduced this with `url` 2.5.8.

This affects URLs selected from `--engine`, `III_URL`, or the Compose file, unless `--port` overrides the result. It can send an invocation to a different engine on the same host, recreating the silent wrong-endpoint failure this PR is intended to fix.

Please preserve whether the input authority explicitly contained a port and apply `DEFAULT_PORT` only when the port was actually omitted. Simply switching to `port_or_known_default()` would not preserve the intended 49134 default for a URL without a port. Add regression coverage for explicit `:80`, an omitted port, and a non-default explicit port.

engine/src/cli_trigger/mod.rs
```diff
@@ -65,12 +76,229 @@ pub struct TriggerArgs {
...
+    let authority_end = rest.find(['/', '?', '#']).unwrap_or(rest.len());
+    let authority = &rest[..authority_end];
+    match authority.rfind('@') {
+        Some(at) => format!(
+            "{}***@{}{}",
+            &raw[..rest_start],
+            &authority[at + 1..],
+            &rest[authority_end..]
+        ),
+        None => raw.to_string(),
```
guibeira, line 109: **[P2] Avoid echoing credentials when the authority has missing or extra slashes**

The redaction helper and the URL parser disagree about where the authority starts. For `ws:///user:secret@engine.test:4300`, this helper stops at the extra slash and returns the input unchanged, while `url::Url::parse` successfully recognizes `user` and `secret` as credentials. The resulting validation error is:

```text
--engine "ws:///user:secret@engine.test:4300" must not carry credentials; expected ws://host or ws://host:port
```

The malformed input `ws:/user:secret@engine.test:4300` also exposes the credentials through the earlier scheme-validation error. Both cases were reproduced in isolated tests using the PR's logic and `url` 2.5.8; the existing regular-authority redaction case passes.

A typo in an endpoint can therefore leak a password into terminal or CI logs despite the redaction guarantee. The safest fix is to omit the supplied endpoint text from validation errors and retain only the source, diagnostic detail, and expected format. Please add regression tests for both missing- and extra-slash inputs, including the rejection path before URL parsing.

engine/src/cli_trigger/mod.rs
```diff
@@ -65,12 +76,229 @@ pub struct TriggerArgs {
...
+/// a broken compose file must never fail a command that did not ask about it.
+/// A file that does state an engine address is held to the same rule as every
+/// other source, and an unusable one there fails the call.
+fn compose_file_engine_url() -> Option<String> {
+    let path = std::path::Path::new(iii_compose::cli::DEFAULT_COMPOSE_FILE);
+    let text = std::fs::read_to_string(path).ok()?;
+    // Only the engine section, so a container the running binary cannot parse
+    // does not cost the caller the address the file plainly states.
+    let engine = iii_compose::config::parse_engine_section(&text, path).ok()??;
+    (!engine.url.trim().is_empty()).then_some(engine.url)
```
guibeira, line 198: **[P2] Do not turn Compose parsing failures into a silent default-engine selection**

`.ok()??` collapses every `parse_engine_section` error into an absent URL. That parser does more than extract `engine.url`: it also expands engine values, deserializes the engine schema, and validates `engine.workers`. For example, a file can still contain the correct `engine.url: ws://127.0.0.1:49934`, but an unsupported worker name under `engine.workers` makes `validate_engine` return an error. This helper then discards the address, and with no endpoint flags or `III_URL`, resolution silently selects `localhost:49134`. An unresolved variable in the declared URL takes the same fallback path.

This matters when an existing project's file is temporarily broken or contains configuration from a newer release: a command intended for its running engine can reach another engine rather than report the configuration problem. Reading only the engine section avoids container-graph validation, but it does not make URL extraction independent of unrelated engine settings.

Please distinguish an absent file/engine from an existing configuration that cannot be interpreted. Extracting the endpoint without validating worker configuration would preserve a usable declared address; when that address itself cannot be resolved, return a source-specific error instead of choosing the default. Add tests that exercise the actual file-reading path with a valid non-default URL plus an invalid worker entry, and with an unresolved variable in `engine.url`.

guibeira (commented):
The endpoint-resolution change addresses the original wrong-engine issue, but I found three remaining edge cases worth fixing before merging. Inline comments include the impact and suggested regression coverage. Validation: isolated Rust tests using the PR's endpoint/redaction logic and url 2.5.8 reproduced the explicit-port and credential-redaction issues; the Compose fallback finding is based on tracing parse_engine_section and its error handling. I did not run the full repository test suite.

andersonleal (dismissed):

sergiofilhowz (dismissed):

engine/src/cli_trigger/mod.rs
anthonyiscoding: Confirmed and fixed in a481ad43. `ws://engine.test:80` resolved to `engine.test:49134`, exactly as you describe: `Url::port` drops a port that matches the scheme default, and 80 is the default for `ws`.

Your note on `port_or_known_default` is right, so the port is read back out of the authority only when the parser reports none:

```rust
url.port().or_else(|| explicit_port(raw)).unwrap_or(DEFAULT_PORT)
```

`explicit_port` takes the text between the scheme and the first `/`, `?` or `#`, drops anything before an `@`, and for a bracketed IPv6 host reads only past the `]`, so the colons inside the address are not mistaken for a port.

Coverage: `an_explicitly_named_port_80_survives` (`ws://engine.test:80`, the same with a trailing slash, and `ws://[::1]:80`), `a_non_default_explicit_port_is_kept`, and the existing `engine_url_without_a_port_keeps_the_engine_default` for the omitted case. `cargo test -p iii --bins cli_trigger`: 56 passed.

engine/src/cli_trigger/mod.rs
anthonyiscoding: Confirmed and fixed in a481ad43. Both shapes reproduced: the helper computed an empty authority, found no `@`, and returned the value unchanged, while `Url::parse` still read `user` and `secret` as credentials.

Fixed by looking where the parser looks. The helper now skips the leading slashes before it reads the authority:

```rust
let rest_start = rest_start
    + raw[rest_start..].bytes().take_while(|byte| *byte == b'/').count();
```

```
--engine "ws:///***@engine.test:4300" must not carry credentials; expected ws://host or ws://host:port
--engine "ws:/***@engine.test:4300" must start with ws://; expected ws://host or ws://host:port
```

I kept the value in the message rather than dropping it. A rejected endpoint is usually a typo, and the host and port are what tell the operator which value was read; your `ws:///` case is itself an example, since the error is only useful if it shows where the extra slash is. Both inputs are now in `credentials_never_reach_the_error_text`, including the second one, which is reported before the URL is ever parsed.

If you would still rather the text went away entirely, say so and I will cut it, the tests are already in place either way.

engine/src/cli_trigger/mod.rs
anthonyiscoding: Confirmed and fixed in a481ad43. You are right that reading only the engine section keeps the container graph out of the way but leaves the URL hostage to unrelated engine settings: `parse_engine_section` ends in `raw.engine.map(validate_engine)`, and `validate_engine` rejects the whole section over a worker name before it ever reaches `engine.url`.

New `config::parse_engine_url` reads and expands that one value and nothing else, so a worker entry this binary does not know costs the caller nothing:

```rust
let Some(mut url) = document
    .as_mapping()
    .and_then(|mapping| mapping.get(&engine_key))
    .and_then(serde_yaml::Value::as_mapping)
    .and_then(|engine| engine.get(&url_key))
    .cloned()
else {
    return Ok(None);
};
crate::interpolate::expand_tree(&mut url, path, &|name| std::env::var(name).ok())?;
```

On your second point I took the rule the rest of the PR uses: a file that names an address but cannot be read for one fails the call, naming the source, rather than selecting a default endpoint that belongs to some other engine. An unresolved variable in `engine.url` reaches `engine_url_endpoint` as the literal it expanded to and is rejected there like any other unusable value. Only an absent file, or one that states no `engine.url`, is still silent, since `iii trigger` has to work outside a project at all.

Coverage: `config::tests::the_engine_url_survives_an_unsupported_worker_name` asserts `parse_engine_section` errors on that document while `parse_engine_url` returns `ws://127.0.0.1:49934`, and `a_file_with_no_engine_url_reports_no_address` covers the quiet cases. The existing `an_unusable_compose_url_errors_and_names_the_file` covers the failing address. `cargo test -p iii-compose` and `cargo test -p iii --bins cli_trigger` both pass.

